PR #32, opened 2016-12-20
fixed css path

Gcaufy:


QingWei-Li:
`themes/` is automatically generated by `npm run build`. I don't think we need to modify this path.

Gcaufy:
I see, I didn't run `npm run build`, that's why I did't get the 'themes/' directory.
I was following your instrocutions:

>* Fork it!
>* Run npm i && npm run dev
>* open localhost:3000

then I get a missing css page, that's why I changed the css path.

so I guess you should change Step2 to `Run npm i && npm run build && npm run dev`

QingWei-Li:
`npm run dev` should execute `npm run build`

I will fix the script

PR #39, opened 2016-12-27
Add new feature: access password

LancerComet:
```javascript
// In app.js
...
const keepItSecret = true  // Now a password is require to access this docsify :)
...
```


QingWei-Li:
Thank you for your Pull Request.

I have two questions.
- Why move `lib/` folder to `public/`?  It will break the current use.
- How do we use this feature in GitHub Pages? 

However, i likes the idea. 😄

LancerComet:
@QingWei-Li 🎈🎅
1. Sorry for doing that, I just want to have a folder to store static files.
2. This feature is intended to be used in local server, I didn't expect to use it on github, what a shame!😂😂

QingWei-Li:
Maybe we should add this feature into [docsify-cli](https://github.com/QingWei-Li/docsify-cli), not here.

But I do not understand why need password in local server.

LancerComet:
@QingWei-Li 
In my case, I run docsify on an open server but I just want people who have password can get access to our docs.

QingWei-Li:
cool!

Move the feature to  docsify-cli, what do you think? We can add a command that enable HTTP authentication. such as https://github.com/zeit/serve#authentication

LancerComet:
@QingWei-Li 
Cool, I will go to check it later 😆

PR #47, opened 2017-01-01
add showcase and edit Development Tips

Rozbo:


QingWei-Li:
You can try it. it is work.
```html
<script src="//unpkg.com/docsify"></script>
<script src="//unpkg.com/prismjs/components/prism-bash.js"></script>
<script src="//unpkg.com/prismjs/components/prism-php.js"></script>
```

PR #132, opened 2017-03-15
fix some minor issues

jrappen:
See commits for details.

> I did **not**:
> * rebuild
> * bump the version
> * update the `yarn.lock` file as I'm not using yarn

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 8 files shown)


jrappen:
@QingWei-Li I was just looking at <https://github.com/QingWei-Li/docsify/blob/master/.gitignore> and upon inspection I thought QingWei-Li/docsify@54da5fb might contain an error, where `"themes"` shouldn't have been removed from `"files"`? If I made a mistake I'm sorry, I'm not fluent in JavaScript.

QingWei-Li:
Yes, we need to keep "themes"

jrappen:
see #133

PR #143, opened 2017-03-30
Properly handle anchor links local within a file. Fix #142

### LINKED ISSUE #142: How to do anchor links within a single page
Hello, really enjoy docsify. Is there any way to be to have anchor links within a single page?

My use case is that I generate markdown API documentation automatically from JSDoc, and would like to be able to just add that as a single page within docsify project / site. The generated markdown automatically generates a "table of contents" that's a listing of all functions and classes of all the API within the project at the top of the file, and the table of contents links to the actual documentation for those that are elsewhere later in the file.

So for example it may end up being something like:

```markdown
<a href="#MyClass">MyClass</a>
...
<a name="MyClass"></a>
### MyClass
```
I can't seem to get this to work even if I try modifying the anchor links or setting `data-id` properties.

Is this possible to accomplish using docsify somehow (even with some hacking or manual manipulation of the generated markdown).

If not perhaps point me in the right direction and I can try and do a PR.

Thanks

bojand:
in toURL test for anchor links and if so prepend the currentPath to t…he generated path. fixes #142.

* [x] DO NOT include files inside lib directory.

### DIFF
--- a/src/core/route/hash.js
+++ b/src/core/route/hash.js
@@ -66,14 +66,14 @@ export function parse (path = window.location.href) {
  * @param  {object} qs   query params
  */
 export function toURL (path, params) {
-  const inline = path[0] === '#';
+  const inline = path[0] === '#'
   const route = parse(replaceSlug(path))
 
   route.query = merge({}, route.query, params)
   path = route.path + stringifyQuery(route.query)
   path = path.replace(/\.md(\?)|\.md$/, '$1')
 
-  if (inline) path = currentPath + path;
+  if (inline) path = currentPath + path
 
   return cleanPath('#/' + path)
 }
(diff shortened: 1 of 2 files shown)


src/core/route/hash.js
```diff
@@ -66,11 +66,14 @@ export function parse (path = window.location.href) {
  * @param  {object} qs   query params
  */
 export function toURL (path, params) {
+  const inline = path[0] === '#';
```
QingWei-Li: Please remove the semi.

bojand:
fixed. sorry i actually don't use semi's myself, but got used to looking at the built version which had semis, and didn't realize the actual code here omits them.

bojand:
I'll have to properly redo this. I didn't realize the code is a bit different than the built / bundled app.

PR #258, opened 2017-09-16
Create vue.rtl.css

mrfarhadir:
adding rtl version for vue template

### DIFF
--- a/src/themes/vue.rtl.css
+++ b/src/themes/vue.rtl.css
@@ -1,0 +1,1011 @@
+@import url('https://fonts.googleapis.com/css?family=Roboto+Mono+Material-Design|Source+Sans+Pro:300,400,600');
+
+*{
+  -webkit-font-smoothing: antialiased;
+  -webkit-overflow-scrolling: touch;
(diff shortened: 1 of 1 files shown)


QingWei-Li:
Hi @mrfarhadir 
sorry for the late reply. I think it only needs to code a little style can solve, do not need to add a new theme. So you do not support each theme, this is too bloated.

mrfarhadir:
@QingWei-Li 
Thanks for your reply, I was started to create rtl version for dark theme too it was almost finished !
ok now I will create a rtl.css to support all themes.
and I have another question which is not related to rtl issue , for better SEO it's good to have ! after # in urls , will you add this feather ?

QingWei-Li:
@mrfarhadir 
I would also like to better support SEO, but unfortunately I can not do so. The current docsify is a SPA project that itself is not friendly to SEO support. Perhaps you can try the SSR feature. Or you can use other static compiled document tools, such as githook.

mrfarhadir:
Thanks You @QingWei-Li

Issue #494, opened 2018-05-19
[FEATURE REQUEST] Add option to generate skip link for screen reader users


kburk1997:
I am currently using docsify to generate organization documentation for an organization on my school's campus. I've noticed that when using a screen reader to test accessibility, it first tabs through the `aside` before going to the main content.

The current standard is to add a [skip link](https://webaim.org/techniques/skipnav/) to the main content.

I've tried adding a skip link manually with no luck, and it appears I have to dig into the docsify code in order to be able to add one.

That being said, an option to generate a skip link will really help with screen reader accessibility for documentation generated through this project.

trusktr:
Sorry that no one replied to you since 2018! The project is being more actively maintained lately.

>  I have to dig into the docsify code in order to be able to add one.

It is currently a little difficult to modify the underlying source that generates the markup. Maybe we should change to a component-based system that encapsulates markup in each component.

But more back on topic, this is a good feature to add. @anikethsaha @jhildenbiddle Possible to add this in v4?

jhildenbiddle:
Based on the [comments](https://github.com/docsifyjs/docsify/pull/495#issuecomment-390524354) in #495, the decision back in 2018 was to implement this feature as a third-party plugin. I'm guessing that this never happened given that a "skip link" plugin isn't listed on [awesome-docsify](https://github.com/docsifyjs/awesome-docsify).

Accessibility is something we should absolutely consider moving forward and do a better job with in v5. Regarding this particular issue, it's important to realize that the linked document describing best practices for "skip links" was last updated on October 23, 2013. Time have obviously changed since then, and so have accessibility best practices. I think we should better understand these best practices before we agree to add skip links by default. If there is something quick and low-risk that we can do for v4, great. If the work is more involved, let's create a PR and target it for v5.

@kburk1997 -- Are you still using docsify for your organization? If so, do you still have a need for "skip link" functionality or have you worked around this issue?

kburk1997:
Apologies for the late reply - I have graduated the university I was attending when I first opened the issue. I have since found another use case for docsify - is there any update on adding a skip link?

jhildenbiddle:
Thanks for creating the issue and for your patience white waiting for it to be addressed, @kburk1997.

I've created a draft PR here: #2253. In that PR you will find a preview link which will allow you to review the changes:

[...]

PR #495, opened 2018-05-19
[Issue #494] New config option skipLink for accessibility

dmed256:

---

- Updated prismjs from 1.9.0 to 1.13.0
- #494 Added `skipLink` config option + updated docs

### DIFF
--- a/src/core/render/index.js
+++ b/src/core/render/index.js
@@ -80,6 +80,11 @@ function renderNameLink(vm) {
   }
 }
 
+function updateSkipLink(vm) {
+  const skipLink = dom.getNode('#skip-link')
+  skipLink.href = `/#${vm.route.path}?id=main`
+}
+
 export function renderMixin(proto) {
   proto._renderTo = function (el, content, replace) {
     const node = dom.getNode(el)
@@ -202,6 +207,8 @@ export function renderMixin(proto) {
   proto._updateRender = function () {
     // Render name link
     renderNameLink(this)
+    // Update skip-link link
+    updateSkipLink(this)
   }
 }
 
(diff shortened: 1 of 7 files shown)


dmed256:
Seems `href="#?id=main"` only works for `/`, not sure how to generalize it

dmed256:
We might need to add an `onclick` event to properly focus on `#main`

src/core/render/index.js
```diff
@@ -251,6 +258,15 @@ export function initRender(vm) {
     dom.before(navAppendToTarget, navEl)
   }
 
+  if (config.skipLink
```
dmed256: I tried adding this to `<body></body>` but `<nav>` was always added first

QingWei-Li:
Yes.

trusktr:
I think accessibility should be in core, right? Isn't what the site "looks" like important, not just for people with eyes? @jhildenbiddle wdyt?

jhildenbiddle:
@trusktr --

Accessibility should absolutely be a concern for the core team, and we can and should do better moving forward. See https://github.com/docsifyjs/docsify/issues/494#issuecomment-647092391 for related feedback.

I'm guessing that the above recommendation was made based on the assumption that adding a skip link meant that the link would be visible on all sites and, since many people may not want that, users could "opt-in" by either installing the plugin or not. Again, just a guess. Fortunately, this isn't the case. We can add a "Skip Navigation" link to the top of the sidebar that is hidden by default but becomes visible when it receives focus.

Let's work through the problem in #494 and see where we land. I think this is a good feature to add and accessibility features like this should be part of docsify core. That said, I'm not sure this is the sort of thing we should add to v4. A significantly improved a11y experience would also be a pretty big achievement for v5.

Issue #679, opened 2018-11-02
Colon (`:`) between 2 words in header disappears


sylwekb:
Hey,

I use a header like this:

```md
### `/autocontrast:3/`
```

The result is:
<img width="190" alt="screen shot 2018-11-02 at 12 36 18" src="https://user-images.githubusercontent.com/2721405/47913315-f1a1a580-de9b-11e8-881a-2f0e22c2c372.png">

And the html looks like this:

<img width="395" alt="screen shot 2018-11-02 at 12 37 06" src="https://user-images.githubusercontent.com/2721405/47913342-09792980-de9c-11e8-8e6f-d57441487dc7.png">

Tried on Firefox and Chrome, in both cases the `:3` is missing. I have the `noEmoji: true,` setting but it seems docsify still tries to parse.

Please help :)

sylwekb:
Hey, thanks for a quick answer, unfortunately it didn't work:

```md
### \`/autocontrast:3/\`
```
Produces:
<img width="234" alt="screen shot 2018-11-02 at 14 41 44" src="https://user-images.githubusercontent.com/2721405/47918657-7fd25780-dead-11e8-95f9-7632baa2a519.png">

cheng-kang:
Ah, sorry, I missed the `:3` part. My guess is that the problem is with config parsing. Because Docsify allows users to configure by adding `:CONFIG_NAME=CONFIG_CONTENT`.

Will check later.

sylwekb:
> ... Docsify allows users to configure by adding `:CONFIG_NAME=CONFIG_CONTENT`.

Thanks! I would be equally happy with a workaround if a proper fix is too complicated ;) Can you point to the documentation which describes this `:config` behavior? For me it looks like it's `alias` https://docsify.js.org/#/configuration?id=alias but i'm not sure

cheng-kang:
This page describes a few configs you can use: https://docsify.js.org/#/helpers?id=customise-id-for-headings

sylwekb:
To everybody who has the same problem. The [regexp is here](https://github.com/docsifyjs/docsify/blob/v4.8.5/src/core/render/compiler.js#L22). The workaround I found is to use other colon: https://unicode-table.com/en/A789/
Normal colon: `:`
Modifier Letter Colon: `꞉`

sylwekb:
Just realized it not good workaround if you  allow users to copy it and use somewhere where the fact that it's a real colon is important ;D

sylwekb:
[...]

PR #680, opened 2018-11-02
#679 Remove only first custom id for a heading instead of all of them.

sylwekb:

### DIFF
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -19,7 +19,7 @@ export function getAndRemoveConfig(str = '') {
     str = str
       .replace(/^'/, '')
       .replace(/'$/, '')
-      .replace(/:([\w-]+)=?([\w-]+)?/g, (m, key, value) => {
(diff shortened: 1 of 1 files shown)


sylwekb:
Actually that would break other modifiers... closing for now

src/core/render/compiler.js
```diff
@@ -19,7 +19,7 @@ export function getAndRemoveConfig(str = '') {
     str = str
       .replace(/^'/, '')
       .replace(/'$/, '')
-      .replace(/:([\w-]+)=?([\w-]+)?/g, (m, key, value) => {
+      .replace(/:([\w-]+)=?([\w-]+)?/, (m, key, value) => {
```
cheng-kang: This is for extracting all possible configs. In the case of headings, we have only one at the moment. 

I guess a better solution would be only match strings at the end. We may need to introduce a delimiter  for determining where the starting point of configs is.

Issue #689, opened 2018-11-06
Feature request: multi-line tips.


linkeo:
It would be great if tips or warns can be writen in multiple lines, like blockquotes. Such as:

```markdown
!> Important things:
!> 1. Foo
!> 2. Bar

?> Tips:
?> - Foo
?> - Bar
```

At this time, if I write like this, I will got some "!>"s or "?>"s displayed.

Issue #413, opened 2018-03-06
Friendly doc Helpers


QingWei-Li:
docsify extends Markdown syntax to make your documents more readable.

```markdown
> [!] **Time** is money, my friend!
```

In Markdown

> [!] **Time** is money, my friend!

In docsify

![image](https://user-images.githubusercontent.com/7565692/37012669-ce73c2ce-2130-11e8-9a4a-813e23fd3204.png)

------

```markdown
> [?] _TODO_ unit test
```

In Markdown

> [?] _TODO_ unit test

In docsify

![image](https://user-images.githubusercontent.com/7565692/37012686-e4028f26-2130-11e8-9704-6e645d51da98.png)

----

## New helpers

```markdown
> [v] Good
```

```markdown
> [x] Bad
```

```markdown
> [detail] summary
> detail detail deatail
> xxxx
```

jhildenbiddle:
Friendly request for code support in `<details>` elements. :smile:

```
> [details] Sample Code
> ```javascript
> console.log('foo');
> ```
```

jhildenbiddle:
Three helper-related items worth mentioning:

### 1. Blockquote helpers should support nested block-level elements

Consider how generic markdown supports multiple paragraphs in a single blockquote:

```markdown
> Paragraph 1
>
> Paragraph 2
```

HTML output:

```html
<blockquote>
  <p>Paragraph 1</p>
  <p>Paragraph 2</p>
</blockquote>
```

Unfortunately, the existing "important content" and "general tips" do not support nested block-level elements properly:

```markdown
?> Paragraph 1
?>
?> Paragraph 2
```

HTML output:

```html
<p class="tip">Paragraph 1
!&gt;
!&gt; Paragraph 2</p>
```

The correct HTML output should be:

```html
<div class="tip">
  <p>Paragraph 1</p>
  <p>Paragraph 2</p>
</div>
```

### 2. Docsify-specific markdown

Markdown is an environment-agnostic format. Markdown used with docsify will likely be viewed, edited, and previewed outside of docsify by site devs and content authors. Consideration should be given to the experience of working with docsify-specific markdown in these environments, not just how it will render in the final docsify site.

From the original [markdown project](https://daringfireball.net/projects/markdown/) page:

> "Markdown is a text-to-HTML conversion tool for web writers. Markdown allows you to write using an easy-to-read, easy-to-write plain text format, then convert it to structurally valid XHTML (or HTML).”

[...]

PR #690, opened 2018-11-06
Support multi-line tips and warns

linkeo:
Add support for multi-line tips and warns. Implement for #689 .

I did not find a way to add custom block elements to marked, so I just modified helper function to reach the aim.

Changes:
1. Because multiple lines may contain multiple paragraphs or other block elements, I changed container to `div`, and modified styles about it (and fix margin of the first or last child in the container).
2. Use RegExp to remove `!>` or `?>` at the beginning of every lines, then use a default marked to parse into html content.
3. Add package "he" to decode html to markdown (inline elements will not be reverted to markdown, but it's equivalent to write inline elements in html form)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3329,6 +3329,11 @@
         }
       }
     },
+    "he": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/he/-/he-1.2.0.tgz",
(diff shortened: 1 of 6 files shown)


QingWei-Li:
Thanks Keo!

BUT I will refactor this feature in v5.0

ref #413

PR #708, opened 2018-11-27
Fix getAndRemoveConfig regex

bashkirtsevich:
Update getAndRemoveConfig regular expression for prevent cases, when header has name like `# foo::bar::baz`. 
We use docsify for perl, our regular case: `### Some::Module`, and without fixed regex we have header like `### Some:`.

Compatible with FireFox.

### DIFF
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -19,7 +19,7 @@ export function getAndRemoveConfig(str = '') {
     str = str
       .replace(/^'/, '')
       .replace(/'$/, '')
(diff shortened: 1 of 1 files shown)


sylwekb:
It is not clear what you have changed in the PR. AFAIU regexp now works only if there is some whitespace before `:` (colon)?. It fixes my issue then, thanks 👍

krohrsb:
This will also fix my issue :) Waiting for it to get published. Thanks!

sylwekb:
@krohrsb You can you my dev build till this time https://uplycdn.com/docs/shEOcKLcylbm/docsify.min.js

Issue #351, opened 2018-01-17
Direct link to sub-item not scrolling to correct position


adrianbj:
I am sure I must have some setting incorrect, but the first time I load this:
https://adrianbj.github.io/TracyDebugger/#/debug-bar?id=module-disabler

it has the "module-disabler" section at the bottom of the page. I browser reload brings it to the top as expected.

However I don't see the problem with the docs page for docsify, eg: https://docsify.js.org/#/configuration?id=load-navbar works as expected on the first load.

Am I doing something incorrectly?

Thanks, and I hope this is the best place to ask for help.

QingWei-Li:
Because there is a picture on your page, the picture height is unknown at first load.

adrianbj:
Any chance you could move the scroll feature code to be triggered after all images have been loaded?

adrianbj:
Hi @QingWei-Li - any chance I can get this back on your radar please?

cheng-kang:
@adrianbj just raised a PR for this issue: https://github.com/docsifyjs/docsify/pull/723

Would be great if you could help testing it 😄

IssuehuntBot:
@issuehuntfest has funded $40.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/74260508/issues/351)

rizdaprasetya:
Took quite a while to wait proposed PR to be merged, I end up creating [a plugin to overcome this issue](https://github.com/rizdaprasetya/docsify-fix-pageload-scroll)
Just include this script tag like the other default Docsify plugins (search, emoji, etc). Put it after Docsify script tag.
```html
<script src="https://cdn.jsdelivr.net/gh/rizdaprasetya/docsify-fix-pageload-scroll@master/index.js"></script>
```

or add this on your `$docsify.plugins` declaration:
```javascript

function (hook, vm) {
    hook.ready(function () {
        // true = show debug log
        let dd = false 
        let TARGET_QUERY = 'id'
        let SCROLL_DELAY = 2000 // in milisecond
        let location = window.location

        dd&&console.log('custom scroll plugin called!')
        let currentUrlWithoutHash = new URL(
            location.origin+location.pathname+
            location.search+location.hash.substring(1)
        )
        let urlQueryParam = currentUrlWithoutHash.searchParams
        let isUrlHasIdQuery = urlQueryParam.has(TARGET_QUERY)
        if(isUrlHasIdQuery){
            dd&&console.log('url has id, will scroll to element')
[...]

PR #723, opened 2018-12-23
fix: scrollToView not working properly if images exist

cheng-kang:
Closes: https://github.com/docsifyjs/docsify/issues/351

-----

### DIFF
--- a/src/core/event/scroll.js
+++ b/src/core/event/scroll.js
@@ -1,5 +1,6 @@
 import {isMobile} from '../util/env'
 import * as dom from '../util/dom'
+import image from '../util/image'
 import Tweezer from 'tweezer.js'
 
 const nav = {}
@@ -123,8 +124,31 @@ export function scrollActiveSidebar(router) {
   })
 }
 
+export let pausedScrollToView
+
+export function proceedScrollToView() {
+  if (pausedScrollToView) {
+    const copy = pausedScrollToView.slice()
+    pausedScrollToView = undefined
+    scrollIntoView(...copy)
+  }
+
+  image.unsubscribe(proceedScrollToView)
+}
+
 export function scrollIntoView(path, id) {
   if (!id) {
+    return
+  }
+
+  if (pausedScrollToView) {
+    pausedScrollToView = [path, id]
+    return
+  }
+
+  if (!image.isAllImagesComplete()) {
+    pausedScrollToView = [path, id]
+    image.subscribe(proceedScrollToView)
     return
   }
 
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -299,7 +299,9 @@ export class Compiler {
         url = getPath(contentBase, getParentPath(router.getCurrentPath()), href)
       }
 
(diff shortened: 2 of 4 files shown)


dawit-elias:
Any updates on this @QingWei-Li? This fix will alleviate the confusion that people who currently read our docs experience.

timaschew:
```sh
git commit --amend --no-edit
git push --force
```

timaschew:
Would be nice to have a test case for the problem which is solved by this PR

dawitelias:
@timaschew There's a repro case at the top of this issue: https://github.com/docsifyjs/docsify/issues/351#issue-289125410.

dawit-elias:
@timaschew There's a repro case at the top of this issue: https://github.com/docsifyjs/docsify/issues/351#issue-289125410.

rp-applied:
Anything preventing this from being merged? Would love this change as well! :)

timaschew:
@rp-applied a test case is still missing in this PR

nnblack1980:
@timaschew - what's the best way to deliver a test case?

nnblack1980:
@timaschew 

Here's a test case. If you hit this URL directly you should be scrolled to the `Test 3` section- but due to the images you see the `Test 2` section. 

https://nnblack1980.github.io/#/?id=test-3

Here's the repo: https://github.com/nnblack1980/nnblack1980.github.io

jhildenbiddle:
Hey folks.

It's worth mentioning that while this PR may fix (I haven't tested it) `scrollToView` behavior with regards to image loading, it does not address the fact that the same issue exists for docsify plugins that modify the DOM (which can alter scroll positions by changing the content area dimensions). This bug was brought to my attention via https://github.com/jhildenbiddle/docsify-tabs/issues/8. As I mentioned in a reply to the issue:

> The short explanation is that docsify begins scrolling to the section specified in the URL before plugins (like docsify-tabs) or external content (image, video, AJAX content, etc.) have loaded. When plugins or external content can affect final page height, they also affect the final scroll position of each section. Docsify fails to account for this and scrolls to the *original* (incorrect) scroll position specified in the URL.
> 
> An actual fix would involve using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) (and [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) as a fallback) to continuously monitor for content height changes, then scrolling (if the section has not already been scrolled to) or jumping (if the section was previously reached) to the section until a user-initiated scroll is detected. Monitoring for content height changes is the only way to detect *all* scroll position changes regardless of how or why they happen. Once a user-initiated scroll is detected, observers used to detect content changes and event listeners used to detect user-initiated scrolling can be removed so there's no left over cruft. Most importantly, this solution allows users to land on a page, have the correct section scrolled into view regardless of how content is loading above the target section.

The above solutions handles all cases (instead of just images) and it seems easier to implement that the image-only fix proposed here.

Thoughts?

kleinjm:
Would it be possible to offer the option to disable the `scrollTo` behavior on page load? That way a custom scroll implementation can be written by those using the library

Issue #681, opened 2018-11-02
Breaking change on v4.8.*


dialex:
[This is my live website](https://dialex.github.io/start-testing/#/). Now it's fine because I've fixed docsify's version on `4.7.1`. However, if I change to `4.8.*` several issues occur:
- The background custom color is ignored
- The style of the button disappears
- The sidebar links break

I've already compared my index.html with the latest template and it looks fine. I have no idea what causes that. Examples below.

4.7.1 (ok):
<img width="467" alt="screenshot 2018-11-02 at 17 20 19" src="https://user-images.githubusercontent.com/2140450/47931308-5eca3080-dec6-11e8-887a-ea8a581ee236.png">

4.8.0 (not ok):
<img width="547" alt="screenshot 2018-11-02 at 17 20 08" src="https://user-images.githubusercontent.com/2140450/47931307-5eca3080-dec6-11e8-95b0-3feecbd4d54c.png">

4.8.0 (broken sidebar):
<img width="298" alt="screenshot 2018-11-02 at 17 35 56" src="https://user-images.githubusercontent.com/2140450/47931309-5eca3080-dec6-11e8-887c-6dca04b30eb1.png">

Thoughts?

cheng-kang:
Please update to `4.8.3` and see if the issues stay.

dialex:
Yup, same broken behaviour 😢

cheng-kang:
According to https://marked.js.org/#/README.md#specifications`marked` is following [CommonMark](https://spec.commonmark.org/0.28/). Space in link is not supported as per [the specs](https://spec.commonmark.org/0.28/#link-destination).

Alternatively, for now, you could use `<img data-origin="linear-gradient(to left bottom, #F0FD37 0%, #ACF260 100%)" alt="color">`; so that you can still use latest version and have all other fixes applied to your site.

@QingWei-Li I guess we need to make a decision on whether or not to add support for space in link in Docsify.

QingWei-Li:
@cheng-kang I think we should follow the rules. 
But to be compatible with using space characters in custom background color.

@dialex About the sidebar, You should split the line between the html tag and the markdown content. I don't think this is a bug in docsify, it will also render errors on github.

![image](https://user-images.githubusercontent.com/7565692/48108999-a82cce00-e280-11e8-8329-74d1f4aa6b7e.png)

dialex:
[...]

PR #724, opened 2018-12-24
Update the color example

dialex:
...because the old way is broken, see https://github.com/docsifyjs/docsify/issues/681#issuecomment-437459116

### DIFF
--- a/docs/cover.md
+++ b/docs/cover.md
@@ -54,7 +54,7 @@ The background color is generated randomly by default. You can customize the bac
 
 <!-- background color -->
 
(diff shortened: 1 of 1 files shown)


cheng-kang:
Thanks for the PR, but "the new way" is just a workaround before the bug gets fixed.

Could you try the latest version? It should have been fixed.

dialex:
@cheng-kang I tried the latest version (4.8.6) and...

`color` ✅displays red
`color)` ❌it displays a random color

So it's _half fixed_, you're close!

PR #735, opened 2019-01-05
Add new theme "dolphin"

KsRyY:
Dolphin is a blue theme based on the theme vue, using the font [Thasadith](https://fonts.google.com/specimen/Thasadith) as default.


dialex:
How can I use this theme? There's no example in the official doc https://docsify.js.org/#/themes

Issue #612, opened 2018-08-24
Unexpected autoscroll


bashkirtsevich:
Please, see this issue: https://github.com/docsifyjs/docsify-cli/issues/42

PR #744, opened 2019-01-14
feat(scroll): smarter scroll behavior

solymosi:
Currently Docsify has the following behavior when it comes to scrolling:

* if `auto2top` is **enabled** then:
  - _any_ change to a URL that includes an ID will scroll to that ID once loaded
  - _any_ change to a URL without an ID always scrolls to the top once loaded
  - therefore the browser's scroll restoration behavior is always nullified when moving back/forward in history or reloading the page
* if `auto2top` is **disabled** then:
  - _any_ change to a URL that includes an ID will scroll to that ID once loaded
  - moving back/forward in history and reloading the page uses the browser's scroll restoration behavior, but only if the target does not have an ID
  - clicking a link to navigate to a page that does not include an ID maintains the current scroll position and usually ends up somewhere in the middle of that page (fixes https://github.com/docsifyjs/docsify-cli/issues/42; fixes https://github.com/docsifyjs/docsify/issues/612)

Ideally, the following should happen (in my opinion):

* moving back/forward in history and reloading the current page should not invoke any kind of scrolling behavior and therefore should _not_ override the browser's scroll restoration behavior
* clicking a link to navigate to a page should scroll to the top of that page or to an ID if specified

This PR implements this (in my opinion better) behavior for both the `hash` and the `history` router modes. Scrolling to the top only happens if `auto2top` is enabled.

It does _not_ attempt to fix the inconsistent scrolling issue that's caused by images loaded asynchronously.

* [X] Make sure you are merging your commits to `master` branch.
* [X] Add some descriptions and refer relative issues for you PR.
* [X] DO NOT include files inside `lib` directory.

### DIFF
--- a/src/core/event/index.js
+++ b/src/core/event/index.js
@@ -1,11 +1,26 @@
 import { isMobile } from '../util/env';
 import { body, on } from '../util/dom';
 import * as sidebar from './sidebar';
-import { scrollIntoView } from './scroll';
+import { scrollIntoView, scroll2Top } from './scroll';
 
 export function eventMixin(proto) {
-  proto.$resetEvents = function() {
-    scrollIntoView(this.route.path, this.route.query.id);
+  proto.$resetEvents = function(source) {
+    const { auto2top } = this.config;
+
+    (() => {
+      // Rely on the browser's scroll auto-restoration when going back or forward
+      if (source === 'history') {
+        return;
+      }
+      // Scroll to ID if specified
+      if (this.route.query.id) {
+        scrollIntoView(this.route.path, this.route.query.id);
+      }
+      // Scroll to top if a link was clicked and auto2top is enabled
+      if (source === 'navigate') {
+        auto2top && scroll2Top(auto2top);
+      }
+    })();
 
     if (this.config.loadNavbar) {
       sidebar.getAndActive(this.router, 'nav');
--- a/src/core/fetch/index.js
+++ b/src/core/fetch/index.js
@@ -148,7 +148,10 @@ export function fetchMixin(proto) {
     }
   };
 
-  proto.$fetch = function(cb = noop) {
+  proto.$fetch = function(
+    cb = noop,
+    $resetEvents = this.$resetEvents.bind(this)
+  ) {
     const done = () => {
       callHook(this, 'doneEach');
       cb();
@@ -160,7 +163,7 @@ export function fetchMixin(proto) {
       done();
     } else {
       this._fetch(() => {
-        this.$resetEvents();
+        $resetEvents();
         done();
       });
     }
--- a/src/core/render/index.js
+++ b/src/core/render/index.js
@@ -6,7 +6,7 @@ import { getAndActive, sticky } from '../event/sidebar';
 import { getPath, isAbsolutePath } from '../router/util';
 import { isMobile, inBrowser } from '../util/env';
 import { isPrimitive } from '../util/core';
-import { scrollActiveSidebar, scroll2Top } from '../event/scroll';
+import { scrollActiveSidebar } from '../event/scroll';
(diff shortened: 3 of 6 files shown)


timaschew:
```sh
git commit --amend --no-edit
git push --force
```

solymosi:
@timaschew Done.

timaschew:
Hey @solymosi,
Thank you for your contribution!

FYI: the DNS for the preview has changed: https://deploy-preview-744--docsifyjs.netlify.com

> Ideally, the following should happen (in my opinion):
>
> * moving back/forward in history and reloading the current page should not invoke any kind of scrolling behavior and therefore should not override the browser's scroll restoration behavior
> * clicking a link to navigate to a page should scroll to the top of that page or to an ID if specified

I totally agree with you.

I've just tried some actions:
1. Open [Plugins page](https://deploy-preview-744--docsifyjs.netlify.com/#/plugins?id=pagination)
2. Click in the sidebar on the next page: **Write a Plugin**
3. Go back using browser history

Expected: Scroll position at **pagination**  
Actual: Scroll position is between **emoji** and **External Script**

When using `history` mode it works fine.

Also I don't understand how **auto2top: false** can be useful. For me it's broken because if you open another page the scroll position is only correct if you didn't scroll on the previous page, which is mostly not the case. Or do I miss anything? (This is of course not directly related to your PR)

solymosi:
> I've just tried some actions:
> 
>   1. Open [Plugins page](https://deploy-preview-744--docsifyjs.netlify.com/#/plugins?id=pagination)
>   2. Click in the sidebar on the next page: **Write a Plugin**
>   3. Go back using browser history

Hmm... that looks like a bug, and it's reliably reproducable when:

* the page being navigated away from is _shorter_ than the previous page being navigated back to;
* and the target location on the previous page is lower than the length of the current page.

I _guess_ the reason is that the browser's scroll restoration behavior is invoked _before_ the old page is properly replaced with the new one by Docsify and therefore the target scroll position is capped at the height of the old page. The fact that it's only triggered when using the `hash` mode suggests that browsers treat `popstate` and `hashchange` events differently in this regard.

The problem is that this may require implementing our own custom scroll restoration behavior and turning off the browser's default: https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration. But I'd need to do some digging first because there might be a simpler workaround as well.

In any case, I'd recommend addressing this particular issue in a different PR.

> Also I don't understand how **auto2top: false** can be useful. For me it's broken because if you open another page the scroll position is only correct if you didn't scroll on the previous page, which is mostly not the case. Or do I miss anything? (This is of course not directly related to your PR)

I have no idea how it's useful either, but since it existed I assumed there was at least _some_ reason for it, so I kept it in there. Totally fine with removing it, however.

timaschew:
Merging this PR before the other one, actually let's say publishing a new version based on this PR would be a regression. Because currently it's working fine for URLs with an ID.

 Regarding `auto2top: false` I suggest to remove this in 5.0

/cc @jhildenbiddle

anikethsaha:
cc @solymosi whats the status on this ?
 @timaschew

solymosi:
Let me rebase this and resolve the conflicts. After that it can be merged, unless you'd like me to make additional changes.

solymosi:
Rebased and fixed up the whitespace style to match that of the `develop` branch.

anikethsaha:
I will review it soon

cc @timaschew your thoughts ?

solymosi:
Rebased again due to merge conflicts. Any updates on the review?

anikethsaha:
can you please fix the CI issue.

solymosi:
Linting issues fixed ✔️  &ndash; couldn't do anything about the "security flow", that step seems to be broken.

PR #751, opened 2019-01-18
fix: handler of A link in history router mode

imyelo:
`<A>` link that is nested at multilayers cannot be intercepted under
history router mode.

e.g.:
```html
<a>
  <div>
    <span>TEXT</span>
  </div>
</a>
```

We should use `closest` function to find the `<A>` link in this case.

-----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -918,6 +918,14 @@
       "integrity": "sha1-2jCcwmPfFZlMaIypAheco8fNfH4=",
       "dev": true
     },
+    "closest": {
+      "version": "0.0.1",
+      "resolved": "https://registry.npmjs.org/closest/-/closest-0.0.1.tgz",
+      "integrity": "sha1-JtpvgLPg4X5x+A8SeCgZ6fZTSVw=",
+      "requires": {
+        "matches-selector": "0.0.1"
+      }
+    },
     "cmd-shim": {
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/cmd-shim/-/cmd-shim-2.0.2.tgz",
@@ -4702,6 +4710,11 @@
       "version": "0.5.1",
       "resolved": "https://registry.npmjs.org/marked/-/marked-0.5.1.tgz",
       "integrity": "sha512-iUkBZegCZou4AdwbKTwSW/lNDcz5OuRSl3qdcl31Ia0B2QPG0Jn+tKblh/9/eP9/6+4h27vpoh8wel/vQOV0vw=="
+    },
+    "matches-selector": {
+      "version": "0.0.1",
+      "resolved": "https://registry.npmjs.org/matches-selector/-/matches-selector-0.0.1.tgz",
+      "integrity": "sha1-HfUmIkOuNBwaCATdMCBIJnrHE7s="
     },
     "math-expression-evaluator": {
       "version": "1.2.17",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "postinstall": "opencollective postinstall"
   },
   "dependencies": {
+    "closest": "0.0.1",
     "marked": "^0.5.1",
     "medium-zoom": "^0.4.0",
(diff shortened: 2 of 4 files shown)


timaschew:
```sh
git commit --amend --no-edit
git push --force
```

timaschew:
the module has a dependency which doesn't work in Node.js even if the source on GitHub works, but it seems that it was not published to npm.

Funny fact there is an alternative module with the same issues. But fortunately I have npm permission to publish, so this module works in Node.js: https://www.npmjs.com/package/component-closest

Be careful, the API is a bit different (the last argument)

jhildenbiddle:
Do we really need to add a dependency for this? Why not just bake a [polyfill](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Polyfill) into docsify core so it can be easily removed once docsify drops IE support?

timaschew:
So true. We need it only for IE. So maybe some day we can even remove it.

imyelo:
> The **closest** module has a dependency which doesn't work in Node.js even if the source on GitHub works, but it seems that it was not published to npm.
> 
> Funny fact there is an alternative module with the same issues. But fortunately I have npm permission to publish, so this module works in Node.js: https://www.npmjs.com/package/component-closest
> 
> Be careful, the API is a bit different (the last argument)

Ah! I did want to use component-closest, that I have used in [other projects](https://github.com/imyelo/docsify-pagination) before. Looks like I got mixed up.

jhildenbiddle:
@timaschew --

I was also referring to closest.

I don't think we should be adding dependencies for third-party libraries that replace native methods unless there is a very good reason to do so. For example, consider the differences in usage for the native `Element.closest()` method vs. [component-closest](https://www.npmjs.com/package/component-closest):

```js
// Native
elm.closest('p');

// component-closest
closest(elm, 'p');
```

If we adopt component-closest and use its syntax throughout the codebase, it becomes more difficult to remove when we drop IE11. Also, those contributing to docsify may be unaware of the component-closest dependency and use the native Element.closest() method instead which could throw an error in IE if uncaught. By polyfilling Element.closest() instead of introducing a new library that accomplishes the same thing, we avoid both of these issues.

On a separate note, we can keep things nice and tidy by separating these kinds of ponyfills into a separate `legacy.js` file (or something with a similar name) and conditionally loading them based on the current browser. When the time comes to drop IE support, we simply remove this file.

jhildenbiddle:
Another idea...

What about conditionally loading polyfills from https://polyfill.io? This might allow us to keep polyfill out of the docsify repo and still support legacy browsers like IE11 with little effort. Docsify wouldn't be as performant since it would require yet-another request, but maybe this isn't that big of an issue given IE11's market share.

@timaschew -- what do you think?

imyelo:
> @timaschew --
> 
> I was also referring to closest.
> 
> I don't think we should be adding dependencies for third-party libraries that replace native methods unless there is a very good reason to do so. For example, consider the differences in usage for the native `Element.closest()` method vs. [component-closest](https://www.npmjs.com/package/component-closest):
> 
> ```js
> // Native
> elm.closest('p');
> 
> // component-closest
> closest(elm, 'p');
> ```
> 
> If we adopt component-closest and use its syntax throughout the codebase, it becomes more difficult to remove when we drop IE11. Also, those contributing to docsify may be unaware of the component-closest dependency and use the native Element.closest() method instead which could throw an error in IE if uncaught. By polyfilling Element.closest() instead of introducing a new library that accomplishes the same thing, we avoid both of these issues.
> 
> On a separate note, we can keep things nice and tidy by separating these kinds of ponyfills into a separate `legacy.js` file (or something with a similar name) and conditionally loading them based on the current browser. When the time comes to drop IE support, we simply remove this file.

I agree with you @jhildenbiddle ; that is why in `src/core/router/history/html5.js` I did not refer to dependencies directly but instead was forwarded by `src/core/util/dom`. Adding dependencies is not a good idea if Docsify has taken care of the closest method.

PR #774, opened 2019-02-16
Fix github assets url

louislivi:


timaschew:
@docsifyjs/core I vote to merge this as a quick fix.

PR #777, opened 2019-02-20
Update emoji (from official repo github/gemoji)

up9cloud:
* [x] Make sure you are merging your commits to `master` branch.
* [x] Add some descriptions and refer relative issues for you PR.
* [x] DO NOT include files inside `lib` directory.

The original aliases are from [All-Github-Emoji-Icons](https://github.com/scotch-io/All-Github-Emoji-Icons).
It's too old, and not include all the emojis.
So i did newest from [official](https://github.com/github/gemoji), see diff package.json and build/emoji-aliases.js

And, please merge https://github.com/docsifyjs/docsify/pull/774 first.
The assert url is broken.

PR #778, opened 2019-02-21
Generate GitHub emoji aliases from original source

timaschew:
Generate GitHub emoji aliases from original source

see https://github.com/docsifyjs/docsify/pull/777


timaschew:
This should fixed by https://github.com/docsifyjs/docsify/issues/779 which would be even a better improvement.

timaschew:
Closing due to #779

PR #782, opened 2019-02-22
Fixed iframe overflow bug for mobile Safari

### LINKED ISSUE #781: iframe overflows in mobile Safari
On mobile Safari, 100% width iframes can have horizontal overflow from their parent container. This is a quirky Safari thing, but should be pretty easy to account for in the theme styling.

Below is a screenshot of a Codepen embed exhibiting the behavior.

image

trevoreyre:
On mobile Safari, 100% width iframes can have horizontal overflow from their parent container. This modification to the base theme fixes it.

Fixes #781

### DIFF
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -323,6 +323,7 @@ body.sticky
 
 .markdown-section iframe
   border 1px solid #eee
+  /* fix horizontal overflow on iOS Safari */
   width 1px
   min-width 100%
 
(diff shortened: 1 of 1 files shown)


src/themes/basic/_layout.styl
```diff
@@ -323,6 +323,8 @@ body.sticky
 
 .markdown-section iframe
   border 1px solid #eee
+  width 1px
+  min-width 100%
```
timaschew: Hi Trevor,
thank you for your contribution.

Do you mind to add a comment to explain that this is a fix for iOS safari.

PR #784, opened 2019-02-25
fix: re-render gitalk when router changed

spiritree:


timaschew:
Hi @spiritree  
thanks for you contribution.

Can you explain why it's required to render this plugin again?  
Maybe other plugins are affected as well?  
In that case the logic should be extracted into a common helper function or something like that.

@QingWei-Li any feedback?

spiritree:
@timaschew 
The re-render logic is similiar with disqus plugin(https://github.com/docsifyjs/docsify/blob/master/src/plugins/disqus.js).

There is no need to be extracted into a common helper function because it's a unique plugin.

timaschew:
Okay, but why is it required to render it again actually?

spiritree:
@timaschew 
This plugin is a comment component based on Github issue.
So when router changed,you must re-render it to create unique issue.

Q-Angelo:
reference: https://www.nodejs.red/#/tools/docsify

PR #801, opened 2019-03-29
add scoring and sort to search plugin

spion:
Updates the search plugin and adds scoring to search results. Search results that contain multiple keywords are scored higher than those that contain only one and appear first on the result list.

Headings add score +3, content matches +2. How will that work?
* For two keywords, a heading containing only one will be scored lower than body text containing both. But if a heading contains both, it will score higher than body text containing both.
* For 3 keywords, a heading containing 2 of the 3 will score the same as body text containing all 3
* For 4 keywords, a heading containing 3 of the 4 will score higher than body text containing all 4

------

Before:
image

------

After:
image

### DIFF
--- a/src/plugins/search/search.js
+++ b/src/plugins/search/search.js
@@ -96,29 +96,27 @@ export function search(query) {
 
   for (let i = 0; i < data.length; i++) {
     const post = data[i]
-    let isMatch = false
+    let matchesScore = 0
     let resultStr = ''
     const postTitle = post.title && post.title.trim()
     const postContent = post.body && post.body.trim()
     const postUrl = post.slug || ''
 
-    if (postTitle && postContent) {
-      keywords.forEach(keyword => {
+    if (postTitle) {
+      keywords.forEach( keyword => {
         // From https://github.com/sindresorhus/escape-string-regexp
         const regEx = new RegExp(
           keyword.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&'),
           'gi'
-        )
+        );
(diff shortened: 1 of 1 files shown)


spion:
p.s. we could also use tf-idf (term frequency inverse document frequency) for the scoring scheme, so that instead of every word counting the same, they have a score

`numberOfOccurancesInDoc / numberOfDocsWhereItOccurs`

timaschew:
```sh
git commit --amend --no-edit
git push --force
```

spion:
done

spion:
bump?

src/plugins/search/search.js
```diff
@@ -96,29 +96,27 @@ export function search(query) {
...
     const postContent = post.body && post.body.trim()
     const postUrl = post.slug || ''
 
-    if (postTitle && postContent) {
-      keywords.forEach(keyword => {
+    if (postTitle) {
+      keywords.forEach( keyword => {
         // From https://github.com/sindresorhus/escape-string-regexp
-        const regEx = new RegExp(
+        let regEx = new RegExp(
```
timaschew: Why you changed it to `let`?

spion: No reason, I don't personally buy the modern const vs let usage so my usual let usage slipped up in this code. Changing.

timaschew: >> edit: done. Should I squash the commit?

Yes, please.

spion: done

PR #866, opened 2019-06-08
Fix #791

### LINKED ISSUE #791: Multiple embedded files can break rendering
Under certain circumstances the rendering of pages breaks in the presence of multiple embedded elements. As far as I can tell this is related to wrong assumptions in the preprocessing of the markdown tokens.

__Disclaimer__: My JavaScript is not really great, so my analysis below could be off and is most definitely incomplete.

This could be related to https://github.com/docsifyjs/docsify/issues/718 and https://github.com/docsifyjs/docsify/issues/710 .

# Short version

The condition
```javascript
if (++count >= step) {
```
in [line 46 of embed.js](https://github.com/docsifyjs/docsify/blob/master/src/core/render/embed.js#L46) works under the assumption that the loop has finished before any of the [constructed callbacks](https://github.com/docsifyjs/docsify/blob/master/src/core/render/embed.js#L16-L50) are called. This is not true under several conditions (for example due to the caching behavior in [ajax.js](https://github.com/docsifyjs/docsify/blob/master/src/core/fetch/ajax.js#L19-L21)), which results in broken rendering for pages with embedded elements.

This can be fixed by replacing the condition:
```javascript
if (++count >= embedTokens.length) {
```

# Long version

## Problem description

In my concrete example I set up two separate pages with several embedded code fragments, with fragments from one particular code file being presented on both. When rendered individually (i.e. after a reload) they look fine. However, when switching to the other page e.g. via the menu that page would contain now other elements but some of the fragments (no text, headings, etc.).

Interestingly, this behaviour is symmetrical: reloading one page and switching to the other breaks rendering of the second one.

## Possible cause

My investigation lead me to the preprocessing for the embedded elements in [embed.js](https://github.com/docsifyjs/docsify/blob/master/src/core/render/embed.js). As far as I can tell the condition in [line 46](https://github.com/docsifyjs/docsify/blob/master/src/core/render/embed.js#L46)
```javascript
if (++count >= step) {
```
is meant to check, if we've handled all elements. The assumption seems to be, that the constructed callbacks will only be called __after__ the loop has ended and `step` is equal to the number of tokens. The assumption checks out if the callback is scheduled/queued (for example as part of an AJAX call). However, due to caching in [ajax.js](https://github.com/docsifyjs/docsify/blob/master/src/core/fetch/ajax.js#L19-L21) for some elements the callback is actually not scheduled but executed immediately, triggering the condition early. This happened in my particular case, with the same code file being fetched on two different pages.

I think this particular bug also affects other types of embedded elements, as for all of these the callback is never scheduled but called immediately.

Sadly, my grasp on the project structure is not good enough yet to understand how signalling the end of the iteration early is producing the wrong results.

## Possible fix

In a local copy I was able to fix the issue by modifying the condition in [line 46 of embed.js](https://github.com/docsifyjs/docsify/blob/master/src/core/render/embed.js#L46). Instead of comparing `count` with `step` (relying on the loop being completed before any of the callbacks are called) we could just compare it with `embedTokens.length`:
```javascript
if (++count >= embedTokens.length) {
```

I don't mind preparing a PR for this, but since I don't fully trust my understanding of the code, I would appreciate a second opinion on this first.

aenniw:
Signed-off-by: Martin Mihálek <aenniw@gmail.com>

**Summary**

Fixed out of orded embedding of `include` files

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome 74.0.3729.169
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

---


aenniw:
Addresses #791

anikethsaha:
@aenniw please rebase and resolve the conflicts.
also change the commit message. make it more detailed in order to fit better for changlog

anikethsaha:
can you resolve the conflicts

aenniw:
Rebased, however not sure whats `Security Flow` about...

aenniw:
Not sure how to approach the tests, as It occurs only when embedded files are downloaded out of order mainly due their size difference and network delay... 
On local server setup despite big difference in embedded file sizes only way how to partially reproduce the issue is with throttling in browser...

aenniw:
Added e2e scenario and updated `live-server` to delay serving of one of embedded files to replicate the issue...

cypress/support/commands.js
```diff
@@ -26,9 +26,8 @@
 
 import { addMatchImageSnapshotCommand } from 'cypress-image-snapshot/command'
 addMatchImageSnapshotCommand({
-  failureThreshold: 10.0,
+  failureThreshold: 0.10,
```
anikethsaha, line 29: `10` is quite big one to accept ! any speicific reason to back this ?
can we try with anything `<3` or `<5` ? WDYT

docs/_media/delayed.md
```diff
@@ -0,0 +1,7 @@
+- This is from the `delayed.md`
+
+```bash
+sleep 1000
+echo delayed
+exit 0
+```
```
anikethsaha, line 7: not sure why these files ?

can we add a note that this is just an example ?

docs/_media/non-delayed.md
```diff
@@ -0,0 +1,7 @@
+- This is from the `no-delay.md`
+
+```bash
+sleep 0 
+echo non-delayed
+exit 0
+```
```
anikethsaha, line 7: same as above ?

cypress/support/commands.js
aenniw: 10.0 is in real 10000% thus no 2e2 test could ever fail via diff... see [cypress-image-snapshot/issue/121](https://github.com/palmerhq/cypress-image-snapshot/issues/121)

docs/_media/delayed.md
aenniw: I needed 2 files that will occupy a lot of space so that it's clear that the order of emmbedding is wrong...

docs/_media/non-delayed.md
aenniw: Yep, mandatory part is `- This is ...` the bash is just filler for diff mismatch boost...

docs/embed-files.md
```diff
@@ -53,6 +53,17 @@ Example:
...
+
+[delayed](_media/delayed.md ':include')
+
+---
+
+[non-delayed](_media/non-delayed.md ':include')
+
+You can check the original content for [delayed.md](_media/delayed.md ':ignore'), [non-delayed.md](_media/non-delayed.md ':ignore').
 
 ## Tag attribute
```
Koooooo-7, line 69: what we need is just telling users we will render embedded file in order by default. 
and I think it should do like this logically.

anikethsaha:
I agree with @Koooooo-7 , we cant show testing fixture in docs. 
@aenniw can you make those files just for the testing. try added script [`here`](https://github.com/docsifyjs/docsify/blob/develop/cypress/setup.js).
that file is copy-pasting content from the original docs. so after the copy-pasting, you can add function to create a file and write the content as you need or in some other way whichever you prefer. 
I know its a bit of extra work but we don't have strong test. and the issue addressing in this PR is quite common we noticed in chats. so it would be great to have a good setup and it will help in reviewing this better

docs/embed-files.md
aenniw: reverted, block is inlined by cypress post fs sync

aenniw:
Extracted test specific blocks from docs, updated cypress `setup.js` to inline content before predefined `tag`

jthegedus:
Looking good

aenniw:
CI fixed by updating reference img as new lines were added to docs via [PR](https://github.com/docsifyjs/docsify/pull/1156/files) and it wasn't updated...

trusktr:
Unfortunately (or fortunately, depending on your view) the test system changed from cypress to playwright, making a lot of conflicts in the PR. We need to update this to playwright.

trusktr (changes requested):
needs conflicts resolved, moving from cypress to playwright.

PR #913, opened 2019-08-22
buble theme: strong font-weight increased to 600

xmedeko:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/src/themes/buble.styl
+++ b/src/themes/buble.styl
@@ -35,9 +35,12 @@ $sidebar-width = 16rem
   color #333
   font-weight 400
 
+.markdown-section strong
+  color #333
+  font-weight 600
+
 .markdown-section a
   color var(--theme-color, $color-primary)
-  font-weight 400
 
 .markdown-section p, .markdown-section ul, .markdown-section ol
   line-height 1.6rem


anikethsaha:
Is there any specific issue or link which might support this change ?
I need some reason to back this changes.

Thanks

xmedeko:
@anikethsaha The bug is, that the bold font is not recognizable from the normal fint in the buble theme. I do not know what kind of link should I provide. Should I make one more bug issue about that?

anikethsaha:
@xmedeko actually the optimal flow is to have an issue and then discussing there we can decide whether its a bug or not and PR is required or not.

Leave it for this one. No need to create an issue. We can discuss this here just for this PR.

Also, can you provide some screenshots comparing `before-after` .

src/themes/buble.styl
```diff
@@ -33,7 +33,7 @@ $sidebar-width = 16rem
 
 .markdown-section h1, .markdown-section h2, .markdown-section h3, .markdown-section h4, .markdown-section strong
   color #333
-  font-weight 400
+  font-weight 600
```
anikethsaha: The default of these is `bold` in the browser. I am pretty sure this is an intended code.

xmedeko:
If you make markdown bold `**sometext**. i.e. `<strong>` in HTML, then the text is almost no different to other, normal text in the buble theme. So it's unusable.  I have taken the code from other themes: doplhin, vue, ... see the similar code there.

It's possible the fix is not well for this theme. It should affect headings also, but the buble theme has just a few lines bellow:
```styl
.markdown-section a
  ...
  font-weight 400
```

I think the theme has a few more troubles, but I like it. I am not so strong in CSS. Since the docsify has been kind of unmaintained with lot of open issues, I have decided to post this hot fix PR instead of opening one more issue.

If you can make a better fix of buble theme, then go ahead (also not wrong spelling of "buble").

anikethsaha:
https://github.com/docsifyjs/docsify/pull/913#discussion_r349230930

Actually thats where we are struggling a bit. The former maintainers I guess are not active and not maintaining so we are struggling to get the indentions of these few codes. Like this one. I am not sure why at first place these where `400` as these are not the `strong` styling (as you said too)
I am actually +1 for this fix but not getting the motive of keeping it `400` at first place.

I think we will go with this PR and if required we will revert this if it is suggested.

I am down to `600` or `700`. can you please investigate if anything is breaking or looking weird with having `700` as that's the default by browsers

xmedeko:
Since the doplhin and vue are using 600, I vote for 600. I thing the reason is to be kind of lighter feeling. Also, because both these styles has same value for  `.markdown-section a`, I may amend the commit or do it by yourself.
Thanks.

anikethsaha:
> Since the doplhin and vue are using 600, I vote for 600

Ok let's go with `600`.

xmedeko:
Hold on, I've been playing with that a bit and seems to me the 400 suites the headings and links better in this theme. Maybe separate strong style could be better.

anikethsaha:
> Hold on, I've been playing with that a bit and seems to me the 400 suites the headings and links better in this theme. Maybe separate strong style could be better.

separate the strong from this group 👍

xmedeko:
Ok, will do it tomorrow.

anikethsaha:
ohh shoot!!
I forgot, can you please submit PR against the `master` branch

xmedeko:
@anikethsaha I've separated strong and set it to 600. I've removed the `a { font-weight: 400}`, because it has overridden the headings h1, ...  font-weight below.

(Links has blue color in the buble theme, so they do not need the slight font-weight increase. I also do not know how to keep font-weight and do not override headings, because headings are links to. I aware of CSS `!important`, but I think this solution is better for this situation.)

I've rebased to master. You should set the default branch to master from develop.

anikethsaha:
> You should set the default branch to master from develop.

We are discussing that. Actually `develop` is for v5 but there were quite a good number of bugs that needed to be fixed before v5 so we are working on final v4 release.
So we need to have that discussing then we can change the default branch. 
So fast merging I would recommend submitting a new PR against master

xmedeko:
Ok, closing this PR.

PR #964, opened 2019-12-01
fix: allow link targets to override externalLinkTarget

transitive-bullshit:
**Summary**

Fix a bug where if you set `externalLinkTarget`, manual link overrides no longer take effect.

This is a simple logic error where if `config.target` exists on a link, it will get overridden by the global `linkTarget` as things currently stand.

To reproduce the issue, set `externalLinkTarget` to `_self` and change an external link to use the inline `:target=_blank`.

What should happen is that the inline target overrides the global default, but prior to this PR that inline link would still be set to `_self`.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome

### DIFF
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -227,7 +227,7 @@ export class Compiler {
       return `<pre v-pre data-lang="${lang}"><code class="lang-${lang}">${hl}</code></pre>`
     }
     origin.link = renderer.link = function (href, title = '', text) {
-      let attrs = ''
+      const attrs = {}
 
       const {str, config} = getAndRemoveConfig(title)
       title = str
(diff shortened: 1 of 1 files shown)


src/core/render/compiler.js
```diff
@@ -241,24 +241,32 @@ export class Compiler {
...
         href = router.toURL(href, null, router.getCurrentPath())
-      } else {
-        attrs += href.indexOf('mailto:') === 0 ? '' : ` target="${linkTarget}"`
+      } else if (href.indexOf('mailto:') < 0) {
+        attrs.target = `"${linkTarget}"`
       }
 
       if (config.target) {
-        attrs += ' target=' + config.target
+        attrs.target = `"${config.target}"`
```
transitive-bullshit: the important part here is to override the target instead of appending it and having one arbitrarily chosen.

anikethsaha:
How this changes are helping ? 
What I can see is that, these are doing the exact same thing but with object.
And if perf view, this PR introduces extra work of converting the object -> string.
looping of object keys, and object assignments are also generally slower (I guess !!) than string concatenation. 

Is there anything else I am missing ?

trusktr:
Seems like a good change. I wouldn't worry about the perf difference unless it actually make a difference.

anikethsaha:
sorry @trusktr @transitive-bullshit But I am not in total support for this PR.
May be you can give a better point to reopen this.

again sorry , and thanks for your time @transitive-bullshit

PR #986, opened 2019-12-23
[WIP] test(e2e): added jest env and puppeteer for e2e

anikethsaha:
## E2E testing using puppeteer and jest

- [x] added puppeteer
- [x] removed mocha
- [x] added jest
- [x] created custom jest env for supporting the live-server instance in tests suites
- [x] npm script support
- [x] eslint fixes ( #987 )
- [ ] test cases using puppeteer
  - [x] setup image snapshot using [`jest-image-snapshot`](https://github.com/americanexpress/jest-image-snapshot#readme)
  - [ ] sideBar testing
  - [x] search field testing
  - [x] cover page
- [x] fix jest config to support running other tests in `test` folder
- [x] travis additions
- [ ] coverage (nyc or jest and codecoverage for PRs)
- [x] Github action if required

Currently, This PR contains a fixture which is a copy of our docs. So these tests will be run against the updated docs with updated themes and plugins.
Refactored the `e2e/fixtures/docs/index.html` in order to use packages from local project

**This PR doesnt contain SSR tests**

More tests and fixtures can be added in later PR and in future.

## Update/Note

> Currently I have assigned `failureThreshold` to `10%` but in future I think we should consider `<1%`

> Test timing is currently `30sec` , may consider to increase it if required `max 60sec`


e2e/__tests__/index.spec.js
```diff
@@ -0,0 +1,30 @@
+import puppeteer from 'puppeteer'
+import { toMatchImageSnapshot } from 'jest-image-snapshot'
+var browser
+var page
```
jamesgeorge007: Why not using `let`?

anikethsaha: Will change it !!!

_Note a fan of `let` tough_

PR #1017, opened 2020-02-10
update: .markdown-section max-width 800px to 80%

jhoneybee:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

---


anikethsaha:
@jhoneybee can you please describe the reason or any point for this change

jhoneybee:
@anikethsaha  

The reason to modify 800px to percentage is as follows

#### 1. The document is not beautiful under the huge screen.

800px

image

80%

image


> I found 80% is the most beautiful after trying . sorry, because English is not my native language, I answered late

yzhang-gh:
I would suggest restoring `max-width` to `100%` if the viewport width is less than 768px.
https://github.com/docsifyjs/docsify/blob/bea19edaa8f525f2a3b9ec547be840eb0e9528da/src/themes/basic/_layout.styl#L406

This PR breaks the layout on mobile devices 😥.

anikethsaha:
can you create an issue

yzhang-gh:
I created a PR (#1273) as it is so slight a problem.

yzhang-gh:
@anikethsaha Could you spare a few minutes to take a look. Thanks.

jhildenbiddle:
@anikethsaha --

I know this has already been merged, but setting text width to 80% of the screen resolution is a bad idea. Standard practice for readability is 45-75 characters per line. Here's more detail:

https://www.smashingmagazine.com/2014/09/balancing-line-length-font-size-responsive-web-design/

> THE IDEAL MEASURE: 45 TO 75 CHARACTERS
>
> We have multiple “rules” for facilitating a horizontal reading motion, one of which is to set text at a reasonable measure. As James Craig wrote in his book Designing With Type (originally published in 1971, now it its fifth edition):
>
> Reading a long line of type causes fatigue: the reader must move his head at the end of each line and search for the beginning of the next line.… Too short a line breaks up words or phrases that are generally read as a unit.
If a casual reader gets tired of reading a long horizontal line, then they’re more likely to skim the left edge of the text. If an engaged reader gets tired of reading a long horizontal line, then they’re more likely to accidentally read the same line of text twice (a phenomenon known as “doubling”).
>
> 65 characters (2.5 times the Roman alphabet) is often referred to as the perfect measure. Derived from this number is the ideal range that all designers should strive for: 45 to 75 characters (including spaces and punctuation) per line for print. Many web designers (including me) apply that rule directly to the web. I’ve found, however, that we can reliably broaden the range to 45 to 85 characters (including spaces and punctuation) per line for web pages.

Filling more of the empty space may be visually preferable to some, but there's little argument to be made that this is better for everyone. My recommendation would be to revert this change and let those who prefer content to stretch for wide monitors to add their own custom CSS to a `<style>` block in their `index.html` page.

anikethsaha:
@jhildenbiddle how about https://github.com/docsifyjs/docsify/pull/1273 ?

PR #1067, opened 2020-03-12
[Enhancement]: Added the exit function of gitter live chat.

wan230114:
Added the exit function of gitter live chat, press keyboard esc to exit gitter or click anywhere outside the area to exit.

**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [x] Other, please describe: added the exit function of gitter live chat

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [x] Edge
- [ ] IE

**Other information:**

---


Koooooo-7:
I test it,  work correctly.
but I m not sure whether it is necessary.
I think it is better to have a place collect those tips .

docs/index.html
```diff
@@ -106,6 +106,16 @@
   };
   </script>
   <script src="https://sidecar.gitter.im/dist/sidecar.v1.js" async defer></script>
+  <script>
+    function closechart(params) {
+      const buts = document.getElementsByClassName("gitter-chat-embed-action-bar-item");
```
trusktr, line 111: What does `buts` stand for? We should prefer meaningful code that is easy to understand.

docs/index.html
```diff
@@ -106,6 +106,16 @@
...
+      for (let but of buts) {
+        if (but.className == "gitter-chat-embed-action-bar-item gitter-chat-embed-action-bar-item-collapse-chat") { but.click(); }
+      }
+    }
+    document.onkeydown = function () {if (event.keyCode == 27) {closechart()}};
```
trusktr, line 116: Why is this called `closechart`? Gitter chat is not a chart.

docs/index.html
```diff
@@ -106,6 +106,16 @@
...
+    document.onclick = function (e) {if (e.target.className.indexOf("gitter-open-chat-button") == -1) {closechart()}};
```
trusktr, line 117: We should not use `.onclick` or any `.on` properties on elements. These are not robust and any other code can ruin these handlers.

We should use `document.addEventListener('click', ...)` and `document.addEventListener('keydown', ...)`.

anikethsaha:
> @anikethsaha What's your thought on `type="module"`? Microsoft has moved onto Chromium-based Edge, so all browsers support it now. In my opinion let's roll with the modern browsers. Do we want to support IE 11? (thinks to himself: please say no, please say no).

I dont think we need that. We want to support IE11. I was actually thinking of finding/searching for breakpoints where its not working properly in IE11 and we should fix it. For CSS we should add some postcss based polyfill . 


> place this code in a separate .js file, and include it via <script> tag. This makes it easier for anyone who wants to use the feature to also include it as a script tag without having to copy/paste the code.

yes, maybe we should move the edit button plugin too in a separate file just for our docs.
I think that should be done in separate PR.

docs/index.html
anikethsaha: I think it is for button ?

may be it should be rename to `btn` instead or something else, `buts` doesn't sound good 👍

Koooooo-7 (commented):
Close by remove Gitter to Discord.

Issue #925, opened 2019-09-25
When `relativePath: true`, relative path links in _sidebar do not resolve reasonably


StoneyJackson:
## Bug Report

First, let me say, I'm loving Docsify! Great stuff! 👍

Now on to my issue... 

When using both `loadSidebar:true` and `relativePath:true` relative paths in _sidebar.md files are resolved against the page being rendered rather than the location of the _sidebar.md file. This leads to surprising results. When one writes a relative path in a file, one expects it to be relative to that file, not some other file that implicitly includes it in some subdirectory.

***The work around*** is to only use absolute paths in _sidebar.md files since they may be rendered in arbitrarily nested subdirectories. The problem with this is that it makes it difficult to move an entire subdirectory that contains a _sidebar.md, and it requires authors to know and encode the full path to their subsite.

I am in an environment where we pull documentation from multiple projects and integrate them into one website. I don't want authors to encode the full path to their project's documentation in the main site. This creates an unnecessary dependency, making it challenging to reorganize the structure of the combined site, and it also places additional burden on authors to know the path of their subsite.

So I'm hoping that it's possible to resolve relative URLs in _sidebar.md files against the location of the _sidebar.md file. 🤞 

##### GIVEN

```bash
$ tree
.
├── README.md
├── _sidebar.md
├── index.html
├── subdir1
│   └── README.md
└── subdir2
    └── README.md

2 directories, 5 files

$ cat _sidebar.md
- [Home](README.md)
  - [Subdir1](subdir1/README.md)
  - [Subdir2](subdir2/README.md)

$ cat index.html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>LibreFoodPantry</title>
  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
  <meta name="description" content="Description">
  <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
  <link rel="stylesheet" href="//unpkg.com/docsify/lib/themes/vue.css">
</head>
<body>
  <div id="app"></div>
  <script>
    window.$docsify = {
      loadSidebar: true,
      relativePath: true,
    };
	</script>
	<script src="//unpkg.com/docsify/lib/docsify.min.js"></script>
</body>
</html>
$
```

##### WHEN

1. Run the server
2. Browse to http://localhost:3000/#/
3. Click Subdir1
[...]

Issue #429, opened 2018-03-19
Absolute path for loadNavbar and loadSidebar


chauffer:
Hi,

Is this possible?

Right now putting `/README.md` as one of the fields just puts `/README.md` in the URL, while I want the sidebar to be the one in the root path of the domain.

Thanks!

dmed256:
You could do
```html
<a href="./README.md">Hello</a>
```
or
```md
[Hello](/)
```

By default, Docsify automatically transforms relative directories (end with `/`) into the `README.md` inside that directory

PR #1076, opened 2020-03-15
Support sidebarAbsolutePath to fix relative link and nested sidebar when I just provide one sidebar.md

tangshuang:
**Summary**

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

Referer to issues:
- https://github.com/docsifyjs/docsify/issues/925
- https://github.com/docsifyjs/docsify/issues/429

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -203,6 +203,14 @@ window.$docsify = {
 };
 ```
 
+## sidebarAbsolutePath
+
+In some case, you have only one sidebar.md in your project. But you may put doc files in different directories. You can set `sidebarAbsolutePath` to be `true`, so that it will automaticly make the given `loadSidebar` to be the only one to point at.
+
+It should works with `loadSidebar`. If you do not set `loadSidebar`, don't set this option any more.
+
+It works well with `relativePath` and `basePath`. You do not need to worry about the path conflict.
+
 ## coverpage
 
 - Type: `Boolean|String|String[]|Object`
--- a/packages/docsify-server-renderer/index.js
+++ b/packages/docsify-server-renderer/index.js
@@ -1,13 +1,13 @@
 import { readFileSync } from 'fs';
 import { resolve, basename } from 'path';
 import resolvePathname from 'resolve-pathname';
+import fetch from 'node-fetch';
+import debug from 'debug';
 import { AbstractHistory } from '../../src/core/router/history/abstract';
 import { Compiler } from '../../src/core/render/compiler';
 import { isAbsolutePath } from '../../src/core/router/util';
 import * as tpl from '../../src/core/render/tpl';
 import { prerenderEmbed } from '../../src/core/render/embed';
-import fetch from 'node-fetch';
-import debug from 'debug';
 
 function cwd(...args) {
   return resolve(process.cwd(), ...args);
--- a/src/core/render/compiler/link.js
+++ b/src/core/render/compiler/link.js
@@ -1,8 +1,14 @@
 import { getAndRemoveConfig } from '../utils';
 import { isAbsolutePath } from '../../router/util';
 
-export const linkCompiler = ({ renderer, router, linkTarget, compilerClass }) =>
-  (renderer.link = (href, title = '', text) => {
+export const linkCompiler = ({
+  renderer,
+  router,
+  linkTarget,
+  compiler,
+  isSidebar = false,
+}) => {
+  const link = (href, title = '', text) => {
     let attrs = [];
     const { str, config } = getAndRemoveConfig(title);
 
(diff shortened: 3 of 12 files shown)


tangshuang:
@anikethsaha I’m sorry, I did not find the way to fix. Could you please have a look to this problem?

anikethsaha:
its because you added a header  `##` in `configuration.md` docs. so, the fixtures need to be changed.

add `sidebarAbsolutePath` in array [`here`](https://github.com/docsifyjs/docsify/blob/develop/cypress/integration/sidebar/config.spec.js#L28)

tangshuang:
> its because you added a header `##` in `configuration.md` docs. so, the fixtures need to be changed.
> 
> add `sidebarAbsolutePath` in array [`here`](https://github.com/docsifyjs/docsify/blob/develop/cypress/integration/sidebar/config.spec.js#L28)

Fixed. Thanks!

anikethsaha:
I will review this soon

test/unit/util.test.js
```diff
@@ -2,8 +2,8 @@
 require = require('esm')(
   module /* , options */
 ); /* eslint-disable-line no-global-assign */
-const { resolvePath } = require('../../src/core/router/util');
 const { expect } = require('chai');
+const { resolvePath } = require('../../src/core/router/util');
```
trusktr, line 6: Let's avoid changes that don't have meaning related to the current pull request. Thoughts @anikethsaha ?

src/core/router/history/base.js
```diff
@@ -42,8 +42,20 @@ export class History {
     const { config } = this;
     const base = this.getBasePath();
     const ext = typeof config.ext === 'string' ? config.ext : '.md';
+    const alias = config.alias || {};
+    const rules = Object.assign({}, alias);
+
+    // auto alias sidebar which is set to be absolute path
+    if (config.sidebarAbsolutePath) {
+      const sidebar = config.loadSidebar || '_sidebar.md';
```
trusktr, line 50: Do we need this `sidebarAbsolutePath`? Or can we just provide an absolute path in `loadSidebar`?

docs/configuration.md
```diff
@@ -203,6 +203,14 @@ window.$docsify = {
...
 ```
 
+## sidebarAbsolutePath
+
+In some case, you have only one sidebar.md in your project. But you may put doc files in different directories. You can set `sidebarAbsolutePath` to be `true`, so that it will automaticly make the given `loadSidebar` to be the only one to point at.
+
+It should works with `loadSidebar`. If you do not set `loadSidebar`, don't set this option any more.
+
+It works well with `relativePath` and `basePath`. You do not need to worry about the path conflict.
+
```
trusktr, line 213: These docs don't describe what the option actually does. It doesn't mention that it can be a string. Can you please update?

src/core/router/history/base.js
trusktr: Think about this: what happend if they set `sidebarAbsolutePath` to true, but it isn't an absolute path? Should it be an error? Will they be confused?

Here's an idea on how to detect absolute paths with a regex: https://stackoverflow.com/a/19709846/454780 (but we should allow paths starting with `/` to return true if we want.

test/unit/util.test.js
anikethsaha: yes 👍 
this should be removed

src/core/router/history/base.js
anikethsaha: > Here's an idea on how to detect absolute paths with a regex: https://stackoverflow.com/a/19709846/454780 (but we should allow paths starting with `/` to return true if we want.

Yes, we should consider detecting too.

anikethsaha:
thanks @trusktr for the detailed review. 
just to add a point, this will be a part of `v5` cause I dont think we can do any feature or enhancement release as of now, it all should be released with `v5` if this gets approved.

anikethsaha:
I will create a `v5` branch soon or will need to discuss where to start the development and how.

test/unit/util.test.js
tangshuang: this was changed by cli tool, not by manual

docs/configuration.md
arelstone: I agree with @trusktr. There also seems to be some gramma thats not correct

packages/docsify-server-renderer/index.js
```diff
@@ -1,13 +1,13 @@
 import { readFileSync } from 'fs';
 import { resolve, basename } from 'path';
 import resolvePathname from 'resolve-pathname';
+import fetch from 'node-fetch';
```
arelstone, line 4: I think this should be reverted to the previous state. There is no point in rearranging imports if this is the only changes in the file

src/core/render/compiler/link.js
```diff
@@ -1,23 +1,37 @@
...
 
     if (
       !isAbsolutePath(href) &&
-      !compilerClass._matchNotCompileLink(href) &&
+      !compiler._matchNotCompileLink(href) &&
       !config.ignore
     ) {
-      if (href === compilerClass.config.homepage) {
+      const sidebarAbsolutePath = compiler.config.sidebarAbsolutePath;
+      const basePath =
```
arelstone, line 23: To me the following lines has a weird formatting.

I would suggest doing this instead:
```suggestion
const basePath = typeof sidebarAbsolutePath === 'string'
          ? sidebarAbsolutePath
          : router.getBasePath();
const relativeTo = isSidebar && sidebarAbsolutePath 
          ? basePath 
          : router.getCurrentPath();
```

packages/docsify-server-renderer/index.js
trusktr: This is a result of the automated husky lint process, which modifies the files. The `git commit` step should not leave modifications in the working tree.

PR #1102, opened 2020-04-02
fix #970, updated the insertion of embedded tokens for async returns

### LINKED ISSUE #970: Multiple embed files not loaded correctly
#### Steps to reproduce
Embedding multiple bullet list .md files, for example:

**Sources in _test-embedding.md**
```
## A
- `1`

[filename](./media/a.md ':include')

## B
- `2`

[filename](./media/b.md ':include')

## C
- `3`

[filename](./media/c.md ':include')
```

**Sources in media/a.md**
```
- `A1` AAA
- `A2` AAA
- `A3` AAA
```

**Sources in media/b.md**
```
- `A1` AAA
- `A2` AAA
- `A3` AAA
```

**Sources in media/c.md**
```
- `A1` AAA
- `A2` AAA
- `A3` AAA
```

And also try refreshing page sometime it will show up incorrectly like this:
Screenshot_227

After page loaded, sometimes embedding files are shown in wrong order.
Sometimes there are some files are not embedded. Have to refresh the page.

All files embedding in order correctly.

- Browser version: Chrome latest version
- Docsify version: Latest
- Docsify plugins: None

tschaible:
**Summary**

fix #970 - updated the insertion of embedded tokens so it correctly handles embedded token results returning in any order

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

**Other information:**

The previous implementation incorrectly assumed that contents of embedded content would be returned in the order they were requested in the document.  This created unpredictable race conditions when multiple embeds were included.

I have run into this issue with embedded markdown, and have only attempted test-cases surrounding the markdown content specifically.

---

* [X ] DO NOT include files inside `lib` directory.

### DIFF
--- a/src/core/render/embed.js
+++ b/src/core/render/embed.js
@@ -1,6 +1,6 @@
+import stripIndent from 'strip-indent';
 import { get } from '../fetch/ajax';
 import { merge } from '../util/core';
-import stripIndent from 'strip-indent';
 
 const cached = {};
 
@@ -117,17 +117,27 @@ export function prerenderEmbed({ compiler, raw = '', fetch }, done) {
     }
   });
 
-  let moveIndex = 0;
+  // keep track of which tokens have been embedded so far
+  // so that we know where to insert the embedded tokens as they
+  // are returned
+  const moves = [];
   walkFetchEmbed({ compile, embedTokens, fetch }, ({ embedToken, token }) => {
     if (token) {
-      const index = token.index + moveIndex;
+      // iterate through the array of previously inserted tokens
+      // to determine where the current embedded tokens should be inserted
+      let index = token.index;
+      moves.forEach(pos => {
+        if (index > pos.start) {
+          index += pos.length;
+        }
+      });
 
       merge(links, embedToken.links);
 
       tokens = tokens
         .slice(0, index)
         .concat(embedToken, tokens.slice(index + 1));
-      moveIndex += embedToken.length - 1;
+      moves.push({ start: index, length: embedToken.length - 1 });
     } else {
       cached[raw] = tokens.concat();
       tokens.links = cached[raw].links = links;


tschaible:
When I initially opened this, 2 of the e2e test phases were marked as failed.

I was able to investigate this morning.  The only errors were timeouts in the before phases of the test suite.

I ran the e2e tests locally without issue and have re-pushed the change.  This time everything passed without issue.

I believe the original failure was simply an intermittent failure of the test framework setup.

I have also done some basic testing with Firefox and Safari, in addition to my additional tests in Chrome.  I am not planning further updates, at this time.  This change seems to effectively fix #970 in my use cases.

Please let me know if there is anything else I can do to assist in getting this merged.

anikethsaha:
thanks for the PR and sorry for the delayed.
the tests are not so strong as of now, it bugs most of the time. 
I will try to review it soon .

anikethsaha:
cc @docsifyjs/reviewers

anikethsaha:
@Koooooo-7 can you `approved` the PR if its looks fine by you. I am in the process of adding 2 required approval before merging.

erbridge:
This is affecting us, too. Any chance it could be merged?

src/core/render/embed.js
```diff
@@ -117,17 +117,27 @@ export function prerenderEmbed({ compiler, raw = '', fetch }, done) {
     }
   });
 
-  let moveIndex = 0;
+  // keep track of which tokens have been embedded so far
+  // so that we know where to insert the embedded tokens as they
+  // are returned
+  const moves = [];
```
anikethsaha, line 123:
```suggestion
  let moves = [];
```

tschaible: Can I ask why you're recommending this change?  Is there a style guide or anything that you're using to recommend this?

The recommended change is counter to my understanding of `const`/`let` in Javascript.  `const` ensures that a variable cannot be reassigned, however, it does not prevent modification objects/arrays.

In the code, even though the `moves` array has values inserted into it, it is never re-assigned.  Usage of `const` in this case prevents any accidental re-assignment of `moves` that wasn't intended.

If there's a style guide I should be following, I'm happy to change it, but on the surface, the requested change seems confusing and could potentially lead to an error in the future.

PR #1108, opened 2020-04-09
chore: remove redundant instruction from lint-staged config

jamesgeorge007:
**Summary**
Bump `lint-staged` dep. From `v10` modified files are automatically added to the `git commit index`.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [x] Other, please describe: bump dependency

- [ ] Yes
- [x] No

**Other information:**

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,10 +54,7 @@
     }
   },
   "lint-staged": {
-    "*.js": [
-      "npm run fixlint",
-      "git add"
-    ]
+    "*.js": "eslint --fix"
   },
   "dependencies": {
     "marked": "^0.7.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,10 +54,7 @@
     }
   },
   "lint-staged": {
-    "*.js": [
-      "npm run fixlint",
-      "git add"
-    ]
+    "*.js": "eslint --fix"
```
anikethsaha, line 57: lets keep the `fixlint` can change the script if you want to do any changes.

jamesgeorge007: Since we're specifying target files with a glob pattern it would be kinda redundant to have `fixlint` instead.

anikethsaha: I guess you can change the `fixlint` by removing the `.` .
but I forgot lint-stage does check only the staged file, and not sure `eslint --fix` and `eslint . --fix` will lint the same files or not

jamesgeorge007:
`Some of your tasks use git add command. Please remove it from the config since all modifications made by tasks will be automatically added to the git commit index`

Presently, the above warning shows up while making a commit.

Issue #1126, opened 2020-04-19
vulnerability report


Sharifi-Amin:
## Bug Report

I have found a security vulnerability in docsify.sj. How would you like me to report it?

#### Other relevant information

- [ ] Bug does still occur when all/other plugins are disabled?

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 
- Docsify plugins: 

#### Please create a reproducible sandbox 

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/307qqv236)

anikethsaha:
here only. fill the template and you can submit the report here

Sharifi-Amin:
I would advise against disclosing security vulnerabilities in a public manner before there is fix. please confirm that you want me to disclose details of the vulnerability here. I'll be happy to provide a detailed report and try to help to push a fix as soon as possible .

anikethsaha:
> I would advise against disclosing security vulnerabilities in a public manner before there is fix.

thanks for the alert. no don't report here.
Actually we didn't set up any security policy as of now (I know its bad, I will set up this as soon as possible after discussing with the team ).

- I would suggest reporting it to [Snyk Security Team](https://snyk.io/vulnerability-disclosure/) first. They will help triage the security issue and work with all involved parties to remediate and release a fix.
- if you want to contact me, you can contact through [my email](anik220798@gmail.com)
and if you want to share some sensitive report ( still I would suggest to reach snyk team first ), mention in the mail and we can discuss there how to do so !

trusktr:
How severe is this issue? Is it in an NPM module that affects local development?

It is hard to imagine any issue with static HTML (markdown) sites being insecure. Docsify sites are purely static by default, with no user information (or did I miss something?).

@anikethsaha Maybe we can make a security issue template, and it can specify contact instructions there.

anikethsaha:
we need to set up a policy [here](https://github.com/docsifyjs/docsify/security/policy)

trusktr:
Ah cool, I didn't know about that.

anikethsaha:
> It is hard to imagine any issue with static HTML (markdown) sites being insecure. Docsify sites are purely static by default, with no user information (or did I miss something?).

true, but we do support GA, codefund plugins and markdown may contain embedded files so it may be harmful in those cases. not sure though  😅 . 

[...]

PR #1128, opened 2020-04-21
update: added html sanitizer for remote rendering

anikethsaha:
**Summary**
Added `DOMPurifier` for validating content coming from remote url
https://github.com/docsifyjs/docsify/issues/1126

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


anikethsaha:
cc @Sharifi-Amin

package.json
```diff
@@ -60,6 +60,7 @@
     ]
   },
   "dependencies": {
+    "dompurify": "^2.0.8",
```
QingWei-Li, line 63: The size of this package is too big. May be better with a plugin?

If there is a remote link, it will detect whether there is a XSS filter plugin, if not, then load it asynchronously from unpkg.com or other CDN. And remind users that it is better to manually plugin the XSS filter.

anikethsaha: I agree the package size is big. but in our docs we have mentioned `~21KB gzipped` and I think we are around `~25KB` and this package is `~6KB` and after minimizing through terser, it might be around 2-3 KB more than `~25KB`.
Having plugin is good for us and it will be optional so I don't think existing users would be notified about the plugin. 
yes, we can put a console message but wouldn't it be better to ship it from the core ? 

I am in favor of having it in core or as a plugin but as its a quite security issue I was thinking of shipping initially with core and we can later move out to plugin may be in `v5`? cause already we overshoot the size limit.

trusktr: Regardless how we do it, the change should be in core to prevent the security hole.

packages/docsify-server-renderer/index.js
```diff
@@ -1,18 +1,45 @@
...
 import debug from 'debug';
+import DOMPurify from 'dompurify';
 
 function cwd(...args) {
   return resolve(process.cwd(), ...args);
 }
 
+function isExternal(url) {
+  let match = url.match(
+    /^([^:\/?#]+:)?(?:\/\/([^\/?#]*))?([^?#]+)?(\?[^#]*)?(#.*)?/
```
trusktr, line 19: We can probably remove some of the escape chars here. I like to test regexes over at https://regexr.com. Besides the nice regex features, it has a nice account system: you can log in and save your favorite regexes, and also see others from the community.

packages/docsify-server-renderer/index.js
```diff
@@ -1,18 +1,45 @@
...
+  }
+  if (
+    typeof match[2] === 'string' &&
+    match[2].length > 0 &&
+    match[2].replace(
+      new RegExp(
+        ':(' + { 'http:': 80, 'https:': 443 }[location.protocol] + ')?$'
+      ),
+      ''
+    ) !== location.host
```
trusktr, line 36: Suppose the input `url` is `https://foo.com:1234/foo/bar.md` and the site is at `https://foo.com:1234`.  This would cause it to always return true even if the `url` isn't external, I think, because it would take the `foo.com:1234` and convert it to `foo.com` which would not match `foo.com:1234`, although it should because they are the same host.

packages/docsify-server-renderer/index.js
trusktr: I think using this approach may be easier: https://j11y.io/snippets/getting-a-fully-qualified-url/

I'll make a PR onto this PR...

packages/docsify-server-renderer/index.js
trusktr: I think using this approach may be easier: https://j11y.io/snippets/getting-a-fully-qualified-url/

I'll make a PR onto this PR...

packages/docsify-server-renderer/index.js
```diff
@@ -1,18 +1,34 @@
...
+    url = img.src; // get qualified url
+    img.src = ''; // prevent the server request
+    return url;
+}
+
+function isExternal(url) {
+  url = qualifyURL(url)
+  url = new URL(url)
+  return url.origin !== location.origin
+}
```
trusktr, line 30: Here's the new `isExternal` check, with no room for human error because the URLs are handled the same as the internal browser engine.

GitHub wouldn't let me make a PR to your branch for some reason, so I made a new commit in the same branch.

packages/docsify-server-renderer/index.js
```diff
@@ -95,9 +112,8 @@ export default class Renderer {
       this._renderHtml('cover', await this._render(coverFile), 'cover');
     }
 
-    const html = this.html;
+    const html = this.isRemoteUrl ? DOMPurify.sanitize(this.html) : this.html;
```
trusktr, line 115: This `docsify-server-renderer` package is only for server side though, right? So the bundle size may actually not matter much in this case.

Then this won't fix the client-side issue mentioned in https://github.com/docsifyjs/docsify/issues/1126.

For that, I think we need to modify something in https://github.com/docsifyjs/docsify/tree/develop/src/core/router.

trusktr: Oh wait, I see the fix for client side too. Let me update that.

src/core/render/index.js
```diff
@@ -10,6 +10,7 @@ import { scrollActiveSidebar } from '../event/scroll';
 import { Compiler } from './compiler';
 import * as tpl from './tpl';
 import { prerenderEmbed } from './embed';
+import DOMPurify from 'dompurify';
```
trusktr, line 13: Hmmm... Just an idea, but, what if we use the builtin `DOMParser` to load the HTML, then we can use `querySelector` to find bad parts and remove them, then finally stick the resulting `Node` into the DOM.

This would eliminate the need for the library.

trusktr:
Ooops, let me fix lint errors.

trusktr:
There's also another `isExternal` check here:

https://github.com/docsifyjs/docsify/blob/b2e6123980e98c75a6a6e6cb1f366b19e9856c13/src/core/render/index.js#L260

It should probably re-use the same trick.

trusktr:
Well, I guess that didn't fix lint error. I suppose we should fix everything with `npm run lint` locally. This is all the time I have today, I'll be back.

packages/docsify-server-renderer/index.js
```diff
@@ -1,18 +1,34 @@
...
-import fetch from 'node-fetch';
-import debug from 'debug';
 
 function cwd(...args) {
   return resolve(process.cwd(), ...args);
 }
 
+// Borrowed from https://j11y.io/snippets/getting-a-fully-qualified-url.
+function qualifyURL(url) {
+  const img = document.createElement('img');
```
anikethsaha, line 19: document will be undefined in server !

packages/docsify-server-renderer/index.js
```diff
@@ -1,18 +1,34 @@
...
+  img.src = url; // set string url
+  url = img.src; // get qualified url
+  img.src = ''; // prevent the server request
+  return url;
+}
+
+function isExternal(url) {
+  url = qualifyURL(url);
+  url = new URL(url);
+  return url.origin !== location.origin;
```
anikethsaha, line 29: same for `location`, it will be undefined in server !

anikethsaha:
cc @trusktr Update ?

src/core/render/index.js
trusktr: Taking a look at this now.

src/core/render/index.js
```diff
@@ -259,10 +260,9 @@ export function initRender(vm) {
 
     if (config.logo) {
       const isBase64 = /^data:image/.test(config.logo);
-      const isExternal = /(?:http[s]?:)?\/\//.test(config.logo);
       const isRelative = /^\./.test(config.logo);
 
-      if (!isBase64 && !isExternal && !isRelative) {
+      if (!isBase64 && !isExternal(config.logo) && !isRelative) {
```
trusktr, line 265: Re-using the same `isExternal` logic here.

packages/docsify-server-renderer/index.js
trusktr: Ah ok. Let me see....

packages/docsify-server-renderer/index.js
trusktr: > All checks have passed

I suppose we need testing for server side. Or does it have `document` in the test env perhaps?

packages/docsify-server-renderer/index.js
anikethsaha: cc @trusktr document will be undefined

packages/docsify-server-renderer/index.js
anikethsaha: ditto

anikethsaha: we dont have test in server side. We need to manually test it in our machine

trusktr: We can use this in Node.js: https://nodejs.org/api/url.html#url_url_resolve_from_to

trusktr: But what does `isExternal` mean on the server side, if it does not know what the origin is?

anikethsaha: the server will get the absolute link (URL) of the request path. so we can only allow if the origin is the same as the server. 

In my original commit, it what setup like that to validate the link

https://github.com/docsifyjs/docsify/pull/1128/commits/1e3dd13095c28d5815d908d618b5aae708f34bda#diff-dcea65610b6f84686083cbc6ea82dcb2R17-R42

trusktr: Here's another idea: What if we use `jsdom` on the server side, which not only includes `img`, `document`, `location`, but also has `DOMParser` which we can replace `DOMPurify` with? Then the purification code (our use of DOMParser) will be the same on the client and server, and the client will be without the DOMPurify dependency.

anikethsaha: we only need to purify/sanitize the remote URLs

anikethsaha: JSDOM is quite a big package.

https://bundlephobia.com/result?p=jsdom@16.2.2.

trusktr: Sorry, GitHub updates are slow, I just refreshed and see all your other comments.

Looking at dompurify devDependencies, it already uses `jsdom` for the Node.js side. https://github.com/cure53/DOMPurify/blob/master/package.json#L72

> we dont have test in server side. We need to manually test it in our machine

I haven't used any of the server side stuff. What should I do to test it?

> the server will get the absolute link (URL) of the request path. so we can only allow if the origin is the same as the server.

Ah ok. If we do know the origin, I think `url.resolve` will be useful for it then.

trusktr: Confirmed: if you look at this file, https://unpkg.com/dompurify@2.0.11/dist/purify.cjs.js, that's the one that gets imported in Node.js, and it includes `DOMParser` from `jsdom`.

trusktr: So if we load that on the server, then `window`, `location`, `document`, etc, will all be there.

trusktr: I think as an exercise I'd like to write a test for this, to make sure.

anikethsaha: DOMParser is a purifier ? or its the parser ? I think we need to manually purify it right ?

> Then the purification code (our use of DOMParser) will be the same on the client and server, and the client will be without the DOMPurify dependency.

also, can you explain this, I didn't get it. SSR is optional so if user is not using ssr then we still need to purify the links in client

anikethsaha: a clarification on ⬆️ would be great and yea,  it seems github is not working properly to load the comments , I too have to manually refresh it. 


yes, test for SSR would be awesome. 

there is an issue will router and SSR because of `location` undefined so I think we can use `JSDOM` in the server but still we need `DOMpurifier` for client I think.

trusktr: > DOMParser is a purifier ? or its the parser ? I think we need to manually purify it right ?

After some more research, turns out DOMPurify is quite simple, and is written on top of DOMParser. DOMPurify basically has a list of all the allowed attributes and tags, and removes any that aren't allowed.

For example, [attrs.js](https://github.com/cure53/DOMPurify/blob/master/src/attrs.js)

And they make a whitelist here: https://github.com/cure53/DOMPurify/blob/master/src/purify.js#L167-L183

Seems they could make the lib smaller if they instead made a blacklist. I would assume there are many more allowed tags/attributes than disallowed tags/attributes. However a blacklist is more insecure, for example if new tags or attributes are added then they might slip through, but with a white list all new tags/attrs are ignored.

> SSR is optional so if user is not using ssr then we still need to purify the links in client

I just meant that if we used DOMParser manually on both sides, then that could work too. But turns out DOMPurify is basically doing exactly what I was thinking to do.

anikethsaha: So we are going to keep both client and server same right ? with using DOMpurifier ?

trusktr: Yeah, sounds good, especially if we'll also use jsdom on the serverside for to get plugins working in SSR, and static rendering. :+1: Ok, I'm off for tonight!

packages/docsify-server-renderer/index.js
trusktr: We should be fine, now that I verified we have `JSDOM` running on the server-side. Let me verify if the `img` trick works. Also should write a test while we're at it.

anikethsaha:
I will try to write tests for this. Then I guess we can merge it.

PR #1175, opened 2020-05-13
fix 1098

### LINKED ISSUE #1098: Click the search results on the left, the right page cannot be located
#### Steps to reproduce
show this in stackoverflow
https://stackoverflow.com/questions/60945607/docsify-click-the-search-results-on-the-left-the-right-page-cannot-be-located

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 
- Docsify plugins: 

sy-records:
**Summary**

fix #1098

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

**Other information:**

---

### DIFF
--- a/src/plugins/search/component.js
+++ b/src/plugins/search/component.js
@@ -174,7 +174,9 @@ function bindEvents() {
   Docsify.dom.on(
     $search,
     'click',
-    e => e.target.tagName !== 'A' && e.stopPropagation()
+    e =>
+      ['A', 'H2', 'P', 'EM'].indexOf(e.target.tagName) === -1 &&
+      e.stopPropagation()
   );
   Docsify.dom.on($input, 'input', e => {
     clearTimeout(timeId);
--- a/src/plugins/search/component.js
+++ b/src/plugins/search/component.js
@@ -170,7 +170,14 @@ function bindEvents() {
   const $inputWrap = Docsify.dom.find($search, '.input-wrap');
 
   let timeId;
-  // Prevent to Fold sidebar
+  
+  /**
+   * Prevent to Fold sidebar.
+   * 
+   * When searching on the mobile end,
+   * the sidebar is collapsed when you click the INPUT box,
+   * making it impossible to search.
+   */
   Docsify.dom.on(
     $search,
     'click',
--- a/src/plugins/search/component.js
+++ b/src/plugins/search/component.js
@@ -170,13 +170,13 @@ function bindEvents() {
   const $inputWrap = Docsify.dom.find($search, '.input-wrap');
 
   let timeId;
-  
+
   /**
-   * Prevent to Fold sidebar.
-   * 
-   * When searching on the mobile end,
-   * the sidebar is collapsed when you click the INPUT box,
-   * making it impossible to search.
+    Prevent to Fold sidebar.
+
+    When searching on the mobile end,
+    the sidebar is collapsed when you click the INPUT box,
+    making it impossible to search.
    */
   Docsify.dom.on(
     $search,


src/plugins/search/component.js
```diff
@@ -174,7 +174,9 @@ function bindEvents() {
   Docsify.dom.on(
     $search,
     'click',
-    e => e.target.tagName !== 'A' && e.stopPropagation()
+    e =>
+      ['A', 'H2', 'P', 'EM'].indexOf(e.target.tagName) === -1 &&
```
anikethsaha, line 178: what about other tags ?

sy-records: Are there any others? All i see is this.
image

anikethsaha: ahh, yeah I checked the code, it seems these are enough. we can add them later if we missed something 👍

trusktr: What about using `preventDefault` instead of `stopPropagation`? It would make the event handleable by high up code.

Also, maybe we can make this more generic and not maintain a list of specific DOM nodes, so that code won't break if we update markup structure.

How about something like 

```js
// select any anchor and anything inside the anchors
const linkContent = $search.find('a *')

// ...

    e =>
      !$(e.target).is(linkContent) && e.preventDefault()?
```

?

trusktr: In general I think we should improve output so Docsify content is clear, and we can rely on something like

```js
const linkContent = $search.find('.docsify-link *')
```

I had started doing that in my one messy PR. I will split it up soon. But feel free to pluck changes if I'm too slow!

trusktr: Oh, I think `stopPropagation` is to prevent the menu from closing in another event handler, isn't it? Nevermind preventDefault then (for now).

In that case, it may be cleaner to consolidate that other code with this code, so that the intent is clear, but that's a bigger task.

stepozer: Today I also found this code :) I have question, what if we totally delete this section? 
```

// Prevent to Fold sidebar   -   what does it mean???
    Docsify.dom.on(
      $search,
      'click',
      function (e) { return e.target.tagName !== 'A' && e.stopPropagation(); }
    );
```

I deleted it and nothing happens. What bug this code fix? Do you know? I think better will delete this code instead try determine which tag type user click.

sy-records: > 今天，我也找到了这段代码：)我有一个问题，如果我们完全删除此部分该怎么办？
> 
> ```
> 
> // Prevent to Fold sidebar   -   what does it mean???
>     Docsify.dom.on(
>       $search,
>       'click',
>       function (e) { return e.target.tagName !== 'A' && e.stopPropagation(); }
>     );
> ```
> 
> 我删除了它，没有任何反应。此代码修复了哪些错误？你知道吗？我认为最好删除此代码，而不是尝试确定用户单击哪种标记类型。

When searching on the mobile end, the sidebar is collapsed when you click the INPUT box, making it impossible to search.

sy-records: > Today I also found this code :) I have question, what if we totally delete this section?
> 
> ```
> 
> // Prevent to Fold sidebar   -   what does it mean???
>     Docsify.dom.on(
>       $search,
>       'click',
>       function (e) { return e.target.tagName !== 'A' && e.stopPropagation(); }
>     );
> ```
> 
> I deleted it and nothing happens. What bug this code fix? Do you know? I think better will delete this code instead try determine which tag type user click.

When searching on the mobile end, the sidebar is collapsed when you click the INPUT box, making it impossible to search.

sy-records: @trusktr Hi, I think it's okay to modify it this way for now, and there should be no other DOM.
If it's going to be modified, it should be in v5, and the next release shouldn't be v5.

stepozer: In this case I suggest:

1. Write commend that describe why this code needed. For example:
```
When searching on the mobile end, the sidebar is collapsed when you click the INPUT box, making it impossible to search.
```

2. I suggest try to search closest parent link `A` instead all of this tags. Maybe it will be more correct? 

What do you think about it?

sy-records: @stefek99 As I said in https://github.com/docsifyjs/docsify/pull/1175#discussion_r424240260, there are only so many DOMs... 
if there are other more excellent way you can modify this PR, thanks!

sy-records: @stepozer Thanks to the first point of suggestion, which has been revised. 

On the second point, as I said in https://github.com/docsifyjs/docsify/pull/1175#discussion_r424240260, there are only so many DOMs... 

if there are other more excellent way you can modify this PR, thanks!

stepozer: @sy-records as I suggested you can:

1. stopPropagation if current element == `A` tag
2. stopPropagation if current element has parent == `A` tag

Because how I understand all this tags always wrapped in `A`. What do you think?

sy-records: `stopPropagation if current element == A tag` This doesn't seem right. 🤔

stepozer: For pure JS exists equivalent https://stackoverflow.com/questions/38469830/what-is-the-pure-javascript-equivalent-of-jquerys-parents-method

stepozer: > `stopPropagation if current element == A tag` This doesn't seem right.

Sorry, I mean `!==`. Right now you has logic:

if  e.target.tagName !== 'A' then stopPropagation. But in your tag `A` exists other tags, so you need check parent for this tag:

stopPropagation if current element !== A AND current element has NO parents with A tag.

stepozer: I tried write code, I think it will be something like this:

```
function getParentByTag(e, parentTag) {
    for (var p = e && e.parentElement; p; p = p.parentElement) {
      if (p && p.tagName == parentTag) {
        return p;
      }
    }
    return null;
  }

  function elementIsLinkOrHasLinkOnParents(element) {
    var parenLink = getParentByTag(element, 'A')
    return element.tagName === 'A' || parenLink != null;
  }

// Prevent to Fold sidebar
Docsify.dom.on(
  $search,
  'click',
  function (e) { return !elementIsLinkOrHasLinkOnParents(e.target) && e.stopPropagation(); }
);
```

Sorry, I am back-end developer :)

stepozer: I tested on mobile, seems to be working...

sy-records: It's a little complicated... Let's keep it simple. 😥

stepozer: I will create separated PR for this :) I think it will be more right way

sy-records: Sure, feel free to submit a PR. 👍🏻

stepozer: Will wait until your PR will be merged

anikethsaha:
cc @trusktr @Koooooo-7

Issue #1178, opened 2020-05-16
On hovering side menu items, It should tooltip the menu item. because long menu item text not fully visible in the screen


kumaravel95:
## Feature request
On hovering side menu items, It should popover the text.

#### What problem does this feature solve?
Long menu item text not fully visible in the screen. because we added ellipsis. so, tooltip will be helpful at that time.

![image](https://user-images.githubusercontent.com/9389944/82124019-05fe7800-978c-11ea-83d2-a6bd5c7dd2bc.png)

#### What does the proposed API look like?
Tooltip

#### Are you willing to work on this yourself?
Yes

PR #1179, opened 2020-05-16
Tooltip added for sidemenu subitems

kumaravel95:
**Summary**
On hovering side menu items, It should tooltip the menu item. because long menu item text not fully visible in the screen #1178

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

**Before fix:**
image

**After Fix:**
image

- [ ] Yes
- [x] No

- [x] On hovering side menu items, It should tooltip the menu item. because long menu item text not fully visible in the screen #1178

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Edge
- [x] IE


kumaravel95:
gitpod url to verify : https://gitpod.io/#https://github.com/kumaravel95/docsify

trusktr:
> work to finish before v5 can ship.

Just curious, does it specifically matter if we ship a version that has the number `5` in it? I think it may be more important that we're all on board with any feature that comes out, regardless if it is v4 or v5.

I think having the words wrap is a reasonable change. Would it be considered a breaking change and needs to be v5? Maybe whatever the solution is, it should be off by default, so that an option can enable it? If we do that then we can stay in v4 until we're ready to turn the feature on by default.

There are still a lot of issues that we need to triage and figure out what to do with. PRs too.

jhildenbiddle:
Why don't we just leverage the link "tooltip" functionality that ships with every browser via the `title` attribute?

Change this:

https://github.com/docsifyjs/docsify/blob/b9c38924d8c0429a04032db68ee1d966a664d161/src/core/render/tpl.js#L96

To this (notice the `title` attribute):

```js
innerHTML += `<li><a class="section-link" href="${node.slug}" title="${node.title}">${node.title}</a></li>`;
```

Here's what it looks like:

Screen Shot 2020-06-13 at 1 18 36 AM

No new markup, no CSS changes, familiar behavior and styling, easy to implement in v4, and no concerns about impacting v5. For v5 I would argue that allowing sidebar links to wrap is a better solution because tooltips only display on hover which requires a pointing device (touch-based devices are out of luck).

anikethsaha:
I think the `title` one as mentioned above seems much better. It does work in `IE11` as well .

kumaravel95:
I updated my PR with **title attribute** and removed other changes, It won't create bugs and we can merge this PR without any fear.

jhildenbiddle:
@anikethsaha --

Looks good for nested sidebar links. 

Any reason why we wouldn't do the same for top-level sidebar links? For example, shouldn't "Quick start" get a `title` attribute just like it's sub links?

anikethsaha:
Yea we can totally have that. 
Either in this PR or in a separate, both will work 👍 

cc @kumaravel95 do you want to add them in This  ?

kumaravel95:
Title attribute can be added to sub-level menu because those DOM structure are constructed in `tpl.js`. But, top-level sidebar is generated from **markdown file and compiler**.

top-level iems are inside anchor tags. so still we have an option to add title attribute from compiler re-render origin function(**link**) available in `compiler.js`. but if any changes will reflect in all the archer tags the site.

**link.js**
from:
`return '<a href="${href}" ${attrs.join(' ')}>${text}</a>';`
to:
`return '<a href="${href}" ${attrs.join(' ')} title="${text}">${text}</a>';`

But its a **bad idea** to make this change. because, It will break some anchor tags that contains another DOM, (Eg: Translation drop-down items contains image tag inside ancher tag).

I checked any markdown format for anchor title. but could't find solution for that.

Any good way available to add title attribute for the top-level items?

trusktr:
> Why don't we just leverage the link "tooltip" functionality that ships with every browser via the title attribute

Doesn't work on mobile (there's no hover except for on some Samsung phones where you can actually hover your finger without touching the glass). Not a blocker, but it'd be nice to have a solution for that. I think wrapping lines is the easiest for that.

With the current `title` change, it's an incremental change worth merging now I think. But yeah, it will be desirable to make it work everywhere (maybe even on body links).

> Title attribute can be added to sub-level menu because those DOM structure are constructed in tpl.js. But, top-level sidebar is generated from markdown file and compiler.

Yeah, some consolidation is needed. I was noticing that too. Let's not worry about it yet, but if any ideas, please do open a new issue for any refactor, and otherwise I'll approve this as it is a good non-breaking forward-looking change.

trusktr:
Feel free to merge if ready!

jhildenbiddle:
@kumaravel95 --

Agreed on not adding a `title` tag to top-level link for the reason you describe. I'm not familiar with that part of the codebase so I was unaware of the risks (I incorrectly assumed it would be an easy change). Thanks for looking into that though!

FYI, I've added the following to #1212:

```md
- [ ] Allow sidebar links to wrap instead of being truncated (see #1179)
```

Didn't want this to get lost in the shuffle.

PR #1182, opened 2020-05-17
Updated docs site dark and light mode with switch and redesigned search bar using docsify-darklight-theme

boopathikumar018:
**Summary**

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

Updated docs site dark and light mode with switch and redesigned search bar using [docsify-darklight-theme](https://www.npmjs.com/package/docsify-darklight-theme) you can refer the docs and see full preview [here](https://boopathikumar018.github.io/docsify-darklight-theme)

# Light Mode

## before

image

## after

image

# Dark Mode

## before

image

## after 

image

- [ ] Yes
- [x ] No

- [x ] Chrome
- [x ] Firefox
- [ ] Safari
- [ x] Edge
- [ ] IE

- [ x] A convincing reason for adding this feature
  -  Default theme mode detection in supported browser versions.
  - Theme Switcher.
  - Themes are customizable based on your color preferences.
  - Option for other plugins to support for (Dark/Light) mode. View setup guide here.
  - Preferences can be modified directly in window.$docsify configuration object.
  - Toogle icons can be configured based on your preference using configuration object.
  - Default theme(Dark/Light) can be configured based on your needs.
  - Themes are remembered and retrieved from local storage.
  - Redesigned search box.
- [x ] Related documents have been updated
  [Documentation & Preview](https://docsify-darklight-theme.boopathikumar.me)

**Other information:**

---

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -12,7 +12,7 @@
   <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
   <link 
     rel="stylesheet"
-    href="//cdn.jsdelivr.net/npm/docsify-darklight-theme@latest/dist/style.min.css"
+    href="//cdn.jsdelivr.net/npm/docsify-darklight-theme@3/dist/docsify-style.min.css"
     title="docsify-darklight-theme"
     type="text/css"
   />
@@ -69,6 +69,9 @@
           '/': 'Search'
         }
       },
+      darklightTheme: {
+        bodyFontSize : '15px',
+      },
       formatUpdated: '{MM}/{DD} {HH}:{mm}',
       plugins: [
         function (hook, vm) {
@@ -100,7 +103,7 @@
   <script src="//cdn.jsdelivr.net/npm/prismjs/components/prism-markdown.min.js"></script>
   <script src="//cdn.jsdelivr.net/npm/prismjs/components/prism-nginx.min.js"></script>
   <script 
-    src="//cdn.jsdelivr.net/npm/docsify-darklight-theme@latest/dist/index.min.js"
+    src="//cdn.jsdelivr.net/npm/docsify-darklight-theme@3/dist/index.min.js"
     type="text/javascript">
   </script> 
   <script>
--- a/docs/index.html
+++ b/docs/index.html
@@ -10,13 +10,18 @@
   <meta name="keywords" content="doc,docs,documentation,gitbook,creator,generator,github,jekyll,github-pages">
   <meta name="description" content="A magical documentation generator.">
   <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify/lib/themes/vue.css" title="vue">
+  <link 
+    rel="stylesheet"
+    href="//cdn.jsdelivr.net/npm/docsify-darklight-theme@latest/dist/style.min.css"
+    title="docsify-darklight-theme"
+    type="text/css"
+  />
+  <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify/lib/themes/vue.css" title="vue" disabled>
   <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify/lib/themes/dark.css" title="dark" disabled>
   <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify/lib/themes/buble.css" title="buble" disabled>
   <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify/lib/themes/pure.css" title="pure" disabled>
  
   <script src="//cdn.jsdelivr.net/npm/docsify-plugin-codefund/index.js"></script>
-  <link rel="stylesheet"  href="//cdn.jsdelivr.net/npm/docsify-dark-mode@0.6.1/dist/style.css"/>
   <style>
     nav.app-nav li ul {
       min-width: 100px;
@@ -64,11 +69,6 @@
           '/': 'Search'
         }
       },
-      darkMode: {
-        light: {
-          toggleBtnBg: '#42b983'
-        }
-      },
       formatUpdated: '{MM}/{DD} {HH}:{mm}',
       plugins: [
         function (hook, vm) {
@@ -99,7 +99,10 @@
   <script src="//cdn.jsdelivr.net/npm/prismjs/components/prism-bash.min.js"></script>
   <script src="//cdn.jsdelivr.net/npm/prismjs/components/prism-markdown.min.js"></script>
   <script src="//cdn.jsdelivr.net/npm/prismjs/components/prism-nginx.min.js"></script>
-  <script src="https://cdn.jsdelivr.net/npm/docsify-dark-mode@0.6.1/dist/index.js"></script>
+  <script 
+    src="//cdn.jsdelivr.net/npm/docsify-darklight-theme@latest/dist/index.min.js"
+    type="text/javascript">
+  </script> 
   <script>
   ((window.gitter = {}).chat = {}).options = {
     room: 'docsifyjs/Lobby'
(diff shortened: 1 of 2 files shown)


anikethsaha:
also, can we have the old font ?

boopathikumar018:
> Can we have the search bar not changed?

are you sure you want to have old search bar ?

> also,please fix the CI issue.

Can you please explain this in detail

> also, can we have the old font ?

sure!

anikethsaha:
> are you sure you want to have old search bar ?

yes, that's the one shipped by default. So we need to show that.

boopathikumar018:
> yes, that's the one shipped by default. So we need to show that.

Yes I will change it to default style.

boopathikumar018:
## Requested changes completed

 - Changed to default search bar
 - Updated font-family

## Dark Mode

image

## Light Mode

image

docs/index.html
```diff
@@ -64,10 +70,8 @@
           '/': 'Search'
         }
       },
-      darkMode: {
-        light: {
-          toggleBtnBg: '#42b983'
-        }
+      darklightTheme: {
+        bodyFontSize : '15px',
```
anikethsaha, line 74: whats the default ?

boopathikumar018: the default font size of docsify site is 15 px. In my plugin I gave 17px as default. So I am overriding here.

trusktr:
that new search bar was neat, but better to keep the default (non-breaking change) for now. We can however provide style addons. Would be happy to take a PR that can add (for example) a CSS file that someone can optionally import to have an alternative search bar, etc.

Also take a look at docsify-themeable by @jhildenbiddle. That plugin makes great use of CSS vars for customization. I think it would be great to have a system like that in core (but with the default theme), and then @jhildenbiddle's theme can be an additional (optional) theme that builds on top of that system.

anikethsaha:
did you meant this search bar ? https://github.com/docsifyjs/docsify/pull/1182#issuecomment-631022152

trusktr:
I meant the one with rounded corners in the "after" shots. But the new dark one looks great too.

anikethsaha:
rounded one was the proposed one,  I asked for the old one. do you prefer the rounded one ?

jhildenbiddle:
Please hold off on this merge. I will provide more detail before the end of the day. Thanks!

jhildenbiddle:
@anikethsaha, @trusktr --

1. Generally speaking, we should be cautious about which (if any) third-party docsify-related addons we use on the public docsify site. Most visitors will view the site as an example of what they will get with an "out of the box" docsify installation. If after following our installation guide their site looks or behaves differently (which will be the case if docsify.js.org uses this theme), confusion can lead to issues being filed (which we have to manage) and an overall frustrating experience. We could try to address this by listing all of the plugins we use on the site at the end of our installation guide, but then we're essentially endorsing some addons over others which isn't great for the community. Unless there is legitimate need, I believe it's best to let docsify.js.org be a showcase primarily for code from the docsify team.
1. Aside from the general guidance above, the docsify site should not use a third-party theme due to the maintainability issues it creates. Consider what will happen when a docsify UI change is necessary that requires HTML and/or CSS updates. The docsify team can modify the generated HTML, but they will now be reliant on the third-party theme author to update their code before changes can be made public. This is not a position we want to put ourselves in, especially when our own [documentation](https://docsify.js.org/#/themes) lists four officially-supported themes that the docsify team is presumably responsible for maintaining (but they don't actually use?).
1. If the goal is to stabilize v4 and turn our attention to v5, changes like this (new features or non-critical updates) should not be merged. Doing so only increases the amount of v4 testing that needs occur and/or the likelihood that new v4 issues will appear, further delaying the release of v5. For example:
   - This theme does not render properly in IE11. <br>IE11 CSS
   - This theme's plugin JS throws an error in IE11 (officially supported by docsify v4) which breaks all docsify-powered sites. <br>IE11 JS Error
   - This theme/plugin is not compatible with the [theme preview links](https://docsify.js.org/#/themes) on docsify.js.org
   - On small screens (e.g. phones), the toggle switch is visible but positioned under the transparent portion of the Github corner image. This makes it impossible to toggle themes and unexpectedly navigates to Github when attempting to click the toggle.
   - ... and maybe more (but also maybe not).

I don't mean any of this to sound like Mr. Buzz Kill. 

I think @boopathikumar018 has done a nice job with [docsify-darklight-theme](https://docsify-darklight-theme.boopathikumar.me/), so I don't want the above to sound discouraging. My feedback is less about this particular theme/plugin combo and more about what's best for docsify (IMHO, anyway). If a better dark/light theme switcher is what we're after (and that is how this PR was pitched), I suggest the following:

1. I'd love to see @boopathikumar018 offer a more focused plugin (not a theme) that provides dark/light theme functionality with two core features: the ability to specify separate dark and light URLs (so dark/light themes are configurable) and the ability to enable a dark/light URL by default based on the operating systems current dark/light mode settings ([see here for details](https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme)). The only CSS that should accompany this plugin are the styles necessary for the toggle itself.
1. An interesting stretch goal for the plugin would be to allow more than two theme URLs to be specified (not just dark/light). When two URLs are specified, the icon element acts as a toggle. When three-or-more URLs are specified, a drop-down list of themes appears under the icon on hover/tap. I mention this because *both* the docsify and docsify-themeable sites could use this functionality to provide theme previews.
1. Ensure the JS errors have been addressed in IE11. I didn't spend too much time debugging, but it looks like an arrow function is used which IE (obviously) doesn't like.
1. Provide the current docsify-darklight-theme "dark" theme as a separate theme (no JS, just the theme). This will allow users to select it as their theme without the plugin, or use it with the plugin along with docsify's own "vue.css" theme to recreate docsify-darklight-theme's current look and functionality.

The advantage of the steps above comes down to flexibility: both docsify and docsify-themeable could use the plugin to swap between their own "official" dark/light themes (vue.css/dark.css for docsify, "simple" and "simple-dark" for docsify-themeable). More importantly, users would be able to select their own dark/light themes, perhaps mixing themes from different authors.

anikethsaha:
I didnt test that in IE11.  (Sorry, It was my bad )
I think we should revert this if that's not working (which is not as mentioned above. )

Also i agreed with the docs site as a point of reference by docsify users. 

IMO, let's remove any dark mode plugin in the official docs. We have planned to keep it officially from the core in the next version (v5), so I don't think we should have it now. 

@boopathikumar018 You did an amazing work with the plugin, but as mentioned by @jhildenbiddle , your plugin is changing the default theme which we should not have as this is one of the active examples of docsify users. 

I am reverting this and will remove the old dark mode plugin as well.

jhildenbiddle:
Since [docisfy-themeable offers a light and dark theme](https://jhildenbiddle.github.io/docsify-themeable/#/themes) already, these will be available as part of v5.

**"Simple" Theme**

theme-simple-content

**"Simple Dark" Theme**

theme-simple-dark-content

Users can select one of these themes, or they can include both stylesheets and leverage the  `prefers-color-scheme` media query to have supporting operating systems select the light/dark stylesheet that matches their OS mode (I covered this in detail at https://github.com/jhildenbiddle/docsify-themeable/issues/19):

```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsify-themeable@0/dist/css/theme-simple.css">
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsify-themeable@0/dist/css/theme-simple-dark.css" media="(prefers-color-scheme: dark)">
```

We also have the option of simplifying this setup for users by offering light/dark theme selection as part of their docsify configuration. Docsify can then inject the necessary `<link>` elements with the appropriate attributes.

```html
<script>
  window.$docsify = {
    lightTheme: 'https://path/to/light-theme.css',
    darkTheme: 'https://path/to/dark-theme.css'
    // ... etc.
  }
</script>
```

No toggle necessary (for operating systems that support it, which means the latest versions of Windows and macOS), and I'd argue this is probably what users who have opted for "dark mode" on their OS would prefer. A toggle might be a nice option (especially for those on older operating systems that don't offer a "dark mode"), but I'm not convinced it needs to be core docsify functionality. Seems much better a plugin, IMHO.

PR #1189, opened 2020-05-22
Allow configs to be functions

trusktr:
**Summary**

Allows user configs to be functions. A function config is passed an instance of the `vm`, which can be useful for referencing in other parts of the config, like in `markdown.renderer` hooks.

Also adds some unit tests in `docsify.test.js` to

- test that the existing Docsify API is available globally when loaded via script tag.
- test that Docsify accepts a config function (`window.$docsify` being a function that returns a config object)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

- [x] Related unit tests have been updated
- [ ] e2e tests have been updated


trusktr:
Gotta check what's going on with the failed e2e test.

trusktr:
Running `npm run test:e2e` passes for me locally.

trusktr:
Ooooh, is it because of older Node for one run of the test perhaps? I'll get back to this tomorrow.

trusktr:
Looks like I should re-do my commits to use the `feat:`, `fix:`, etc prefixes.

trusktr:
I tried switching to Node v10 locally, ran `npm run test:e2e` and all tests passed. Is there a different way to test it?

anikethsaha:
> Looks like I should re-do my commits to use the `feat:`, `fix:`, etc prefixes.

yes, refer [angular commit convention](https://github.com/angular/angular/blob/master/CONTRIBUTING.md). I will add commitlint soon to guide with the commit message.

src/core/index.js
```diff
@@ -11,7 +11,7 @@ import initGlobalAPI from './global-api';
  * @returns {Number|void} If the page is already laoded returns the result of the setTimeout callback,
  *  otherwise it only attaches the callback to the DOMContentLoaded event
  */
-function ready(callback) {
+export function documentReady(callback) {
```
anikethsaha, line 14: Is it exported for tests ?

trusktr: Yeah, otherwise not imported by anything else currently.

trusktr: But maybe `documentReady` is better in a separate util file.

trusktr:
> I will add commitlint soon to guide with the commit message.

Worth checking out [semantic-release](https://github.com/semantic-release/semantic-release) too.

src/core/index.js
anikethsaha: I think the current file seems ok as it is not really a utility right

anikethsaha:
> > 
> 
> Worth checking out [semantic-release](https://github.com/semantic-release/semantic-release) too. But that would be some work to set up. I feel like we need to clean things up, add tests, etc, first.

yea, this will need changes in the publish script.

trusktr:
I just noticed something else! All tests pass (at least for me locally, not sure why it fails on CI yet), however, the `npm run build` command currently has a Rollup compile error, but the script exits `0` as if nothing happened.

I will fix this...

trusktr:
Ok, now that I could debug it (PR https://github.com/docsifyjs/docsify/pull/1192), I found out that adding the `export` words in `src/core/index.js` causes the Rollup to have that error `You must supply "output.name" for IIFE bundles`.

Let me see here...

trusktr:
Nice. All checks have passed. What I could do though, is remove the config function stuff, and just make this a pull requests that adds the test updates, then make a new minimal PR just for the config function. Let me know if that's what I should do.

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
trusktr, line 11: This is super useful. There's an editorconfig plugin for just about every editor. This configures your editor automatically so you don't have to do it manually per project.

Although `prettier` formats code, it doesn't configure the editor, so if the editor is set to 4-space indent, and prettier is set to 2-space indent, then indenting will automatically happen in the editor using 4 spaces, and when we run prettier it will re-format back to 2 spaces, which can be jarring.

And for example when prettier is set to 4-space indent but the editor is configured to 2-space indent, then we will see extra indent guide lines in the editor (f.e. in VS Code).

So having editorconfig is convenient to ensure people's editors have the correct settings, even if otherwise tests would still pass due to ESLint and Prettier formatting the code.

build/build.js
```diff
@@ -9,8 +9,16 @@ const version = process.env.VERSION || require('../package.json').version
...
-const build = function (opts) {
-  rollup
+/**
+ * @param {{
+ *   input: string,
+ *   output?: string,
+ *   globalName?: string,
+ *   plugins?: Array<import('rollup').Plugin>
+ * }} opts
+ */
```
trusktr, line 19: I'm in VS Code, so the JSDoc comments help the TypeScript language server give me better intellisense. Plus this is self-documenting even if you don't have VSCode/TypeScript.

docs/configuration.md
```diff
@@ -12,6 +12,24 @@ You can configure the `window.$docsify`.
...
+      markdown: {
+        renderer: {
+          code(code, lang) {
+            // ... use `vm` ...
+          },
+        },
+      },
+    };
+  };
+</script>
```
trusktr, line 30: Another approach we could take is to instead make sure we also pass in `vm` to all the markdown renderer hooks. But the current approach is easier and more resilient to changes in the code where marked is being called.

trusktr:
Questions:

- Wdyt about the build mods and tests?
- Thoughts on the config function?
- Thoughts on moving to a new `DOCSIFY` global and not adding a bunch of globals?

anikethsaha:
I havent check the build locally but looks cool. 

Always great to have tests. 

@trusktr I was thinking of having this for `v5`.  in that way we don't have to be compatible with the old one cause we can safely have breaking changes for the next major version.

WDYT ?

build/build.js
anikethsaha: Always great to have jsdoc comments. It helps contributing very smooth. 

Thanks 🎉

docs/configuration.md
anikethsaha: wrapper function for hooks [already has the `vm`](https://docsify.js.org/#/write-a-plugin?id=full-configuration) as a parameter.

trusktr:
> I was thinking of having this for `v5`. in that way we don't have to be compatible with the old one cause we can safely have breaking changes for the next major version.
> 
> WDYT ?

It thought about that too, but there are some issues:

- in order to use `export` in the entry points, so that we can easily import stuff into the test code without relying on globals, this requires setting Rollup's `output.name` option, which will create a global variable of that name. So regardless, it creates globals. Maybe we just don't document this new `DOCSIFY` global yet?
- If we want to avoid adding a new global in the global build, then we need to avoid `export` in the entry point.
  - Maybe I can refactor the code so that then entry point contains only a single line, `documentReady(() => new Docsify)` and does not export anything.
  - Then we move all other stuff into other modules that this new small entry point imports. The test code could then import from the other modules instead of the entry point. This may work: we can avoid making a new global by simply not having `export` words in the entry point, and the test code can import from the other modules and avoid relying globals. 
- A downside of waiting for v5 would be that then we can not try new features now. I think personally I'd rather make improvement now, and let people have them (as long as it doesn't break any existing features), and then when we're ready we can remove deprecated features as a cleanup in a major version (along with any new features at that point too).

build/build.js
trusktr: Plus where you use the function, the editor can show you documentation, like in VS Code:

Screenshot from 2020-06-07 10-58-03

anikethsaha:
fixing https://github.com/docsifyjs/docsify/pull/1193 so that we can test this .

packages/docsify-server-renderer/index.js
```diff
@@ -1,13 +1,13 @@
...
 import resolvePathname from 'resolve-pathname';
+import fetch from 'node-fetch';
+import debug from 'debug';
 import { AbstractHistory } from '../../src/core/router/history/abstract';
 import { Compiler } from '../../src/core/render/compiler';
 import { isAbsolutePath } from '../../src/core/router/util';
 import * as tpl from '../../src/core/render/tpl';
 import { prerenderEmbed } from '../../src/core/render/embed';
-import fetch from 'node-fetch';
-import debug from 'debug';
```
trusktr, line 10: Some of these changes are automatic ESLint changes. I figured to commit the changes so they don't keep showing up.

anikethsaha: cool, 
thanks

docs/configuration.md
trusktr: Those are for Docsify hooks, but not for `marked` renderer plugins.

anikethsaha: make sense

trusktr:
> Can we have some more tests? like function returning wrong config object.

We could, but if that happens it will fail the same way as any wrong config currently does. Those tests wouldn't be specifically testing this new config function feature. We can make more PRs to add more tests on the actual config object.

trusktr:
Plus the PR is already kinda big. But now we're in better shape to add more tests.

anikethsaha:
okay, make sense to have separate PR for test. 
I will have a final look at this one more time soon..

trusktr:
In the last commit, see how I modified the `test` script in `package.json`. It just runs `npm run build` now too. `npm run build` creates the `lib/docsify.js` file, and the tests in `docsify.test.js` are relying on `Docsify` from that file to ensure the global API exists.

trusktr:
Oops, I forgot to push my commit. :laughing: See the commit with message `ensure that the test script runs a prod build`.

package.json
```diff
@@ -28,7 +28,7 @@
     "dev:ssr": "run-p serve:ssr watch:*",
     "lint": "eslint .",
     "fixlint": "eslint . --fix",
-    "test": "mocha ./test/**/*.test.js",
+    "test": "npm run build && mocha ./test/**/*.test.js",
```
anikethsaha, line 31: I think it will still test against the source code instead of the build code ?

cause test files are taking modules from `src/*` directories, right ?

let me know if I missed anything

anikethsaha: I guess if this is so, we can have a separate tests, script for post build testing in a separate PR ?

trusktr: actually we don't need this change because the GitHub Action (`unit.yml`) already runs the production build.

Inside of `docsify.test.js`, it loads the production build file (`lib/docsify.js`).

test/unit/docsify.test.js
```diff
@@ -0,0 +1,133 @@
...
+const { initJSDOM } = require('../_helper');
+
+const docsifySite = 'http://127.0.0.1:3000';
+
+const markup = /* html */ `<!DOCTYPE html>
+  <html>
+    <head></head>
+    <body>
+      <div id="app"></div>
+      <script src="/lib/docsify.js"></script>
```
trusktr, line 17: @anikethsaha This is where it loads the built output file.

test/unit/docsify.test.js
```diff
@@ -0,0 +1,133 @@
...
+/** @type {ReturnType<typeof http.createServer>} */
+let server;
+
+describe('Docsify public API', () => {
+  before(async () => {
+    server = http.createServer((request, response) => {
+      // You pass two more arguments for config and middleware
+      // More details here: https://github.com/zeit/serve-handler#options
+      return handler(request, response);
+    });
```
trusktr, line 31: This creates an HTTP server, when we run JSDOM it will fetch the `<script src="/lib/docsify.js"` from this local server.

test/unit/docsify.test.js
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  it('global APIs are available', async () => {
+    // const DOM = new (require('jsdom').JSDOM)(markup, {
+    const DOM = initJSDOM(markup, {
+      url: docsifySite,
+      runScripts: 'dangerously',
+      resources: 'usable',
+    });
```
trusktr, line 53: Here we instantiate `JSDOM` with the `url` of `http://127.0.0.1:3000`. When it evaluates the HTML inside of `markup`, it will load the `lib/docsify.js` file from `http://127.0.0.1:3000/lib/docsify.js`.

test/unit/docsify.test.js
```diff
@@ -0,0 +1,133 @@
...
+    // and the following APIs should be available:
+    expect(typeof DOM.window.Docsify).to.equal('object');
+    expect(typeof DOM.window.Docsify.util).to.equal('object');
+    expect(typeof DOM.window.Docsify.dom).to.equal('object');
+    expect(typeof DOM.window.Docsify.get).to.equal('function');
+    expect(typeof DOM.window.Docsify.slugify).to.equal('function');
+    expect(typeof DOM.window.Docsify.version).to.equal('string');
+    expect(typeof DOM.window.DocsifyCompiler).to.equal('function');
+    expect(typeof DOM.window.marked).to.equal('function');
+    expect(typeof DOM.window.Prism).to.equal('object');
```
trusktr, line 68: If `lib/docsify.js` was loaded successfully, then these tests will pass.

test/unit/docsify.test.js
anikethsaha: cool

anikethsaha:
@trusktr feel free to fix the merge conflict and merge this.

trusktr:
Yay. Ok tonight or tomorrow I'll fix the conflict and merge.

trusktr:
I fixed the conflict with package-lock.json, and also I opened https://github.com/docsifyjs/docsify/pull/1216 which should prevent conflicts from package-lock.

Also the e2e test failed. It is flaky. There's a race condition. I'll take a look soon. Let me re-run the tests...

build/css.js
```diff
@@ -31,7 +31,14 @@ fs.readdir(path.join(__dirname, '../src/themes'), (err, files) => {
...
             stylusCMD.on('close', (code) => {
-                console.log(`[Stylus Build ] child process exited with code ${code}`);
+                const message = `[Stylus Build ] child process exited with code ${code}`
+
+                if (code !== 0) {
+                  console.error(message);
+                  process.exit(code)
+                } else {
+                  console.log(message);
+                }
```
jamesgeorge007, line 41:
```suggestion
                if (code !== 0) {
                  console.error(message);
                  process.exit(code)
                }
                console.log(message);
```

build/build.js
```diff
@@ -64,56 +76,68 @@ const buildAllPlugin = function () {
...
+      .watch(['src/core', 'src/plugins'], {
+        atomic: true,
+        awaitWriteFinish: {
+          stabilityThreshold: 1000,
+          pollInterval: 100
+        }
+      })
+      .on('change', p => {
+        console.log('[watch] ', p)
+        const dirs = p.split(path.sep)
```
jamesgeorge007, line 111: What about the array destructuring approach?

trusktr: Sure! Feel free to make new PRs to clean up the code. I was aiming to leave code as-is except to add error handling.

trusktr:
tests passed. Merging!

PR #1193, opened 2020-05-23
ci: added codesandbox build preview for PR reviews

anikethsaha:


anikethsaha:
|**blocking**|
|-|

We need the cdn link for the published package so that we can use it with template. 

If we somehow solve that, I will change the `sandbox` property in `./codesandbox/ci.yml` to like this

```cjson
{
  "sandbox":  ["./docs"],
}
```

So that we can test the changes with our preview of live docs.

PR #1192, opened 2020-05-23
Add build error handling

trusktr:
**Summary**

Add error handling to build processes so they don't silently fail.

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix?
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [x] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

---

* [ ] DO NOT include files inside `lib` directory.

PR #1195, opened 2020-05-23
update src/core/index.js to export all global APIs, deprecate old globals…

trusktr:
...in favor of a single global `DOCSIFY`, and add tests for this

@anikethsaha This updates https://github.com/docsifyjs/docsify/pull/1192 to fix the issues I found there.

The build now works, and the change is backwards compatible: it leaves the old globals in place but additionally creates a new single global variable `DOCSIFY` that we should recommend in favor of the other multiple globals which we should deprecate in the docs.

I don't think this needs to be in a major version bump as people can still use the old globals, unaffected (as per the new tests in this PR). Then in the next major we can remove the old global. Plus we can add an `.esm.js` bundle and recommend for people to use that instead of the global version so that they don't pollute the global space.

This new `DOCSIFY` global has the `Docsify` constructor exposed, so we can start heading in a direction where we prefer people to do `new DOCSIFY.Docsify(options)` instead of `window.$docsify = options`. In a following PR, I can add the new behavior but keep the old `$docsify` options working for backwards compatibility without having to save it for a major bump.

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -9,6 +9,14 @@ const version = process.env.VERSION || require('../package.json').version
 const chokidar = require('chokidar')
 const path = require('path')
 
+/**
+ * @param {{
+ *   input: string,
+ *   output?: string,
+ *   globalName?: string,
+ *   plugins?: Array<import('rollup').Plugin>
+ * }} opts
+ */
 async function build(opts) {
   await rollup
     .rollup({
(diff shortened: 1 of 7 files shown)


trusktr:
> Semantic Pull Request — add a semantic commit or PR title

Ah, nice check! Let me fix that!

anikethsaha:
@trusktr can you submit a PR with updated docs where-ever needed.

PR #1207, opened 2020-06-09
Revert "Updated docs site dark and light mode with switch and redesigned search bar using docsify-darklight-theme"

anikethsaha:
Reverts docsifyjs/docsify#1182


anikethsaha:
@jhildenbiddle mentioned some valid points here https://github.com/docsifyjs/docsify/pull/1182#issuecomment-641127892

Issue #1173, opened 2020-05-12
There is an error in opening the mobile page of the onlyCover parameter


sy-records:
#### Steps to reproduce

1. Setting param

```javascript
window.$docsify = {
      onlyCover: true,
}
```

#### What is current behaviour

Should be removed, Otherwise, when you click it, the problem of Figure 2 will appear.

![image](https://user-images.githubusercontent.com/33931153/81659143-0507d680-946c-11ea-8e11-c4f646576ddc.png)

![image](https://user-images.githubusercontent.com/33931153/81659317-2ec0fd80-946c-11ea-9f7a-ad65a1dd0e6b.png)

#### What is the expected behaviour

![image](https://user-images.githubusercontent.com/33931153/81658985-dd187300-946b-11ea-82ed-65c9d2b6f7c9.png)

#### Other relevant information

- [ ] Bug does still occur when all/other plugins are disabled?

- Your OS: mac
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 4.11.3
- Docsify plugins: 

#### Please create a reproducible sandbox 

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/307qqv236)

anikethsaha:
can we keep the burgur icon and fix the sidebar so that when it is clicked, it slides as it suppose to be ?

else we can remove it 👍

sy-records:
What it does is load only the cover page when you visit the home page.

I'm not sure if I should remove the burgur icon when it exists.

sy-records:
https://simps.io/ For my site, I think it should be deleted.

sy-records:
But the sidebar style of the `docsify-dark-mode` theme also needs to be fixed.  
https://docsify.js.org/ website was affected.

anikethsaha:
ok cool , I didnt notice the `onlyCover` option. feel free to fix it.

> But the sidebar style of the `docsify-dark-mode` theme also needs to be fixed.
> https://docsify.js.org/ website was affected.

go ahead 👍

trusktr:
Makes sense. I added the `bug` label and put it in the `To do` column of the `4.x` project board. Let's triage issues as we go.

mohammedsahl:
This seems like a quick CSS fix. I can work on this. From my understanding, all that needs to be done is remove the burger icon (sidebar-toggle) from the cover when viewed on a mobile device. @sy-records

[...]

PR #1215, opened 2020-06-12
fix #1213 

### LINKED ISSUE #1213: Sidebar: Overlapped by cover page, toggled by screen width shortening
## Bug Report

When the screen width gets shorter than 768px some unusual UI updates take place. 

#### Steps to reproduce

This [video](https://youtu.be/V47A5XUsrzQ) better demonstrates the conditions needed to reproduce this
Posting screenshots would clog this issue 😅 

#### What is current behavior

- The sidebar is positioned under the cover page UI elements
- The side bar toggles on when shortening the width

#### What is the expected behavior

- The side bar should overlap the cover screen when the sidebar-toggle-button is pressed. 
- Shortening the screen width must not cause the sidebar must not toggle on

- Your OS: 
- Node.js version: v13.11.0
- npm/yarn version: 6.13.7
- Browser version: 83.0.4103.97
- Docsify version: 4.4.1 (CLI version)

mohammedsahl:
**Summary**
For devices having a width of `768px` or less the sidebar is overlapped by the cover page
This PR fixes that by sliding the cover page over, similar to the content page

The bug is reproduced by loading the page on a device browser having a screen size of `786px` or less
and clicking the `sidebar-toggle` button

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

**Before**
image
**After**
image

- [ ] Yes
- [X] No

- Fix #1213 

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

* [X] DO NOT include files inside `lib` directory.

### DIFF
--- a/src/themes/basic/_coverpage.styl
+++ b/src/themes/basic/_coverpage.styl
@@ -22,7 +22,6 @@ section.cover
     flex 1
     margin -20px 16px 0
     text-align center
-    z-index 1
 
   a
     color inherit
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -423,7 +423,7 @@ body.close
     left - $sidebar-width
     transition transform 250ms ease-out
 
-  .content
+  .content, .cover
     left 0
     max-width 100vw
     position static
@@ -441,6 +441,7 @@ body.close
   body.close
     .sidebar, .cover
       transform translateX($sidebar-width)
+      transition transform 250ms ease-out
 
     .sidebar-toggle
       background-color rgba($color-bg, 0.8)
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -439,7 +439,7 @@ body.close
     padding 30px 30px 10px 10px
 
   body.close
-    .sidebar
+    .sidebar, .cover
       transform translateX($sidebar-width)
 
     .sidebar-toggle
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -423,7 +423,7 @@ body.close
     left - $sidebar-width
     transition transform 250ms ease-out
 
-  .content, .cover
+  .content
     left 0
     max-width 100vw
     position static
@@ -439,7 +439,7 @@ body.close
     padding 30px 30px 10px 10px
 
   body.close
-    .sidebar, .cover
+    .sidebar
       transform translateX($sidebar-width)
 
     .sidebar-toggle
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -441,7 +441,6 @@ body.close
   body.close
     .sidebar, .cover
       transform translateX($sidebar-width)
-      transition transform 250ms ease-out
 
     .sidebar-toggle
       background-color rgba($color-bg, 0.8)


sy-records:
LGTM.👍🏻

jhildenbiddle:
Fix looks good!

trusktr:
@jhildenbiddle Did you check the menu animation (considering there's a new `transition` property)? For me the menu lags now when closing it (tested in Chrome). I think better to remove the animation tweak so we can merge the fix, then we can make animation tweaks in a new PR.

jhildenbiddle:
@trusktr --

Very strange. I'm not seeing any lag on the latest versions of Safari (Mac, iPhone, iPad), Chrome (Mac), or Firefox (Mac). I'm previewing the latest deployment here: https://docsify-preview-qbuv6mk3f.vercel.app/

Have you tried other browsers in hopes of isolating the issue to Chrome? If it is Chrome-specific, verify that Chrome is hardware accelerated by typing in `chrome://gpu/` in the URL bar. It's unlikely that this is the issue, but it's worth checking.

mohammedsahl:
@trusktr The transition was so that the cover "slides" into place rather than "snap". I've tried it on chrome for now.
@jhildenbiddle https://pastebin.com/Ec5ZyBcM. This is what I get from `chrome://gpu`. I do agree with @trusktr that there is a lag when closing the menu

jhildenbiddle:
Nice work, @mohammedsahl!

Still confused as to why I don't see the lag, but it's a moot point if you've solved the issue.

mohammedsahl:
I think there's a misunderstanding here, the lag issue is not entirely solved 😅
As it currently stands, we either have
- The cover page shift when the sidebar is toggled, done by adding `.cover` translation
- The sidebar overlaps as before, done by removing the `.cover` translation (the buggy landing page)

There is a third option where the user is not allowed to toggle the side bar when the cover page is displayed but that would, in my opinion, not be in line with the current design

@jhildenbiddle

sy-records:
#1173 I also feel that if the mobile terminal chooses to display the cover, it can not need the sidebar.

jhildenbiddle:
Apologies for the misunderstanding, @mohammedsahl.

So I dug into this a little bit...

The issue is caused the [docsify-dark-mode](https://github.com/Plugin-contrib/docsify-plugin/tree/master/packages/docsify-dark-mode) plugin's style sheet. Specifically, [this ruleset](https://github.com/Plugin-contrib/docsify-plugin/blob/be9a708c7d34df3b7b3eb66b8dbd264b02679727/packages/docsify-dark-mode/src/style.css#L6-L13):

```css
html,
body,
main,
.sidebar-toggle,
.sidebar,
aside {
  background: var(--docsify_dark_mode_bg);
}
```

The issue is that the CSS custom property `--docsify_dark_mode_bg` is `undefined` in normal mode, which is essentially the same as setting `background` to `none`. You can see this rule overriding the background color of the sidebar in using dev tools:

Screen Shot 2020-06-13 at 8 12 14 PM

An easy way to verify this is to block the following stylesheet using Chrome dev tools (Network tab, select the stylesheet, select "Block request URL", then refresh).

```html
<link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify-dark-mode@0.6.1/dist/style.css">
```

Three ways to fix:

1. Remove the `docsify-dark-mode` plugin and stylesheet. This is my recommendation for the same reasons I listed in #1182:

   > Generally speaking, we should be cautious about which (if any) third-party docsify-related addons we use on the public docsify site. Most visitors will view the site as an example of what they will get with an "out of the box" docsify installation. If after following our installation guide their site looks or behaves differently (which will be the case if docsify.js.org uses this theme), confusion can lead to issues being filed (which we have to manage) and an overall frustrating experience. We could try to address this by listing all of the plugins we use on the site at the end of our installation guide, but then we're essentially endorsing some addons over others which isn't great for the community. Unless there is legitimate need, I believe it's best to let docsify.js.org be a showcase primarily for code from the docsify team.

   There are also issues with this plugin+theme combo (poor readability due to low text/background color contrast of text, non-dark-mode search box and code blocks, breaks in IE11, possibly others but I haven't tested extensively). Even if these issue are addressed, I still think the above statement justifies removal of the plugin+theme on docsify.js.org.

   Screen Shot 2020-06-13 at 8 22 14 PM

2. Create a [docsify-dark-mode](https://github.com/Plugin-contrib/docsify-plugin/tree/master/packages/docsify-dark-mode) issue and wait for the author to fix the bug.

3. Modify our CSS (as we have been doing in this PR) to try and work around the issue with docsify-dark-mode. I recommend against this in favor of either of the options above.

Any opposition to just removing the docsify-dark-theme plugin and stylesheet?

(@trusktr, @anikethsaha)

anikethsaha:
> Any opposition to just removing the docsify-dark-theme plugin and stylesheet?

Removed already, https://github.com/docsifyjs/docsify/pull/1207,  need a release

also, the preview link should not have the `docsify-dark-mode` [plugin in it](https://github.com/docsifyjs/docsify/blob/develop/index.html).

mohammedsahl:
Thank you @jhildenbiddle for the investigative work 😄
Are there any places I need to look into to make the appropriate changes? I'd really like to increase my familiarity with the code base and hence make working with issues a little easier. I can get started on either of the three fixes mentioned above but I'd need to know which one to work on first 😅.

jhildenbiddle:
Apologies for the confusion, everyone.

There are actually two completely separate issues, both of which coincidentally present themselves in similar ways. To make things even more confusing, I was debugging the code on https://docsify.js.org/ (master branch) when I should have been debugging the previews generated for this PR. Sorry about that. Hopefully the following screenshots and explanations will help clarify both issues and my recommendation on how to wrap up this PR.

1. The first issue (as stated above) is/was the `docsify-dark-mode` theme which set the sidebar's background color to `undefined` (which is the same as setting it to `none`). Since this plugin was removed in #1207 we just need to publish these changes to address this issue. 

   Here is a screenshot from https://docsify.js.org/, which matches the "before" screenshot listed at the top of this PR (which is why I stated this was the _**original**_ issue):

   Screen Shot 2020-06-14 at 5 02 28 PM

   Removing the plugin addresses the missing background color as seen here:

   Screen Shot 2020-06-14 at 5 02 44 PM

   Good news? The background color is fixed. Bad news? The stacking order of the sidebar and coverpage content is wrong. On to the second issue...

2. The second issue is related to stacking contexts. When certain CSS properties (like `transform` or `opacity`) are applied to an element, the element is rendered as a "composite layer" that is accelerated by the GPU. These layers exist in a completely different stacking context than non-accelerated layers with a `z-index` applied. In this particular case, the sidebar's `translateX` value and the `<div class="cover-main">` element's `z-index` value cause them to be rendered in separate stacking contexts, with the sidebar being rendered below the `<div class="cover-main">` element. Because these elements are in separate stacking contexts, the order can't be fixed by simply adjusting `z-index` values. 

   The easiest fix is to simply remove the `z-index` value from the `<div class="cover-main">` because it isn't needed. 
   https://github.com/docsifyjs/docsify/blob/da9af18c4e4aa8b2c42773a01104f5a7c0a628a4/src/themes/basic/_coverpage.styl#L25

   This allows the sidebar element to render above the coverpage content as intended.

   Screen Shot 2020-06-14 at 5 03 00 PM

Summary: I believe if @mohammedsahl simply removes changes made thus far along with the `z-index` declaration listed above, both issues will be resolved.

Whew. :)

anikethsaha:
Thanks for the detailed explanation @jhildenbiddle . 
I believe as this is a small change, we can include that in this PR itself. 

@mohammedsahl can you add the fix here as mention above ?

trusktr:
I like the cover sliding to the right when the menu opens because it matches the same pattern as with the main content. I think it would be great to go with that solution (apart from fixing the background color).

The problem only happens in mobile; I can reproduce while in mobile mode in devtools or on my phone.

Guess what! I just remembered about the `touch-action` CSS property.  The browser is the one making the delay. It does this by default to see if the user will be performing a double click (in which case it performs a zoom).

I set `touch-action: none;` on the `button` for the menu, and that fixes the issue.

trusktr:
Turns out https://docsify.js.org has the same issue on mobile. I am happy to merge this then, but we should now also fix:

- background
- add `touch-action: none` (wdyt @jhildenbiddle?)

jhildenbiddle:
@trusktr --

I don't mind the sliding coverpage. I'd still recommend removing the `z-index`declaration on `.cover-main` for two reasons: 1) it isn't necessary and 2) the stacking context bug would be a hard one to track down if for some reason it became an issue again in the future.

@mohammedsahl -- 

I think this means you can leave the current edits in place, delete the following line, and we should be good:

https://github.com/docsifyjs/docsify/blob/da9af18c4e4aa8b2c42773a01104f5a7c0a628a4/src/themes/basic/_coverpage.styl#L25

trusktr:
I wasn't noticing this before, but on the lates vercel preview, https://docsify-preview-p88m1zgcc.vercel.app, I am not able to see the hamburger menu on mobile unless I scroll:

menu-hidden

trusktr:
Also for some reason on the latest Vercel preview, I no longer see the Translate menu at the top. Strange. I wonder why I the changes. I just checked the original Vercel preview, and I _do_ see the Translate menu at the top right of the screen, but I _do not_ see the hamburger menu without scrolling. Maybe I didn't notice before, but it looks like another regression.

Sorry for the noise, I'll update my review.

jhildenbiddle:
Not sure what's going on @trusktr. No issues on my end.

Let's get the changes I listed in [my comment above](https://github.com/docsifyjs/docsify/pull/1215#issuecomment-643958139) made to this PR before spending more time previews. I believe those changes take everything we've discussed into account. Ideally those changes address all issues so we can close the PR and move on to the next thing.

mohammedsahl:
@trusktr Here's something I'm not able to understand. 

Suppose we're scrolling the contents and we click the sidebar-toggle. Why doesn't the web page zoom out when I double-tap a non-link area of the contents. Since `.content` is also being translated by `$sidebar-width`, shouldn't it also lie outside the view-port? 

The current behavior (observed on the `develop` branch) is that clicking anywhere outside the sidebar closes the sidebar, which I think is what we're going for. Meaning, we'd want the cover-page to shift and clicking anywhere outside the sidebar, closes it

jhildenbiddle:
I think we should refocus on the stated issue: To fix the presentation of the sidebar at <= 768px. In that spirit, I believe the steps should be the following:


1. Remove the sliding coverpage effect (i.e. the changes thus far). If this change introduces a new issues (unwanted horizontal scrolling when the menu is open, which I wasn't aware of) then we should skip it so we can address the stated issue first. That's more important, IMHO.
1. Delete the following line:
   https://github.com/docsifyjs/docsify/blob/da9af18c4e4aa8b2c42773a01104f5a7c0a628a4/src/themes/basic/_coverpage.styl#L25

All other issues / enhancements (sliding coverpage, button delays, hidden content rendered outside of the viewport, etc.) are all outside the scope of this PR. That's not to say they aren't important, but this PR isn't the best place to explore or implement them. If we want to explore additional enhancements, let's file separate issue(s) so we can review and triage accordingly.

mohammedsahl:
Not exactly sure why the e2e tests are failing. Running `npm run test:e2e` locally results in all tests passing

anikethsaha:
the e2e CI is kind of moody 😄 
it needs re-run sometimes.

anikethsaha:
Thanks for your contribution 👍

trusktr:
@anikethsaha This wasn't ready yet. I had requested changes above. This causes the mobile site to zoom and scroll sideways, which we need to fix.

trusktr:
Oh wait, nvm, I see @mohammedsahl made it to the cover doesn't translate. :+1:

PR #1216, opened 2020-06-13
Delete package-lock.json

trusktr:
I think libraries should not include package-lock.json.

- The `npm` command does not publish them to npmjs.com, so end users don't get benefit from them.
- We shouldn't attempt to lock dependencies using shrinkwrap anyways (increases the chance of duplicate dependencies in end user projects, which can cause issues, I can provide some links if curious). 
- The main reason to remove package-lock.json is because as we authors develop Docsify, we can catch in-range breaking changes to any dependencies and fix them before users receive the errors. Having package-lock.json otherwise prevents us from catching those bugs sooner. 
- We can manually pin dependencies as a last resort in package.json if we really need to.

Wdyt? Any reason to keep it?


trusktr:
> If we are removing this, I guess we should remove the lockfile for `packages/docsify-server-renderer` as well and put them in `.gitignore`

good catch!

trusktr:
There we go. I added the config in `.npmrc` which tells `npm` not to generate `package-lock.json` files, but some tools (like Lerna) don't obey this option so in any other cases beside using `npm` the `.gitignore` addition will help.

trusktr:
Let me fix the ironic conflict with package-lock.json. :)

jhildenbiddle (dismissed):
Conceptually this seems okay to me. That said, I'm not familiar with some of the tools used in this repo (like Lerna) so I can't comment on if/how this might affect builds in different environments.

Looks like there are a few tests failing which will need to be addressed though.

anikethsaha:
I think there is some problem with CI related to monorepo (lerna).
Does lerna requires the lock file in order to have the same env ?

trusktr:
Ah, the tests fail because Lerna is running in `ci` mode (uses `npm ci` in the case of `npm`) which requires `package-lock.json`. ci mode is good for applications but not necessarily libraries (for the same above reasons).

I'll make an update.

PR #1229, opened 2020-06-18
[Poc] feat(hook): new hook updateNav for update navbar content

anikethsaha:
This is a working Proof of concept for the new hook `updateNav`

### How to use it. 

```js
  function(hook, vm) {
      hook.updateNav(function(navContent) {
        return `${navContent}\n- new link\n- new link 2`
      });
}
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,6 +55,9 @@
       formatUpdated: '{MM}/{DD} {HH}:{mm}',
       plugins: [
         function (hook, vm) {
+          hook.updateNav(function(nav){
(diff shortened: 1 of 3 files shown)


hamishwillee:
Is there any intent to get this in? 

The problem I have is that I do not want to include the translation path in my sources. If I do that, then my translators have to do make-work of manually updating the path in each URL. So what I want to do is read the current page path, get the translation, and insert is as a prefix in each of the links.

PR #1251, opened 2020-06-27
Update language-highlight.md

trusktr:
**Summary**

Add some details regarding syntax highlight.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

### DIFF
--- a/docs/language-highlight.md
+++ b/docs/language-highlight.md
@@ -22,8 +22,9 @@ For `prism-php.js`, it would be:
 
 ````
 ```php
-function getAdder(int $x): int {
-    return 123
+function getAdder(int $x): int 
+{
+    return 123;
 }
 ```
 ````
--- a/docs/language-highlight.md
+++ b/docs/language-highlight.md
@@ -8,4 +8,24 @@
 <script src="//cdn.jsdelivr.net/npm/prismjs/components/prism-php.min.js"></script>
 ```
 
+To use the new languages, make sure the code block label matches the part after `prism-` in the file name. FOr example, for `prism-bash.js` write a code block labeled with `bash` like this:
+
+````
+```bash
+echo "hello"
+```
+````
+
+?> Note that with GitHub-flavored markdown, `sh` and `bash` are effectively aliases of each other, but this is not the case with Prism. So using `sh` will not enable `bash` syntax in this case.
+
+For `prism-php.js`, it would be:
+
+````
+```php
+function getAdder(int $x): int {
+    return 123
+}
+```
+````
+
 ?> Check the [component files](https://github.com/PrismJS/prism/tree/gh-pages/components) list for more options.


docs/language-highlight.md
```diff
@@ -8,4 +8,25 @@
...
 
+To use the new languages, make sure the code block label matches the part after `prism-` in the file name. FOr example, for `prism-bash.js` write a code block labeled with `bash` like this:
+
+````
+```bash
+echo "hello"
+```
+````
+
+?> Note that with GitHub-flavored markdown, `sh` and `bash` are effectively aliases of each other, but this is not the case with Prism. So using `sh` will not enable `bash` syntax in this case.
```
jhildenbiddle, line 19: This is confusing for a few reasons:

1. GitHub-flavored markdown doesn't handle syntax highlighting (see [spec](https://github.github.com/gfm/)). I think you're referring to how GitHub handles markdown.
1. Docsify supports GitHub-flavored markdown (see the [Supported Markdown specifications](https://marked.js.org/#/README.md) for [marked](https://github.com/markedjs/marked))
1. [Prism's source](https://github.com/PrismJS/prism/blob/c6c62a69a06e80535c387f4041da80e23a3654f0/plugins/file-highlight/prism-file-highlight.js#L18) lists `sh` as an alias for `bash`.
1. [Prism's docs](https://prismjs.com/#supported-languages) list `shell` as an alias for `bash` (and `sh` is an alias for `shell`).

I suggest we just delete the `?> Note...` section.

docs/language-highlight.md
```diff
@@ -8,4 +8,25 @@
...
+
+For `prism-php.js`, it would be:
```
jhildenbiddle, line 21: Do we need two examples? If we want two, why not use one of the built-in languages (`css`, `html`, or `js`) and one that requires loading an additional language?

jhildenbiddle:
@sy-records I realize this has already been merged, but there are a few tweaks in here work making.

Also, the "At least 2 approving reviews are required to merge this pull request" rule generally means "two reviews not including your own if you are the person who submitted the PR".

sy-records:
Sorry, This was PR by @trusktr submitted, I modified a PHP syntax error.

I reviewed it with @Koooooo-7 ...😅

@trusktr Can you resubmit a PR to fix these problems?

sy-records:
Oh no, why is it submitted to the master branch? ? ?😱

jhildenbiddle:
Ahh. Sorry, @sy-records. I saw your edit and thought it was your PR.

Koooooo-7:
@jhildenbiddle  sorry for missing checking the details of documentation. 

@sy-records , we need revert this PR from `master`.

sy-records:
Let's submit another PR. @Koooooo-7 @trusktr

jhildenbiddle:
@Koooooo-7 -- No worries at all. That's what multiple reviewers are for. 😄

docs/language-highlight.md
trusktr: > GitHub-flavored markdown doesn't handle syntax highlighting

Maybe I should say "Markdown on GitHub".

trusktr: > Prism's source lists sh as an alias for bash.

When I add `prism-bash.js` and use `sh` as the syntax it doesn't highlight. Same with `shell`. `bash` works though. Is it new? Does it work for you? Maybe I need an update.

trusktr:
Whoops, I didn't check what branch it merged to when I opened it. I think by default GitHub sets the target branch to whatever the default branch is (should be develop). I'm not on the comp, I'll check later.

Sure, I can also submit updates.

trusktr:
Unless the PR author is unresponsive, I think we should let the PR author merge the PR. The reviewers should not merge it. The author might have ideas or want to change something.

docs/language-highlight.md
jhildenbiddle: I was incorrect about `sh` being an alias for `shell` and `bash` in the Prism source (it was late 🤷 ). Prism simply associates `sh` as a file extension for bash scripts.

The difference between GitHub and Prism highlighting is:

* Prism requires using the alias(es) listed in their [docs](https://prismjs.com/#supported-languages). For Bash, Prism lists only `bash` and `shell`. I can confirm both `bash` and `shell` work as expected in docsify-themeable, while `sh` does not work:

  Screen Shot 2020-06-28 at 11 24 09 PM

* GitHub (it appears) uses file extensions (like `sh`) in addition to the usual aliases. Notice how the same markdown from the screenshot above renders here on GitHub (specifically, the proper `sh` highlighting):

  ```bash
  # bash
  echo "hello"
  ```

  ```shell
  # shell
  echo "hello"
  ```

  ```sh
  # sh
  echo "hello"
  ```

   To confirm this, I tried to find another file type that Prism associates an extension with but does not list as an alias. Batch has a file extension of `bat` in the [source](https://github.com/PrismJS/prism/blob/c6c62a69a06e80535c387f4041da80e23a3654f0/plugins/file-highlight/prism-file-highlight.js#L19) but only lists `batch` as an alias in the [docs](https://prismjs.com/#supported-languages) so this is a good match. Here's how `batch` and `bat` render on docsify-themeable:

  Screen Shot 2020-06-29 at 12 20 17 AM

  And here's how that above markdown renders here on GitHub:

  ```batch
  :: batch
  @ECHO "HELLO"
  ```

  ```bat
  :: bat
  @ECHO "HELLO"
  ```

Summary: Docsify uses Prism aliases only, GitHub uses language aliases and extensions.

jhildenbiddle:
@trusktr -- How about something like this:

---

# Language highlighting

Docsify uses [Prism](https://prismjs.com) to highlight code blocks in your pages. Prism supports the following languages by default:

* Markup - `markup`, `html`, `xml`, `svg`, `mathml`, `ssml`, `atom`, `rss`
* CSS - `css`
* C-like - `clike`
* JavaScript - `javascript`, `js`

Support for [additional languages](https://prismjs.com/#supported-languages) is available by loading the language-specific [grammar files](https://cdn.jsdelivr.net/npm/prismjs@1/components/) via CDN:

```html
<script src="//cdn.jsdelivr.net/npm/prismjs@1/components/prism-bash.min.js"></script>
<script src="//cdn.jsdelivr.net/npm/prismjs@1/components/prism-php.min.js"></script>
```

To enable syntax highlighting, wrap each code block in triple backticks with the [language](https://prismjs.com/#supported-languages) specified on the first line:

````
```html
<p>This is a paragraph</p>
<a href="//docsify.js.org/">Docsify</a>
```

```bash
echo "hello"
```

```php
function getAdder(int $x): int 
{
    return 123;
}
```
````

The above markdown will be rendered as:

```html
<p>This is a paragraph</p>
<a href="//docsify.js.org/">Docsify</a>
```

```bash
echo "hello"
```

```php
function getAdder(int $x): int 
{
    return 123;
}
```

trusktr:
That sounds great! May as well have written that as a PR and I could've hit the accept button already. :P

jhildenbiddle:
@trusktr -- Will do. 😄

PR #1224, opened 2020-06-16
update readme, add Discord to site, other improvements/changes

trusktr:
**Summary**

- Clean up README
- update badges (looks better if they aren't too many at the top and not mixed rounded and square borders), added a Discord badge
- Replace Gitter links with Discord links.

Preview here:  https://github.com/docsifyjs/docsify/tree/update-readme

**TODO:**

- [ ] add Discord badge or widget in the docsify site
- [ ] simplify README: link to the site and docs rather than having small bits of information in the README that are duplicated on the website.
- [ ] If we care about the donate/sponsor stuff, then we should make it look better, or else remove it
- [ ] If we keep sponsor stuff, let's remove the sponsor badges from the top. The links will be completely obvious in the backers/sponsors section once we clean up the length of the README.
- [ ] If we keep sponsor stuff, let's decide how it will work. At the moment it links to different places, one place is opencollective (not under Docsify's own account) and the other place is QingWei-Li's personal donation page on his GitHub account, which has further links to other places. I think we should clean this up.
- [ ] autogenerate the contributors list (if not already?)

**inspiration**

As a source of inspiration, I'm looking at AssemblyScript's [README](https://github.com/AssemblyScript/assemblyscript/blob/master/README.md) (and others):

- It is simple and contains minimal info with links to the website
- The website contains the details
- The top of the README does not contain sponsor/donation info, but rather focuses on what the library is. We should do this too.
- Anything that is duplicated on website and README (namely sponsor stuff) is taken from the site and re-used in the README to avoid duplication of effort.

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [x] A convincing reason for adding this feature
- [x] Related documents have been updated
- [ ] Related tests have been updated

PR #1267, opened 2020-07-03
docs: add vercel logo

anikethsaha:
vercel is sponsoring our pro account. 
This is to add their logo in our docs and readme


docs/README.md
```diff
@@ -30,3 +30,10 @@ Please consider donating if you think docsify is helpful to you or that my work
 ## Community
 
 Users and the development team are usually in the [Gitter chat room](https://gitter.im/docsifyjs/Lobby).
```
sy-records, line 32: Should we change `Gitter` to `Discord`?

Koooooo-7: @sy-records  I remembered that @trusktr updated a new README.md, I thing it will be changed later.

sy-records: #1224

anikethsaha:
> Instead of embedding the SVG code, let's re-use the SVG file.

yea make sense

Issue #1252, opened 2020-06-27
How to create global vue variables in html and make it readable on .md files?


fmcdigital:
How can I make my markdown files to get my Vue variables that I registered in my .html file?

I don't want to register the `script` inside .md files, is it possible to register in .html?

```
<script>
  new Vue({
    el: '#main',
    data: {
      appName: '<b>' + $docsify.name + '</b>',
      fullImg: '<b>1.920 x 1.080 px</b>. Formato: <b>.jpg</b> com <b>72 dpi</b>',
      mobileImg: '<b>1.080 x 1.920 px</b>. Formato: <b>.jpg</b> com <b>72 dpi</b>',
      squareImg: '<b>1.200 x 1.200 px</b>. Formato: <b>jpg</b> com <b>72 dpi</b>'
    }
  })
</script>
```

jhildenbiddle:
Addressed in #1271. Waiting on review and merge. Keep an eye on #752 as well.

jhildenbiddle:
@fmcdigital --

Fixed in #1271. Note that if you intent to use raw HTML as data, you will need to use the `v-html` directive:

```markdown
# Text Data

## {{ appName }}

- {{ fullImg }}
- {{ mobileImg }}
- {{ squareImg }}

<script>
  new Vue({
    el: '#main',
    data: {
      appName: $docsify.name,
      fullImg: '1.920 x 1.080 px. Formato: .jpg com 72 dpi',
      mobileImg: '1.080 x 1.920 px. Formato: .jpg com 72 dpi',
      squareImg: '1.200 x 1.200 px. Formato: jpg com 72 dpi'
    }
  })
</script>
```

```markdown
# Raw HTML DAta

## <span v-html="appName"></span>

- <span v-html="fullImg"></span>
- <span v-html="mobileImg"></span>
- <span v-html="squareImg"></span>

<script>
  new Vue({
    el: '#main',
    data: {
      appName: '<b>' + $docsify.name + '</b>',
      fullImg: '<b>1.920 x 1.080 px</b>. Formato: <b>.jpg</b> com <b>72 dpi</b>',
      mobileImg: '<b>1.080 x 1.920 px</b>. Formato: <b>.jpg</b> com <b>72 dpi</b>',
      squareImg: '<b>1.200 x 1.200 px</b>. Formato: <b>jpg</b> com <b>72 dpi</b>'
    }
  })
</script>
```

Issue #1074, opened 2020-03-14
Dynamic content in Vue components


Fohlen:
## Bug Report

In the documentation it says that one can embed `Vue` instances. I tried building an interactive component for our documentation like so:

```markdown
# Servers

Thanks to [AngrySnout](https://github.com/AngrySnout/SauerTracker) we have excellent statistics for the game.

<div id="server">
    <ul>
        <li v-for="server in servers" :key="server.description"> {{ server.descriptionStyled }} </li>
    </ul>
</div>

<script>
  const API_URL = "https://tomaten.sauertracker.net/api/servers"

  new Vue({
    el: '#server',
    data: () => ({
        servers: []
    }),
    async mounted {}
        console.log("HEY")
        const response = await fetch(API_URL);
        this.servers = await response.json(); 
    }
  })
</script>
```

However when loading the page the content quickly flashes before getting "rendered away".
I have no idea what sorcery is going on behind the scenes but my best bet is that it's an unsupported feature.

#### Steps to reproduce

(see above)

#### What is current behaviour

(see above)

#### What is the expected behaviour

The app that is rendered on screen should allow for dynamic content.

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: OSX 10.15.3 (19D76)
- Browser version: Version 80.0.3987.132 (Official Build) (64-bit)
- Docsify version: 4.11.2
- Docsify plugins: search plugin

anikethsaha:
Sorry I am not familiar with vue so cant really help you here !

Feel Free to investigate and submit the fix.

_Also, try testing with older version and check in which it broke, it will help in investigation_

Koooooo-7:
Actually ,It seems some issues about the Vue supports in docsify.
Currently,  when u create ur own Vue instance,  you will not get the Vue instance in the `mounted` hook by using `this` (`this` directs to the `window` obj  in docsify now).

the Vue instance was covered in  `this.__EXECUTE_RESULT__`. 
there is showing how we [execute script](https://github.com/docsifyjs/docsify/blob/develop/src/core/render/index.js) in docsify.
```js
  // Execute script
  if (
    this.config.executeScript !== false &&
    typeof window.Vue !== 'undefined' &&
    !executeScript()
  ) {
    setTimeout(_ => {
      const vueVM = window.__EXECUTE_RESULT__;
      vueVM && vueVM.$destroy && vueVM.$destroy();
[...]

Issue #790, opened 2019-03-06
Vue not working with reactivity


degustadev:
Hi!

I tried adding the following:

```HTML
<div id="example-1">
  <button v-on:click="counter += 1">Add 1</button>
  <p>The button above has been clicked {{ counter }} times.</p>
</div>
```
```JS
new Vue({
  el: '#example-1',
  data: {
    counter: 0
  }
})
```

I added `<script src="//unpkg.com/vue/dist/vue.min.js"></script>` to `index.html` and `executeScript: true`.

But it doesn't work, documentation says Docsify supports Vue tho.

degustadev:
Related to: https://github.com/docsifyjs/docsify/issues/788

BillyLuo:
same question

jpkraemer:
does not work for me either

jhildenbiddle:
@degustadev --

Fixed in #1271. The example code you provided has been tested and works as-is.

PR #1271, opened 2020-07-05
Fix Vue Reactivity

### LINKED ISSUE #788: Vue code doesn't work
`msg` doesn't change when input value changed

### here is html
```html
...
<script src="https://unpkg.com/vue@2.6.7/dist/vue.min.js"></script>
<script src="https://unpkg.com/docsify/lib/docsify.min.js"></script>
<script>
  window.$docsify = {
    executeScript: true,
  }
</script>
...
```
### here is README.md
```md
# docsify

<div id="tony">
    <div><input v-model="msg"></div>
    <div>msg = {{msg}}</div>
</div>

<script>
  new Vue({
    el: '#tony',
    data: {
      msg: '123'
    }
  })
</script>
```

jhildenbiddle:
Fixes #788 #790 #1074 #1252 #1364 

- Fix inability to instantiate reactive Vue components by 1) handling each child of `#main` instead of `#main` itself and 2)  skipping elements that are already Vue instances
- Retain previous behavior of handling basic Vue rendering without the need for a markdown `<script>` tag.
- Add Vue3 support
- Update documentation and add live Vue examples
- Update `index.html` files to include Vue.js
- Update ~~Cypress~~ Jest+Playwright e2e tests

Summary: Docsify was mounting `#main` as a Vue component on every page change when it detected Vue.js, presumably to process basic Vue content without requiring a `<script>` tag in each markdown file. The issue is that this was happening *after* users had already instantiated child elements of `#main` as Vue components, destroying those instances.


src/core/render/index.js
```diff
@@ -48,22 +46,35 @@ function renderMain(html) {
...
-    }, 0);
-  } else {
-    this.config.executeScript && executeScript();
+  // Execute markdown <script>
+  if (this.config.executeScript || 'Vue' in window) {
+    executeScript();
+  }
+
+  // Handle Vue content not handled by markdown <script>
+  if ('Vue' in window) {
```
jhildenbiddle, line 59: This entire `if ('Vue' in window)` block exists to maintain the current behavior of handling basic Vue rendering without requiring a markdown `<script>` tag. My preference would be to remove this block in v5 and just require users to include `<script>` file in pages that contain Vue content instead of docsify assuming every page has Vue content.

trusktr: We need to double check this and add unit tests for the specific behavior we want.

The behavior here was bad before, it was executing script tags twice due to the two calls to `executeScript`. 

I was planning to to circle back to this particular piece of code based on changes in https://github.com/docsifyjs/docsify/pull/1091 (please see the last item in the TODO list there).

trusktr (dismissed):
This is a good change, but let's add unit tests.

anikethsaha:
> This is a good change, but let's add unit tests.

tests fixtures (e2e-cypress) have been updated. But yea unit/integration tests would be great to add. But is it possible to do unit testing for this change ? if so, it would be great to add.

src/core/render/index.js
```diff
@@ -48,22 +46,35 @@ function renderMain(html) {
...
+    const mainElm = document.querySelector('#main') || {};
+    const childElms = mainElm.children || [];
+
+    for (let i = 0, len = childElms.length; i < len; i++) {
+      const elm = childElms[i];
+      const isValid = ['SCRIPT'].indexOf(elm.tagName) === -1;
```
trusktr, line 65: This could be better: 

```js
const isValid = elm.tagName !== 'SCRIPT';
```

src/core/render/index.js
```diff
@@ -48,22 +46,35 @@ function renderMain(html) {
...
+      const isVue = Boolean(elm.__vue__ && elm.__vue__._isVue);
+
+      if (isValid && !isVue) {
+        new window.Vue({
+          mounted: function() {
+            if (this.$children.length === 0) {
+              this.$destroy;
```
trusktr, line 72: Should this be a method call, `this.$destroy();`?

How does this code work? I mean, assuming the method-call fix, what's the goal?

trusktr: @anikethsaha @Koooooo-7 @sy-records Let's be careful not to approve PRs without fully checking the code. Adding test code will veriy that things work as expected.

trusktr: The old code was destroying the old instance from the previous page using the `__EXECUTE_RESULT__` ref. How does this new implementation (assuming method-call fix) do it?

src/core/render/index.js
jhildenbiddle: @trusktr 

> The behavior here was bad before [...]

Not sure which behavior you're referring to, but I agree there are multiple aspects of the current behavior that should be revisited (see below). That said, my goal with this PR is to provide a fix (actually, an enhancement since this dynamic Vue content has never worked as far as I can tell) without a breaking change.

>  [...] it was executing script tags twice due to the two calls to executeScript

I'm not seeing this behavior in the latest version (4.11.4), verified (manually) using the following demo: https://codesandbox.io/s/docsify-790-9o1w8?file=/index.html

- When Vue.js is **not** detected, markdown `<script>`...
  - is not executed when `executeScript` is `false`
  - is not executed when `executeScript` is falsy or unspecified
  - is executed when `executeScript` is `true`
  - is executed when `executeScript` is truthy
- When Vue.js **is** detected, markdown `<script>`...
  - is not executed when `executeScript` is `false`
  - ⚠️ is executed when `executeScript` is falsy or unspecified
  - is executed when `executeScript` is `true`
  - is executed when `executeScript` is truthy

Markdown `<script>` tags are not being executed twice in any of the above scenarios. 

The ⚠️ icon indicates a what I believe to be a problematic behavior, however it is one that should remain in order to prevent a breaking change. This behavior is as described docsify's [`executeScript` documentation](https://docsify.js.org/#/configuration?id=executescript) documentation (although the phrasing is a little confusing):

  > ## executeScript
  > type: `Boolean`
  > Execute the script on the page. Only parse the first script tag(demo). If Vue is present, it is turned on by default.

  Note that I have made an update to this PR to match the above behavior (specifically, to allow `executeScript:false` to prevent markdown `<script>` from executing even if Vue is detected). Proof that testing is needed, so you'll get no arguments from me. :)

> We need to double check this and add unit tests for the specific behavior we want.

Yes we should add tests, but the behavior we want is the same behavior above _and_ the ability to render dynamic Vue content. Just clarifying that "the specific behavior we want" we want now and down the road may be different.

> I was planning to to circle back to this particular piece of code based on changes in #1091 (please see the last item in the TODO list there).

One less thing to do!

docs/index.html
```diff
@@ -2,30 +2,31 @@
...
-		nav.app-nav li ul {
-			min-width: 100px;
-		}
-	</style>
+  <style>
+    nav.app-nav li ul {
+      min-width: 100px;
+    }
+  </style>
 </head>
```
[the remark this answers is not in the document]
jhildenbiddle: @trusktr --
Prettier is installed as a devDependency with the project I see squiggly lines in the editor when a rule is broken, so my assumption is that it is working as configured. Since I didn't explicitly change these lines, I would expect an error to be shown both on my console and the lint checks here on GitHub. Since that isn't happening, it looks like we need to update our configuration to prevent this from happening.

src/core/render/index.js
jhildenbiddle: I was using `indexOf` in expectation that there might be other tags that should be skipped (e.g. `<br>`, `<link>`, `<meta>`). At the time, `<script>` was the only one Vue was complaining about. With this implementation it's obvious how to skip multiple tags if/when we need to do so.

Happy to make the change though.

src/core/render/index.js
jhildenbiddle: > Should this be a method call, this.$destroy();?

Yes. Each view instance contains its own `destroy()` method.

The `this.$children.length` checks for the existence of child Vue components (not child DOM nodes). It's harmless (because there will never be child components for docsify auto-mounted/destroyed Vue instances) but unnecessary. I'll remove it.

> How does this code work? I mean, assuming the method-call fix, what's the goal?

This is an alternative approach to docsify's "if we detect Vue, we'll automatically process Vue content without you having the add a <script> tag and manually call `new Vue()` in your markdown files" behavior. 

Previously, docsify handled this by creating a Vue instance out of the `#main` element on each page load. The problem with this approach is that it makes it impossible to render dynamic Vue content because either 1) dynamic Vue content is handled first by processing markdown `<script>`, then docsify tramples it when it mounts `#main` as a Vue instance, or 2) docsify mounts `#main` as a Vue instance first, which breaks Vue because docsify is calling `new Vue()` without the properties (`data`, `methods`, etc.) needed to initialize the dynamic Vue content contained in the markdown. It's a chicken-or-the-egg issue.

This new approach allows docisfy to continue automatically processing Vue content while allowing dynamic Vue content to exist. Instead of mounting the `#main` element as a Vue instance, docsify iterates over the children of `#main`, mounts each element that is not already a Vue instance, then immediately destroys the instance because rending is complete. There is no need to retain Vue instance references (like `__EXECUTE_RESULT__`) because these auto-generated Vue instances have no state, no methods, no hooks, etc. For this type of content, Vue is just acting as a very basic template engine. See the [Basic Rendering](https://docsify-preview-i1zwgvmdb.vercel.app/#/vue?id=basic-rendering) section of the updated docs for examples.

docs/index.html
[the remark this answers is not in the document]
anikethsaha: I think we should reconsider the eslint + prettier configuration, I think my workspace is creating this issue. Happened earlier as well. 

I am not sure why it's happening, we do have `.vscode` config which allows everyone to run prettier on save.

src/core/render/index.js
anikethsaha: if we are going to have other tags as well. It does make sense to go with arrays. 

I was thinking of using `Sets` as they are a bit faster.   

thoughts ?

src/core/render/index.js
```diff
@@ -62,18 +62,38 @@ function renderMain(html) {
...
+        }
+      }
+      // Vue 2
+      else if (vueVersion === 2) {
+        const isAlreadyVue = Boolean(elm.__vue__ && elm.__vue__._isVue);
+
+        if (!isAlreadyVue) {
+          new window.Vue({
+            mounted: function() {
+              this.$destroy();
```
jhildenbiddle, line 93: @trusktr --

Good catch on the missed parens on `this.destroy`. Fixed.

docs/index.html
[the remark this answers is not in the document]
jhildenbiddle: Marking as resolved since 1) the changes here align with our prettier settings and 2) getting everyone else's prettier configuration working properly is a separate issue.

src/core/render/index.js
jhildenbiddle: @anikethsaha: No reason to focus on speed improvements for fixed arrays with one item. Also, we can't use `Set` natively due to our browser compatibility list without adding a significant amount of polyfill code.

src/core/render/index.js
jhildenbiddle: @trusktr: `this.$destroy` updated to `this.$destroy()` here: https://github.com/docsifyjs/docsify/pull/1271/commits/b135f8417d9b27b58a545e9f792e70d7b319cc15#diff-250dfe7435d59fa33e80765c47190dd5

src/core/render/index.js
anikethsaha: Make sense

anikethsaha (approved):

anikethsaha (approved):

sy-records (approved):

PR #1227, opened 2020-06-17
Fix validating remote content 2

trusktr:
**Summary**

Expands on stuff from https://github.com/docsifyjs/docsify/pull/1128

- [x] uses DOM APIs (JSDOM in Node.js and SSR, otherwise native APIs) for the `isExternal` function
- [x] consolidate `isExternal` so all code imports the same version
- [ ] add tests for isExternal and DOMPurify (WIP)
- [ ] cleanup

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [x] tests
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

#1126

You have tested in the following browsers: (Providing a detailed version will be better.)

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [x] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

PR #1276, opened 2020-07-08
Jest + Playwright Testing

### LINKED ISSUE #1217: add tests!
This is a meta issue. We should put some focus on stabilizing the source by adding unit and e2e tests. Test infrastructure is already in place thanks to @anikethsaha.

We can create separate issues for testing specific parts, and assign to ourselves so other will know what to work on without duplicating efforst.

jhildenbiddle:
**Summary**

- Added Jest test environment for unit and integration tests
- Added Playwright test environment for e2e tests
- Removed previous unit and integration test environments (mocha + chai + jsdom)
- Removed previous e2e test environment (Cypress.io)
- Ported existing unit, integration, and e2e tests of value to new environments

**Details**

An implementation of Jest+Playwright. This setup offers a number of advantages over the existing test structures (mocha+chai+jsdom & Cypress):

- Ability to perform unit, integration, and e2e tests with one test configuration.
- Separate test environments for unit + integration (Jest+jsdom) and e2e (Jest + Playwright) tests.
- Testing in real browsers (Chromium, Firefox, and Webkit) for e2e tests.
- Significantly faster and easier to create complex tests that require custom docsify configurations and site content thanks to a new `docsifyInit()` helper. This abstraction allows new docsify instances to be dynamically generated declaratively with a fraction of the code that would otherwise be required. It also removes the need to manage a large collection of fixture files (although you can use them if you prefer).
- Documentation content is mapped to `/docs/` on the local testing server, making it easy to use our documentation as fixtures without having to duplicate the content.
- The local `./lib/` directory is also mapped to `/lib/` on the local testing server, making it easy to use the latest docsify JS and CSS.
- Example tests provided to help reduce the learning curve for new contributors

Changes have been made to the GitHub workflows to accommodate these changes:

- A new "Build & Test" workflow has been added. This workflow runs all tests in a single workflow to prevent separate checkout+build operations for each test type. It updates the test matrix to build and test docsify on Linux (`ubuntu-latest`), macOS (`macos-latest`), and Windows (`windows-latest`) on the three latest LTS versions of Node (10, 12, 14).
- The linting workflow has been updated to prevent building docsify, as there is no need to perform a build to link source code (the build output is ignored by eslint).

Part of the motivation for doing this work was to allow me to create proper tests for #1271. The other part was to try and clean up our test configuration so that we can offer a reliable and easy-to-learn environment for devs to work in. I looked at our existing test structures and realized there was no way for me do what I needed to do easily without creating complex tests with repeated boilerplate code and a large number of fixtures. Specifically, testing #1271 requires multiple docsify configurations that must be run in a browser with and without additional third-party scripts (vue.js). With this setup, every custom docisfy test is between 8 and 11 lines of code and zero test-specific file fixtures.

A few tips for reviewers:

1. All Jest and Playwright code is isolated in a  new `./test/` directory. This means this PR can safely be merged and it will not affect existing tests. 
1. Start by reviewing `/test/README.md`. This file provides an overview of the test environments, global variables, and CLI commands available.
1. Review `./jest.config.js` and note that we are using Jest [`projects`](https://jestjs.io/docs/en/configuration#projects-arraystring--projectconfig) to simplify unique environment configurations for unit, integration, and e2e tests.
1. Review `./test/helpers/docsify-init.js`. This helper greatly simplifies the process of creating dynamic docsify sites for individual tests.
1. For Jest (and jsdom) testing, review `/test/unit/example.test.js` and `/test/integration/example.test.js`.
1. For Jest + Playwright testing, review `/test/e2e/example.test.js`.
1. I've created real-world tests for vue.js reactivity (the reason #1271 exists). These are available for review in `/test/e2e/vue.test.js`;
1. VSCode debug configurations have been added to make testings a single test or file easier. You can access these by clicking the debug icon on VSCode's sidebar (or pressing SHFT+CMD/CTRL+D). The two tests currently in place are:
   - Run current test file
   - Run selected test name (requires highlighting `describe()` or `test()` name)
   - Update current test file snapshot(s)
   - Update selected test name snapshot(s) (requires highlighting `describe()` or `test()` name)

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [x] Build-related changes
- [x] Other, please describe: Automated Testing

- [ ] Yes
- [x] No

Fix #1217 

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

- [x] Related tests have been updated


trusktr:
Nice! I will take a look this weekend!

jhildenbiddle:
Made a few updates:

- Add separate folders for e2e, integration, and unit tests 
- Add Jest projects for e2e, integration, and unit tests in jest.config.js
- Add VSCode debug configurations for running tests, running current test file, and running current test by name
- Configure /tests/e2e/ tests for Jest+PlayWright environment
- Configure /tests/integration/ tests for Jest environment
- Configure /tests/unit/ tests for Jest environment
- Configure eslint for e2e, integration, and unit test environments
- Move e2e configuration files into e2e/config folder
- Update e2e example tests
- Update unit example tests
- Update HTML fixtures
- Update docsifyInit() helper

.vscode/launch.json
```diff
@@ -0,0 +1,34 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
sy-records, line 4: Is it necessary to keep this?

sy-records:
image
Looks good.

This script doesn't exist, does it?

```shell
npm run test:jest-example
```

.vscode/launch.json
jhildenbiddle: It's default text created by VSCode when a new launch.json file is created (I didn't write it).

sy-records: Well, I don't use vscode, I use JetBrains's

jhildenbiddle:
@sy-records --

> This script doesn't exist, does it?
> 
> ```shell
> npm run test:jest-example
> ```

That test was removed in the last batch of updates, so it looks like you need to pull to get the latest. That will allow you to see the changes listed [here](https://github.com/docsifyjs/docsify/pull/1276#issuecomment-657943854).

jhildenbiddle:
Thanks for taking to the time to review, @sy-records. Much appreciated!

tests/unit/example.test.js
```diff
@@ -0,0 +1,34 @@
...
+      return addends.reduce(
+        (accumulator, currentValue) => accumulator + currentValue
+      );
+    }
+    const testValue = add(1, 2, 3);
+
+    expect(testValue).toBe(6);
+  });
+
+  test('dom manipulation (jsdom)', async () => {
```
trusktr, line 17: If we can run unit tests in a browser, we can remove jsdom.

Koooooo-7:
> Configure /tests/integration/ tests for Jest environment.

I found the there is the config in `jest.config.js` but I didn't find this folder.

jhildenbiddle:
NOTE: Updated **Summary** in the first comment to better reflect the latest updates.

jhildenbiddle:
> Sorry for the delay.

No worries. Appreciate you finding the time during the work week.

> Few non-blocking suggestion/question
> 
> * is it safe to remove the cypress now ?

In my opinion, yes. The tests currently in Cypress just verify that every link in the sidebar of the docs site is as expected. While somewhat useful, the test are brittle because they rely on screenshots instead of simply verifying that links exist and work as expected. If, for example, a font-size were to change, every one of these tests would fail unnecessarily.

I can recreate these tests using Playwright if we want to continue verifying links on the doc site. That way we can remove Cypress and not lose any of our current tests.

> * can we add the new script in the github action

Yep. I think so. If this just involves editing `/.github/workflows/e2e.yml`, I'm happy to make this change as well.

> * not sure what is `unit` in the `tests`. I couldn't notice any jsdom imports, are those just mockups ?

JSDom is bundled with Jest, so no need for a separate import.

.vscode/launch.json
trusktr: I think we should not commit this file. The `.idea` folder is currently ignored, which serves the same purpose as `.vscode` but for JetBrains IntelliJ users.

This file may vary from user to user.

jhildenbiddle:
> Due to my local modules installed broken in VScode, I can't check all the tests yet.
> 
> > Configure /tests/integration/ tests for Jest environment.
> 
> I found there is the config in `jest.config.js` but I didn't find this folder.
> did I miss something?

Looks like I forgot to add a folder for integration tests. It may be easier to just remove the integration test setup for now since 1) we don't have any existing integration tests and 2) we are probably better served by focusing on e2e tests initially to get wider coverage.

anikethsaha:
> I can recreate these tests using Playwright if we want to continue verifying links on the doc site. That way we can remove Cypress and not lose any of our current tests.

Awesome,  I think this would be good. THough I think this is good to do in separate PR ?

> Yep. I think so. If this just involves editing `/.github/workflows/e2e.yml`, I'm happy to make this change as well.

yes, it just includes adding the npm script in this file. it would be great to see this in the CI. 

> JSDom is bundled with Jest, so no need for a separate import.

😄 , Ohh,,....that's great

packages/docsify-server-renderer/index.js
```diff
@@ -16,7 +16,7 @@ function cwd(...args) {
 
 function isExternal(url) {
   let match = url.match(
-    /^([^:\/?#]+:)?(?:\/\/([^\/?#]*))?([^?#]+)?(\?[^#]*)?(#.*)?/
+    /^([^:/?#]+:)?(?:\/\/([^/?#]*))?([^?#]+)?(\?[^#]*)?(#.*)?/
```
trusktr, line 19: This will be deleted in https://github.com/docsifyjs/docsify/pull/1227

src/core/fetch/index.js
```diff
@@ -22,7 +22,7 @@ function loadNested(path, qs, file, next, vm, first) {
 
 function isExternal(url) {
   let match = url.match(
-    /^([^:\/?#]+:)?(?:\/\/([^\/?#]*))?([^?#]+)?(\?[^#]*)?(#.*)?/
+    /^([^:/?#]+:)?(?:\/\/([^/?#]*))?([^?#]+)?(\?[^#]*)?(#.*)?/
```
trusktr, line 25: And in https://github.com/docsifyjs/docsify/pull/1227 there is not duplicate code (not two `isExternal` definitions).

trusktr:
I see Jest was set up to run unit tests, but these unit tests aren't running inside playwright. We already have Mocha for unit tests, where we have more control and flexibility over how we set up JSDom. We can also run playwright there just like we can in Jest. Can we use the existing Mocha set up for this? Here's the first tutorial on [running playwright in Mocha](https://medium.com/@rogger.fernandes10/e2e-tests-integrating-microsoft-playwright-with-mocha-and-chai-cb37016b63c3) that Google gave.


> > JSDom is bundled with Jest, so no need for a separate import.
> 
>  , Ohh,,....that's great

Separate imports are good. They tell you where stuff comes from. In my opinion it is better to have the flexibility we have with current Mocha setup. With Mocha we explicitly control JSDOM. #1227 is manually taking advantage of jsdom for server testing, for example.

---


I think if we're going with playwright, we should delete Cypress instead of having both. We should move any tests that we think are valuable over to playwright. I wouldn't say get rid of the menu tests, but move them over (and change how they're implemented if nedded) so that we at least cover the original intent.

Sidenote, the cypress tests take so long because the half-second delay between each menu item test.

---

Also how do we do snapshot testing in Playwright? I think this PR should cover that as well, even if it's only one snapshot of one page just to get us started (and testing the menu items doesn't necessarily have to be with snapshots as you said).

Sidenote, I think rather than snapshotting for each menu item, we can snapshot for bigger overall things (like the whole layout). We can take a snapshot of the top of the page, and a snapshot of the bottom of the page. That stuff should not change. What we can do is perhaps make a Lorem Ipsum page, and snapshot with that, so it won't matter if content changes (we don't have to test changing content with snapshots, we just need to test features).

.vscode/launch.json
jhildenbiddle: @trusktr -- Standardizing debug configurations for IDEs is more akin to standardizing NPM and CI scripts than user preferences. Developers don't want to type `./node_modules/.bin/someBinary -a arg1 -b arg2 -c var3` for frequently run tasks, so we simplify by using `npm run something`. Similarly, I don't want to type  `npm run tests:jest -- -i /path/to/file.test.js -t "name of describe of test block"` just to run a single test, so running a debug script from the context of a file (so I don't have to type the path) and/or with a test name selected (so I don't have to type the test name) means I can just press a button to run a single test, making testing easier and more efficient. Win win.

There is no reason to omit these kinds of project-specific files from the repo because they should not vary between users. That's the whole purpose behind storing their configuration as files that can be managed by source control instead of storing them as user/app preferences. They should be managed like any other change to the project: users can add/modify them, submit changes as PRs, then they are reviewed and (hopefully) merged.

tests/unit/example.test.js
jhildenbiddle: JSDom is bundled with Jest, so we can't remove it. It's there if you want to use it, but you don't have to.

anikethsaha:
> A small suggestion, the default `8080` port is conflict with another apps server easily(I met it 😅 ).
> If there would change to another port like `8888` for avoiding unnacessary exception at e2e tests.

@jhildenbiddle does playwright supports automatically jumping to other ports if the port is not free?

packages/docsify-server-renderer/index.js
jhildenbiddle: Sounds good. I was just fixing the unnecessary escape character. Would you prefer I leave as-is or undo this change?

src/core/fetch/index.js
jhildenbiddle: Same as above: leave as-is or undo this change?

tests/unit/example.test.js
jhildenbiddle: > @jhildenbiddle yep, I see, I just make sure the complete folder u mentioned.
> per to the test script tests, it works well. cool !
> 
> A small suggestion, the default `8080` port is conflict with another apps server easily(I met it 😅 ).
> If there would change to another port like `8888` for avoiding unnacessary exception at e2e tests.

Makes sense I'll make the change.

Thoughts on `8181`, `8888`, `3000`, or `3001`?

Koooooo-7: > Thoughts on 8181, 8888, 3000, or 3001?

IMO,  any one of them is ok,except 3000.
cuz docsify default port is 3000.

jhildenbiddle:
> I see Jest was set up to run unit tests, but these unit tests aren't running inside playwright. 

This is intentional. Since there's no controversy using Playwright for e2e tests, the hope is that limiting the use of Playwright to e2e tests will allow the PR to be merged sooner and with less debate. We can revisit this decision to only use playwright for unit and integration tests as well, but I'd vote to have this discussion after Playwright is in place so we can review and discuss separately.

>We already have Mocha for unit tests, where we have more control and flexibility over how we set up JSDom. We can also run playwright there just like we can in Jest. Can we use the existing Mocha set up for this?

We can, but like all things in life there are pros and cons to doing so. 

Mocha and Jest both great, and each has their own set of pros and cons. As test runners, Mocha and Jest (which uses Jasmine under the hood) are remarkably similar. Mocha is what I've always used on past projects and I've never had an issue with it. Jest is great too, and it has [become increasingly popular](https://www.npmtrends.com/jest-vs-mocha) largely due to the fact that it includes everything most people need to perform a wide variety of tests with little-to-no configuration necessary ("convention vs. configuration", as they say). This is the reason why I decided to use Jest instead of leveraging the existing mocha setup in `/test/`. With Jest, we get assertions, mocks, spies, screenshots, code coverage, parallel task runs, etc. all from one vendor along with great documentation and an active community. We can of course get the same features with a custom setup using mocha+chai+sinon+instanbul+various plugins, but that means a more complex setup that is likely unnecessary for a relatively simple project like docsify. Yes, mocha is more flexible, but favoring "flexibility because someday we might need it" over simplicity and reliability comes at a cost, especially for open source projects like docsify that have contributors of various experience levels contributing to the project.

All that being said, if the other maintainers strongly prefer mocha, I can explore a mocha+chai+playwright setup as well. It won't be an apples-to-apples setup though, as the playwright team has clearly put more effort into integrating Playwright with Jest: https://github.com/playwright-community.

Let's discuss Jest vs Mocha+(Stuff) separately so this PR thread doesn't get too crazy. I'll ping @trusktr directly since it seems like he has the vested interest.

> Separate imports are good. They tell you where stuff comes from. In my opinion it is better to have the flexibility we have with current Mocha setup. With Mocha we explicitly control JSDOM. #1227 is manually taking advantage of jsdom for server testing, for example.

People champion flexibility until things get complicated and then they long for simplicity.

I don't think the fact that we have mocha in place today is reason enough to favor it over Jest. We have Cypress in place too, and there doesn't seem to any concerns with removing it in favor of something that is arguably better. We have very few mocha+chai tests in place, and of those tests most are trivial to port to Jest, better handled by e2e tests, or (arguably) of little value and could be removed.

I do appreciate the fact that some tests exist in #1227 that use a custom jsdom configuration (which requires extra work in Jest) so I don't want skip over that. That said, a few tests in an uncommitted PR should not drive the larger test framework decision, nor should we assume that a custom jsdom configuration is required to test docsify's SSR functionality (I am 100% certain it is not).

As mentioned above, let's discuss offline and figure out the best way to move forward. FWIW, my comments here are intended push back on the idea of keeping what we have just because it's already in place, not so much against mocha itself.

> I think if we're going with playwright, we should delete Cypress instead of having both. We should move any tests that we think are valuable over to playwright. I wouldn't say get rid of the menu tests, but move them over (and change how they're implemented if nedded) so that we at least cover the original intent.

Agreed. This is the plan. I'm happy recreate the existing Cypress tests in playwright.

> Sidenote, the cypress tests take so long because the half-second delay between each menu item test.

It's also slow because the current tests rely on screenshots, which for the kind of tests being done makes testing both unnecessarily brittle and slow. We don't need screenshots to verify that links are available in the DOM.

> Also how do we do snapshot testing in Playwright? I think this PR should cover that as well, even if it's only one snapshot of one page just to get us started (and testing the menu items doesn't necessarily have to be with snapshots as you said).

Snapshots are baked into both Jest and Playwright, making snapshot testing easy peasy. Happy to add an example test. They are *not* baked in to mocha, so this would require a separate add-on. Not a big deal, but just one example of how Jest + Playwright keeps things nice and simple.

> Sidenote, I think rather than snapshotting for each menu item, we can snapshot for bigger overall things (like the whole layout). We can take a snapshot of the top of the page, and a snapshot of the bottom of the page. That stuff should not change. What we can do is perhaps make a Lorem Ipsum page, and snapshot with that, so it won't matter if content changes (we don't have to test changing content with snapshots, we just need to test features).

Exactly. Snapshot testing should be used sparingly though, because relying on it can make tests brittle. IMHO, e2e tests should generally favor testing the DOM instead of comparing screenshots, but larger screenshot tests should be used to get larger coverage in a single test. For example, DOM-based tests should be used to verify that docsify is rendering individual elements properly, but full-page screenshots of a lorem-ipsum docsify site can be used to verify that each theme is rendering as expected.

jhildenbiddle:
c

> > A small suggestion, the default `8080` port is conflict with another apps server easily(I met it 😅 ).
> > If there would change to another port like `8888` for avoiding unnacessary exception at e2e tests.
> 
> @jhildenbiddle does playwright supports automatically jumping to other ports if the port is not free?

@anikethsaha -- The test server is just live-server (see [`jest.startup.js`](https://github.com/docsifyjs/docsify/blob/jest-playwright/tests/e2e/config/jest.setup.js)). Not sure if automatically adjusts port settings if the one specified is in use, but I can look into it. For now it's easy enough to adjust though. I'm thinking of using 3001 since the `npm run serve` script runs a live-server instance on port 3000.

anikethsaha:
Cool, let's use some uncommon port number. 

I don't have any queries now. 

+1 for me. 

I will do a final review soon.

packages/docsify-server-renderer/index.js
trusktr: All good, I can resolve the conflict later when it happens

src/core/fetch/index.js
trusktr: Same, I'll fix the conflict

.vscode/launch.json
trusktr: Someone may not want to run in `integratedTerminal` for example. Maybe VS Code shouldn't mix personal configs with actually-useful/convenient project configs. Not sure why they do that. F.e. why should the software project dictate if the test runs in `integratedTerminal` or not?

But I do see the convenience (choose a file, highlight text with cursor, hit the button and it runs only that test). No harm in keeping it.

trusktr: We can ensure the project's required steps (f.e. `npm test`) are not in an editor-specific config (which is currently the case), and reserve the editor configs for conveniences only.

tests/e2e/config/jest-playwright.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  // prettier-ignore
+  browsers: [
+    'chromium',
+    'firefox',
+    'webkit',
```
trusktr, line 6: @jhildenbiddle What OS are you in? Does `webkit` testing work in Linux? I'm in Linux. Haven't tried it yet.

package.json
```diff
@@ -31,6 +31,10 @@
     "test": "mocha ./test/**/*.test.js",
```
trusktr, line 31: If we can get existing tests (those we wish to keep) working with Jest/Playwright, I think it would be fine to update the `"test"` script move the tests over in this PR, if you want to.

I'd like to see how the whole picture works, otherwise if we merge something now we're making a partially-informed decision (even if we're leaning that way).

src/plugins/emoji.js
```diff
@@ -1799,6 +1800,7 @@ const AllGithubEmoji = {
   zombie_woman: 'unicode/1f9df-2640',
   zzz: 'unicode/1f4a4',
 };
+/* eslint-enable camelcase */
```
trusktr, line 1803: Sidenote, we need to make a round of lint fixes and get them all finally done with. The warnings that appear below are annoying, etc.

tests/e2e/config/jest-playwright.config.js
trusktr: It failed for me. Looks like we need some conditionals here depending on OS.

failed-test

tests/e2e/helpers/docsifyInit.js
```diff
@@ -0,0 +1,131 @@
...
+  if (settings.waitForSelector) {
+    await page.waitForSelector(settings.waitForSelector);
+  }
+
+  // Debug
+  if (settings.debug) {
+    await jestPlaywright.debug();
+  }
+
+  return Promise.resolve();
```
trusktr, line 128: This seems to achieve mostly the same thing as just loading a `.html` file would. 

Wouldn't it be simpler to to just load an HTML file from the `live-server`, for example, then run the tests based on that? I think it would be easier to understand.

trusktr: Or to load the HTML content from a template string, for example?

Not pushing back on merging though.  The usage of `docsifyInit` on the outside is simple. Just give it some content, then test the markdown output for example. It's only on the inside where it seems complex with manually handling browser routes, etc.

tests/unit/example.test.js
trusktr: > FWIW, I'd prefer if folks focused on e2e tests in general, since we get the most bang-for-our-buck there. We can test 50 function and still not guarantee that a site renders properly. For example, a single e2e test can answer the question "is the sidebar doing X when setting Y is true` much faster and easier than multiple unit or integration tests.

We need both I think. There's reason for each (which we should all learn about so we test things the right way).

For example, in #1227 there is [this test](https://github.com/docsifyjs/docsify/pull/1227/files#diff-b26412a9e2fa48d069811ef18906adf3R8) that tests that `isExternal` does what it actually should do. I specifically put test cases there that I believe the current `isExternal` will fail at (as well some others). Unit tests are good there, and it is probably something that we're unlikely to catch with e2e tests.

.vscode/launch.json
jhildenbiddle: @trusktr --

Agreed on not putting anything critical in editor-specific scripts/configs. These are convenience only. If folks using webstorm for example want to check in the `.idea` folder, that's fine with me as long as I won't be required to use webstorm to perform some critical function (and you don't do something silly like for an editor theme for everyone else).

As for integrated vs external terminals, this is where I think individuals are free to modify files in their local dev environment and tell git to ignore their own changes to those files. I went with the integrated terminal because it's the default selection and it seems the least intrusive, but folks can change it locally (or if everyone wants to use an external terminal, just file PR and we'll discuss).

tests/e2e/config/jest-playwright.config.js
jhildenbiddle: @trusktr -- macOS. I agree that we should test this on Linux and Windows before considering it ready to go. Let me do a little research on my end regarding Linux and Webkit testing. It will require me to get a Linux VM up and running, but that's a useful this to have anyway.

What Linux distro are you running?

src/plugins/emoji.js
jhildenbiddle: Agreed.

tests/e2e/helpers/docsifyInit.js
jhildenbiddle: @trusktr --

An HTML file is being loaded. By default, `docsifyInit()` loads [`docsify-init.html`](https://github.com/docsifyjs/docsify/blob/jest-playwright/tests/e2e/fixtures/docsify-init.html)) and then modifies it based on the settings you provide. If you prefer to load your own HTML file, that's fine too. The goal is remove the need to create a large number of file fixtures that vary slightly from each other to test, for example, a single configuration option. If you want prefer file fixtures, you're free to use them as well. If you prefer to skip `docsifyInit()` altogether and just use Playwright comments like `await page.goTo(`{serverURL}/path/to/my-index.html`), that's fine too.

tests/unit/example.test.js
jhildenbiddle: @trusktr --

Agreed. Those tests are unit tests, not e2e tests. The follow-up to that statement is that unit tests won't determine if `isExternal()` will fail in Firefox or Webkit because those unit tests are tested in Node (which, along with Chrome, use the V8 JavaScript engine). An e2e test may not cover all scenarios like unit tests do, but testing `isExternal()` behavior with an e2e test using one external and one internal URL is a good idea as well.

Side note: What is the purpose of the `initJSDOM()` function in those tests? Maybe I'm missing something, but I believe those `expect` assertions are being run in Node and aren't associated with jsdom.

trusktr:
It's been a while since last commit. I will see about finding some time this week to thoroughly go over the PR. @jhildenbiddle If you have any un-pushed changes, please push those just in case.

What I will do during review is take at least one of my PRs and re-base it on top of the playwright branch and get it working.

anikethsaha:
cc @jhildenbiddle any update on this ?

jhildenbiddle:
@anikethsaha @trusktr 

All done. :) Feel free to kick the tires and let me know what you think. Note the updated details in the first comment as well.

I have noticed that this branch tends to get stuck on the "build (10.x)" task. This is odd, since it appears as though GitHub is not picking up the changes I made to our CI workflows: https://github.com/docsifyjs/docsify/tree/jest-playwright/.github/workflows. Not sure what's happening there. For now, you can check out the branch, update/install the necessary dependencies, and we can wrap up the review.

sy-records:
image

LGTM.

jhildenbiddle:
@anikethsaha --

> The whole setup and tests looks good to me.

I just flipped the PR from "draft" to "ready". Can you approve the PR if everything looks good?

sy-records (dismissed):

Koooooo-7 (approved):
@jhildenbiddle Awesome.

anikethsaha (dismissed):
Thanks

Issue #1282, opened 2020-07-13
option to put sidebar on the right


trusktr:
## Feature request

An option to put the sidebar on the right instead of the left

#### What problem does this feature solve?

Sometimes people want it on the other side (maybe they embed Docsify somewhere that already has a sidebar on the left, or maybe they have a custom landing page with a custom menu on the right and when they click the link to go to documentation they'd also want the docs to have the sidebar on the same side).

#### What does the proposed API look like?

```js
rightSidebar: true // or false
```

#### How should this be implemented in your opinion?

I think we might be able to achieve this just with CSS, leaving the markup alone.

#### Are you willing to work on this yourself?

Of course (if/when I get the time)!

anikethsaha:
+1 for this.

trusktr:
Turns out it was easy to do, only had to modify 3 numbers in the CSS:

![right-sidebar](https://user-images.githubusercontent.com/297678/87281700-56eace00-c4a8-11ea-80ad-17772fa213b3.gif)

I'll make a draft PR once I circle back (I'd like to finish the SSR tests in the other PR).

jhildenbiddle:
I think this is a feature best targeted for v5 and discussed in that context. 

There's more to consider than just making the menu appear on the right:

* How are other docsify themes affected?
* How are third-party themes affected?
* How are third-party plugins affected (i.e. do any rely on the sidebar-on-left behavior)?
* If the sidebar slides in from the right, should the sidebar toggle also be located on the right?
* How does this behavior look at desktop resolutions?

These are just off the top of my head. I'm sure there are more.

More importantly, not every preference needs to be a core docsify feature. Is this feature useful? Possibly, but given the fact that it has never before been requested in the nearly four years that docsify has existed, I think it's safe to say that this is not a high priority feature. I'd look to our existing issues list and feature requests to get a better idea of where we should be focused.

The good news is that folks that want a sidebar on the right can do so today by adding a custom `<style>` tag to their `index.html` file with the tweaked values you mentioned above. No need to wait for a new docsify option. We can target this issue for v5, see what kind of activity it gets, and determine if/how to implement this feature in v5 when the time comes.

paulhibbitts:
Hi @trusktr , I am really interested in having a right-side Sidebar as well - would you please be so kind to share the exact CSS items needing to be changed re: ` modify 3 numbers in the CSS`? If you could point me in the direction of changes for the menu icon that would be double-awesome🙂

Thanks so much!
Paul

[...]

PR #1288, opened 2020-07-16
[do not merge] add a rightSidebar option

trusktr:
**Summary**

Allows the sidebar to be on the right side. This is something developers might want (f.e. to match UX patterns on other parts of a site).

When sidebar is on the right the navbar and github corner badge are on the left on desktop. No changes to navbar or github corner on mobile (because on mobile they are hidden when the menu is open).

- [ ] This should not be merged until the options I added to index.html files are removed. I added the options so that you can test with vercel. After review, I will remove the options.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

- [x] Related tests have been updated (perhaps not enough tests, but we should settle on testing first, i.e. playwright PR)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -108,7 +108,7 @@ window.$docsify = {
 ## hideSidebar
 
 - Type : `Boolean`
-- Default: `true`
+- Default: `false`
 
 This option will completely hide your sidebar and wont render any content of the side even .
 
@@ -117,6 +117,13 @@ window.$docsify = {
   hideSidebar: true,
 };
 ```
+
+## rightSidebar
+
+- Type: `Boolean|String`
+- Default: `false`
+
+If true, the sidebar will be on the right side instead of the left.
 
 ## subMaxLevel
 
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -79,6 +79,15 @@ li input[type='checkbox']
   vertical-align middle
 
 /* navbar */
+body.right-sidebar .app-nav
+  margin 25px 0 0 60px
+  left 0
+  right unset
+
+  & > ul
+    padding-inline-start: 0;
+    padding-inline-end: 40px; // Chrome's default value.
+
 .app-nav
   margin 25px 60px 0 0
   position absolute
@@ -159,6 +168,10 @@ li input[type='checkbox']
       display block
 
 /* github corner */
+body.right-sidebar .github-corner
+  right unset
+  transform scaleX(-1)
+
 .github-corner
   border-bottom 0
   position fixed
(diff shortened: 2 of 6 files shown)


docs/configuration.md
```diff
@@ -108,7 +108,7 @@ window.$docsify = {
 ## hideSidebar
 
 - Type : `Boolean`
-- Default: `true`
+- Default: `false`
```
trusktr, line 111: This was wrong

trusktr:
@jhildenbiddle I already had this in the pipeline (needed it for my site), so made a PR. Reviewing playwright changes is my next task. Hoping for tomorrow.

src/themes/basic/_layout.styl
```diff
@@ -79,6 +79,15 @@ li input[type='checkbox']
...
 /* navbar */
+body.right-sidebar .app-nav
+  margin 25px 0 0 60px
+  left 0
+  right unset
+
+  & > ul
+    padding-inline-start: 0;
+    padding-inline-end: 40px; // Chrome's default value.
+
```
trusktr, line 90: @jhildenbiddle I aimed to make all the changes in here non-breaking for the end user.

You may have some opinion on how better to organize this change. Happy to hear if you do!

trusktr: The reason for `padding-inline-end: 40px;` thing is to flip it to the other side when the menu is on the left, otherwise it has a default style in Chrome's user-agent style of `padding-inline-start: 40px;`.

Maybe `padding-inline-start: 40px should be hard-coded for non`right-sidebar` mode, otherwise that one can vary from browser to browser (I don't know if they settled on the same values, haven't checked).

jhildenbiddle:
Is the intention to target this for v4, or as a placeholder for v5? If v4, then my questions / concerns outlined in the original issue (#1282) remain:

> I think this is a feature best targeted for v5 and discussed in that context.
> 
> There's more to consider than just making the menu appear on the right:
> 
> - How are other docsify themes affected?
> - How are third-party themes affected?
> - How are third-party plugins affected (i.e. do any rely on the sidebar-on-left behavior)?
> - If the sidebar slides in from the right, should the sidebar toggle also be located on the right?
> - How does this behavior look at desktop resolutions?
>
>These are just off the top of my head. I'm sure there are more.
> 
> More importantly, not every preference needs to be a core docsify feature. Is this feature useful? Possibly, but given the fact that it has never before been requested in the nearly four years that docsify has existed, I think it's safe to say that this is not a high priority feature. I'd look to our existing issues list and feature requests to get a better idea of where we should be focused.
>
> The good news is that folks that want a sidebar on the right can do so today by adding a custom <style> tag to their index.html file with the tweaked values you mentioned above. No need to wait for a new docsify option. We can target this issue for v5, see what kind of activity it gets, and determine if/how to implement this feature in v5 when the time comes.

PR #1291, opened 2020-07-17
chore: update issue template for feature proposal

anikethsaha:
New template for the feature proposal as discussed in discord.

> [Rendered Template](https://github.com/docsifyjs/docsify/blob/issue-template-for-feature-request/.github/ISSUE_TEMPLATE/feature_request.md)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,25 +1,97 @@
----
-name: Feature request
-about: Suggest an idea for this project
+<!-- This is taken from https://github.com/eslint/rfcs -->
 
----
-<!-- Please don't delete this template or we'll close your issue -->
-<!-- Please use English language -->
-<!-- Before creating an issue please make sure you are using the latest version of Docsify. -->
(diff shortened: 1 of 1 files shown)


sy-records:
Maybe we could add a new template instead of modifying this one.

trusktr:
In general most people are too lazy to file issues on libs to begin with. So the simpler the better? What if it were simpler for feature requests, and we get a new template for RFCs?

anikethsaha:
> If it is only for contributors I think it is really good, which has enough details to assess feasibility .
> but I wonder if it is for all users who want request a feature. if so, it seems more complicated.
> 
> As product users, I guess they couldn't fill in the template very well , especially the analysis of the change influences.
> so, I think the _Drawbacks_ and _BCA_ could be optional but recommend users to think about it.

This is for the contributors or someone who is willing to contribute or wants a new feature. 
Yea I think it would be good to make the BCA optional, not sure about drawbacks. 

> Maybe we could add a new template instead of modifying this one.


> In general most people are too lazy to file issues on libs to begin with. So the simpler the better? What if it were simpler for feature requests, and we get a new template for RFCs?

The only thing concerning me is that how to differentiate RFC and a feature request?
I think they are the same if they are related to the core part.

Issue #704, opened 2018-11-19
docsify-server-renderer error


ccforeverd:
My code here:

``` js
// docs-renderer.js
const fs = require('fs')
const path = require('path')
const docsifyRenderer = require('docsify-server-renderer')
const templatePath = path.resolve(__dirname, '../../docs/index.template.html')

const docsifyConfig = {
  basePath: '/src',
  // ...
}

const renderer = new docsifyRenderer({
  template: fs.readFileSync(templatePath, 'utf-8'),
  config: docsifyConfig
})

async function test () {
  renderer.renderToString('docs/')
    .then(console.log)
    .catch(console.log)
}

test()

module.exports = renderer
```

Then:

``` console
TypeError: Cannot read property 'indexOf' of undefined
    at AbstractHistory.parse (/Users/zhangshuyao/Sites/git/docs/node_modules/docsify-server-renderer/build.js:349:30)
    at Compiler.compile (/Users/zhangshuyao/Sites/git/docs/node_modules/docsify-server-renderer/build.js:553:37)
    at Renderer.<anonymous> (/Users/zhangshuyao/Sites/git/docs/node_modules/docsify-server-renderer/build.js:1126:28)
    at Generator.next (<anonymous>)
    at c (/Users/zhangshuyao/Sites/git/docs/node_modules/docsify-server-renderer/build.js:14:99)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:744:11)
    at startup (internal/bootstrap/node.js:285:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3)
```

There is shotcut of `docsify-server-renderer/build.js`

``` js
// line 553
  var curFileName = this$1.router.parse().file;
// line 346-360
  AbstractHistory.prototype.parse = function parse (path$$1) {
    var query = '';

    var queryIndex = path$$1.indexOf('?');
    if (queryIndex >= 0) {
      query = path$$1.slice(queryIndex + 1);
      path$$1 = path$$1.slice(0, queryIndex);
    }

    return {
      path: path$$1,
      file: this.getFile(path$$1),
      query: parseQuery(query)
    }
  };
```

I think maybe there is something wrong with my code or `build.js` after rollup-build, pls help me to check it.

Best

ccforeverd:
My work dir:
--src
  --docs
    --...
    --index.template.html
  --server
    --controller
      --docs-renderer.js
[...]

PR #1299, opened 2020-07-21
Fix #1101, #704

### LINKED ISSUE #1101: Server Side Rendering appears to no longer work
#### Steps to reproduce
Follow the instructions on Server Side Rendering
https://docsify.js.org/#/ssr

Following the SSR instructions results in a blank page. 
It looks like the app and config are not being injected by the server.

Using the package.json configuration, I can get `docsify start` to run. 
It looks like there is also a bug in docsify-cli when using the config file `-c` option .

The content should be rendered into the appropriate areas in the HTML template marked by comments.

```
  <!--inject-app-->
  <!--inject-config-->    
```

- Your OS: OSX
- Node.js version: v12.13.1
- npm/yarn version: 6.12.1
- Browser version: Latest Chrome and Firefox
- Docsify version: docsify@4.11.3, docsify-cli@4.4.0
- Docsify plugins: None, default install

boxsnake:
**Summary**
When updated, SSR won't work for local docs, throwing an error saying `Cannot call 'indexOf' of undefined`.
This fixed the problem by separate file content fetching. When the path is absolute, rather than fetch the content directly, the manner is separated:
- If the path is an external url, use fetch from `node-fetch`
- If the path is a local file with absolute path, just use `readFileSync` 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/packages/docsify-server-renderer/index.js
+++ b/packages/docsify-server-renderer/index.js
@@ -178,12 +178,16 @@ export default class Renderer {
     let content;
     try {
       if (isAbsolutePath(filePath)) {
-        const res = await fetch(filePath);
-        if (!res.ok) {
-          throw Error();
-        }
+        if (isExternal(filePath)) {
+          const res = await fetch(filePath);
+          if (!res.ok) {
+            throw Error();
+          }
 
-        content = await res.text();
+          content = await res.text();
+        } else {
+          content = await readFileSync(filePath, 'utf8');
+        }
         this.lock = 0;
       } else {
         content = await readFileSync(filePath, 'utf8');
--- a/src/core/router/history/abstract.js
+++ b/src/core/router/history/abstract.js
@@ -7,7 +7,7 @@ export class AbstractHistory extends History {
     this.mode = 'abstract';
   }
 
-  parse(path) {
+  parse(path = '') {
     let query = '';
 
     const queryIndex = path.indexOf('?');


boxsnake:
@anikethsaha I thought I made the wrong way. `isExternal` is not accessible there, thus the file won't be loaded properly.

boxsnake:
> LGTM.
> 
> It does make sense to use `readFileSync` for non external.
> 
> cc @trusktr thoughts ?

@anikethsaha 
Sorry, but I think my work is not finished.
There are many bugs related to the failure for local files.
- The default homepage is not `README.md`
- The file resolve is not from `basePath` config
- The `isExternal` function has some bugs (In Node.js, which work for SSR, there is no `location` object, but the code uses `location` in if clauses)

sy-records:
I also found the problem as you said. Also switching routes will 404 and refreshing will only work.

boxsnake:
> I also found the problem as you said. Also switching routes will 404 and refreshing will only work.

I am still working on fixing it, and will make a fresh PR later.

sy-records:
cool, I suggest you switch a branch to resubmit pr and not make changes directly on `develop`.

boxsnake:
> cool, I suggest you switch a branch to resubmit pr and not make changes directly on `develop`.

I'll make a new PR referring to this one.

PR #1307, opened 2020-07-24
FEAT: add pathNamespaces option to search plugin

### LINKED ISSUE #1303: Proposal: dynamic search namespace

Make it possible to dynamically change search namespace to isolate different parts of the documentation.
For example, we can isolate different locales from each other while using aliases. Right now, search index is built for all pages, and hence some queries return results for multiple locales at once (try searching for `index.html` at https://docsify.js.org/).

I'm currently using a patched version of the search plugin which updates `CONFIG.namespace` in `afterEach` callback by using the value from `vm.config.search.namespace`: https://github.com/test-prof/docs/blob/cb7826d37b37a338df64830fa85f6132f82992a6/docs/assets/docsify-search.js#L325

Thus, in order to change the namespace you need to update `vm.config.search.namespace` in your plugin.

That works fine for me but it's not an ideal solution.

Probably, exposing an API to switch the namespace would be better, e.g.: `vm.config.search.switchNamespace(...)` or even `vm.plugins.search.switchNamespace(...)`.

Long term: provide and API to access plugin instances, and thus making it possible to communicate with plugins from other plugins.

Short term: update namespace from the current `vm.config` settings in the `init` function: https://github.com/docsifyjs/docsify/blob/b8690199006366e86084e9e018def7b9b8f46512/src/plugins/search/search.js#L196

Sure.

palkan:
**Summary**

This option allows to dynamically choose the index path depending on the path prefixes in auto mode. Thus, different path namespace could avoid index intersection (e.g., when having multiple locales):

```js
// complete configuration parameters
    search: {
      // ...
      // Use different indexes for path prefixes (namespaces).
      // NOTE: Only works in 'auto' mode.
      //
      // When initialiazing an index, we look for the first path from the sidebar.
      // If it matches the prefix from the list, we switch to the corresponding index.
      pathNamespaces: ['/zh-cn', '/ru-ru', '/ru-ru/v1'],

      // You can provide a regexp to match prefixes. In this case,
      // the matching substring will be used to identify the index
      pathNamespaces: /^(\/(zh-cn|ru-ru))?(\/(v1|v2))?/
    }
```

Closes #1303.

Array version is used here: [docs.anycable.io](https://github.com/anycable/docs.anycable.io/blob/c723d9941469adbc81fbeae247589f65ba45526c/docs/index.html#L69) ([LIVE](https://docs.anycable.io)).

RegExp version is used here: [test-prof/docs](https://github.com/test-prof/docs/blob/2391e0e6ac206eb4c2138ee2f83be8daade54c43/docs/index.html#L56) ([LIVE](https://test-prof.evilmartians.com)).

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -44,6 +44,17 @@ By default, the hyperlink on the current page is recognized and the content is s
       // To avoid search index collision
       // between multiple websites under the same domain
       namespace: 'website-1',
+
+      // Use different indexes for path prefixes (namespaces).
+      // NOTE: Only works in 'auto' mode.
+      //
+      // When initialiazing an index, we look for the first path from the sidebar.
+      // If it matches the prefix from the list, we switch to the corresponding index.
+      pathNamespaces: ['/zh-cn', '/ru-ru', '/ru-ru/v1'],
+
+      // You can provide a regexp to match prefixes. In this case,
+      // the matching substring will be used to identify the index
+      pathNamespaces: /^(\/(zh-cn|ru-ru))?(\/(v1|v2))?/
     }
   }
 </script>
--- a/index.html
+++ b/index.html
@@ -51,7 +51,8 @@
           '/de-de/': 'Suche',
           '/zh-cn/': '搜索',
           '/': 'Search'
-        }
+        },
+        pathNamespaces: ['/zh-cn', '/de-de', '/ru-ru', '/es']
       },
       plugins: [
         function (hook, vm) {
--- a/src/plugins/search/index.js
+++ b/src/plugins/search/index.js
@@ -10,6 +10,7 @@ const CONFIG = {
   maxAge: 86400000, // 1 day
   hideOtherSidebarContent: false,
   namespace: undefined,
+  pathNamespaces: undefined,
 };
 
 const install = function(hook, vm) {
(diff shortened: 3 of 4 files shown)


docs/plugins.md
```diff
@@ -44,6 +44,17 @@ By default, the hyperlink on the current page is recognized and the content is s
       // To avoid search index collision
       // between multiple websites under the same domain
       namespace: 'website-1',
+
+      // Use different indexes for path prefixes (namespaces).
+      // NOTE: Only works in 'auto' mode.
+      //
+      // When initialiazing an index, we look for the first path from the sidebar.
+      // If it matches the prefix from the list, we switch to the corresponding index.
+      pathNamespaces: ['/zh-cn', '/ru-ru', '/ru-ru/v1'],
```
sy-records, line 53: Please modify the [index.html](https://github.com/docsifyjs/docsify/blob/develop/index.html) configuration for testing.

palkan: Done

anikethsaha:
> @anikethsaha yup, @sy-records and I checked the simple usage test including the customized storage name, it works fine. we could consider about giving more magic about the isolated namespace in the future

THats great. Yeah, this looks great

PR #1372, opened 2020-09-12
fix: Cannot serve off `/.../index.html`

### LINKED ISSUE #427: Fix bug where URI path is to a file and not to directory
When trying to access the website page via `domain.example/index.html` instead of `domain.example`, docsify is trying to read the markdown files relative to `index.html` as he thinks this is a directory, which result in HTTP 404 Error.
```
docsify.min.js:1 GET localhost/index.html/README.md 404 (Not Found)
```
Please fix and thank you for your awesome documentation generator.

rgladwell:
**Summary**

fixes #427

Docsify must be hosted on a server that supports a default directory index (i.e. maps `/.../` -> `/.../index.html`).

Some platforms do not support this, however. For example, HTML apps hosted on the popular game/software platform, Itch.io.

This change supports hosting Docsify off an explicit path file, such as `/index.html`. It does this by:

 1. Preventing the `/index.html` part of the path being removed during routing, and
 2. Ensuring we can retrieve the source markdown using explicit path file URIs.

For example, `http://example.org/index.html#/blah` would be mapped to the markdown source file `http://example.org/blah.md`.

This fixes:

https://github.com/docsifyjs/docsify/issues/427

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


rgladwell:
The e2e tests are reported by CI as failing, but are passing locally for me.

anikethsaha (changes requested):
Thanks a lot for fixing the bug. 

It does seems to work and fixes the issue. 

Few questions

- It would be great to have some more test cases. 
- it would be great to have some comments for why we are doing those normalizing and those conditional check.  May be linking to the issues/PR will work as well. 


Also, I doubt this might be a semver-major as it can be a breaking change. cc @docsifyjs/reviewers

rgladwell:
@anikethsaha Thanks for the feedback. Could you clarify what sort of test cases you are looking for? I notice you have Cypress and unit tests. Which would you prefer, and where would they go exactly?

sy-records:
It looks good.

https://docsify-previe-git-fork-rgladwell-fix-base-path-file-not-70a56c.docsify-core.vercel.app/index.html#/quickstart

https://docsify.js.org/#/quickstart

anikethsaha:
> @anikethsaha Thanks for the feedback. Could you clarify what sort of test cases you are looking for? I notice you have Cypress and unit tests. Which would you prefer, and where would they go exactly?

You could add cypress tests for other links like 

`/index.html/`
`/#/index.html/`
`/index.html/#/` (it would be 404 but still good to cover. )

Also if there are other cases you think you can add

rgladwell:
> @anikethsaha  You could add cypress tests for other links like

Done, let me know if they're sufficient.

anikethsaha (commented):
LGTM 🎉 

My only concern is whether this should be considered as breaking or not!

rgladwell:
It shouldn't be breaking, it's an additive change. You could argue that any breakages would be bugs.

Koooooo-7 (commented):
@anikethsaha  I think when it is not sure if it is a breaking change, we can make it as an option. 
Giving the `switch button` to users' hands.:)

rgladwell:
Hmmm I think this change blocks live reloads in development mode.

anikethsaha (approved):

anikethsaha:
I agree that it is not a breaking change as it was a bug so probably it wasn't used before in order to avoid the bug.

rgladwell:
Been trying to add a feature flag for this, but the code is getting very complicated. If this isn't a breaking change, can we live without it?

anikethsaha:
cc @docsifyjs/core

jhildenbiddle:
Hey @rgladwell,

Thanks for the PR! 

We recently merged #1276 into `develop` which is an overhaul of our CI and local test environments. As a result, the e2e tests in this PR will need to rewritten in Jest+Playwright instead of Cypress. The good news is that the test environments should be more stable and easier to work with. The PR contains a summary of the changes, help is available in `/test/README.md` directory, and I'm available if you have any questions.

rgladwell:
Is there some way to get a screenshot or run non-headless for the Jest e2e tests?

jhildenbiddle:
> Is there some way to get a screenshot or run non-headless for the Jest e2e tests?

Apologies for the slow response, @rgladwell. Doing some PR cleanup and stumbled on this.

Configuring which browsers are used for testing and toggling headless mode is done here:

https://github.com/docsifyjs/docsify/blob/c7f4c7cf9b41e86aa2c21715ac18283a2d443721/jest.config.js#L44-L57

For example, you could temporarily modify the `jest.config.js` file to only test locally with Chromium in non-headless mode during development:

```js
testEnvironmentOptions: {
  'jest-playwright': {
    // prettier-ignore
    browsers: [
      'chromium',
      // 'firefox',
      // 'webkit',
    ],
    launchOptions: {
      headless: false,
      // devtools: true,
    },
  },
},
```

You may also want to review the `example.test.js` file in the `/test/e2e/`, `/test/intrgration/`, and `/test/unit/` directories and how the `docsifyInit()` helper is used to spin up dynamic docisify sites for testing.

rgladwell:
@jhildenbiddle @anikethsaha @Koooooo-7  As requested, added e2e tests using Jest framework.

src/core/router/history/hash.js
```diff
@@ -18,7 +18,15 @@ export class HashHistory extends History {
     const path = window.location.pathname || '';
     const base = this.config.basePath;
 
-    return /^(\/|https?:)/g.test(base) ? base : cleanPath(path + '/' + base);
+    // This handles the case where Docsify is served off an
+    // explicit file path, i.e.`/base/index.html#/blah`. This
+    // prevents the `/index.html` part of the URI from being
+    // remove during routing.
+    // See here: https://github.com/docsifyjs/docsify/pull/1372
+    const basePath = path.endsWith('.html')
```
jhildenbiddle, line 26: [`String.prototype.endsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#browser_compatibility) is not supported by IE11 (which Docsify v4 supports). We'll need to use a legacy-compatible method here.

jhildenbiddle (changes requested):

src/core/router/history/hash.js
rgladwell: Fixed in ac5ed770ff25ee7ebb52bf893d1269b30580549b.

sy-records (approved):
LGTM

Koooooo-7 (approved):

jhildenbiddle (approved):

rgladwell:
Thanks!

Issue #1202, opened 2020-06-03
scroll problem


playgithub:
To reproduce the problem:

1. Open https://xmake.io/#/getting_started
2. Locate to "Getting Started / Project Examples", and make sure it's expanded
3. Locate to the next section, i.e. "Getting Started / Configuration"
4. It doen't scroll properly

trusktr:
Not sure I understand.

> 1. Open https://xmake.io/#/getting_started

I go there, and I'm at the top of the page.

> 2. Locate to "Getting Started / Project Examples", and make sure it's expanded

When I click on that, it takes me to that page, at the top, as expected.

> 3. Locate to the next section, i.e. "Getting Started / Configuration"

Then when I click on that, it goes to that page, and I am at the top, as expected.

> 4. It doen't scroll properly

What are you expecting here?

playgithub:
* unexpected
![1](https://user-images.githubusercontent.com/3665590/87898070-ca2fa980-ca7f-11ea-9342-a5a0ee934d92.PNG)
![2](https://user-images.githubusercontent.com/3665590/87898078-ce5bc700-ca7f-11ea-95cf-5bccaa2a9d17.PNG)
![3](https://user-images.githubusercontent.com/3665590/87898083-d1ef4e00-ca7f-11ea-82ba-5ab11e3a1354.PNG)

* expected
![should_be](https://user-images.githubusercontent.com/3665590/87898120-e59ab480-ca7f-11ea-8063-b1e454fe7fda.PNG)

trusktr:
Ah, got it. Your expecting the _sidebar_ to scroll. I don't think this has been implemented (at least not during the time I've been working with the project).

This would be great to have. It's a problem that is exacerbated when the sidebar has long sub-lists. With smaller amounts of items on each page, then the problem isn't so noticeable because when the menu expands usually all items can fit on screen.

Two possible workarounds for now:

- split the page into smaller pages, so that each expander opens to show less amount of items.
- perhaps with a Docsify plugin it is possible to scroll the sidebar after a page change, by detecting which item is active.

trusktr:
As for an official solution, we can modify the code at the same location where the current code is for scrolling the main content. We can make this an option so it is a semver-minor non-breaking change, then in v5 we can change the default to enabled.

playgithub:
thanks

Anilople:
Same problem.

[...]

PR #1384, opened 2020-10-02
Fix scroll event end value

### LINKED ISSUE #1383: Fix rounding error in scroll behavior
## Bug Report

Scroll implementation does not account for floating point values returned from `getBoundingClientRect()`:

https://github.com/docsifyjs/docsify/blob/c2d12ed27fe86b0c5cd690b4d8a22bf06d2a90b9/src/core/event/scroll.js#L21

The result is that scroll event end values will be 1px less than expected because some browsers ignore the value after the decimal point (e.g. `1234.567` is process as `1234` instead of being rounded up to `1235`). Values returned from `getBoundingClientRect()` should be wrapped in `Math.round()` to ensure consistent behavior.

#### Steps to reproduce

I noticed this bug on the docsify-themeable website while testing the [docsify-sidebar-collapse](https://github.com/iPeng6/docsify-sidebar-collapse) plugin for a separate issue (I'm not adding this plugin to the docsify-themeable website). I noticed that when some headings were scrolled to the top of the screen they were 1px away from the top edge.

#### Mention the docsify version in which this bug was not present (if any)

All versions up to current (4.11.6)

jhildenbiddle:
Fixes #1383

**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] IE


sy-records:
Does this pr solve that problem as well? #1202

anikethsaha (approved):

Koooooo-7 (approved):

PR #1409, opened 2020-10-26
feat: Vue components, mount options, global options, and v3 support

### LINKED ISSUE #752: consider support vue data through config globally
maybe we can change this line of code:
https://github.com/docsifyjs/docsify/blob/1a945d445a636d778b242f9d74af4f72c7d89e16/src/core/render/index.js#L60
to include data from config file, code may like this:
```js
      window.__EXECUTE_RESULT__ = new window.Vue({ data: this.config.vueData }).$mount('#main')
```

jhildenbiddle:
Fix #752 #1252  

**Summary**

This PR significantly improves Vue functionality in docsify and addresses a few bugs/issues with the test environment.

Note that the preview site auto-generated for this PR is not built with the correct `index.html` so it will not process Vue content as expected. To preview locally, check out the branch/PR and run the following:

```bash
node ./test/config/server.js --start
```

**Public Changes**

* Add `vueComponents` configuration option
* Add `vueMounts` configuration option
* Add `vueGlobalOptions` configuration option
* Add Vue 3 support
* Add styling for `<output>` element to `vue.css`
* Update `/docs/` site

**Dev Changes**

There was a significant test-related bug fixed in this PR. Due to the way Jest manages the JSDOM environment, global objects like the `window` and `document` objects are not reset after each test in the same file. As a result, event listeners added to the `window` object by docsify were not being cleared after each test, resulting in event listeners from previous previous tests continuing to be triggered.

* Add support to `docsify-init()` for functions in `config`
* ~~Add `.markdownlint.json`~~
* Add `_logHTML.selector` option to `docsify-init()`
* Remove unnecessary global variables from test environment
* Remove `--runInBand` flag from local Jest tests
* Update Vue tests
* Update Jest cleanup/reset in `/test/config/jest.setup-tests.js`
* Fix `waitForSelector()` timeout in `/test/helpers/wait-for.js`
* Fix `docsify-init()` handling of `config.basePath` option

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [x] Docs
- [x] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

_Screenshot of new `<output>` styling (below code block):_

Screen Shot 2020-10-26 at 2 42 11 AM

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

- [x] Related tests have been updated

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,4 @@
+{
+  "MD013": false,
+  "MD033": false
+}
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,4 +1,0 @@
-{
-  "MD013": false,
-  "MD033": false
-}
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -634,7 +634,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Registers Vue components using the component name as the key with an object containing Vue options as the value.
+Registers Vue components using the component name as the key with an object containing Vue options as the value. Component data is not shared and changes will not persist as users navigate the site.
 
 ```js
 window.$docsify = {
@@ -701,7 +701,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies Vue mount elements and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value.
+Specifies Vue mount elements and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area (`#main, .markdown-section`) each time a new page is loaded.
 
 ```js
 window.$docsify = {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -494,15 +494,15 @@ window.$docsify = {
 ```
 
 ## crossOriginLinks
-- type: `Array`  
 
-When `routerMode: 'history'`, you may face the cross-origin issues, See [#1379](https://github.com/docsifyjs/docsify/issues/1379).  
-In Markdown content, there is a simple way to solve it, see extends Markdown syntax `Cross-Origin link` in [helpers](helpers.md).  
+- type: `Array`
+
+When `routerMode: 'history'`, you may face the cross-origin issues, See [#1379](https://github.com/docsifyjs/docsify/issues/1379).
+In Markdown content, there is a simple way to solve it, see extends Markdown syntax `Cross-Origin link` in [helpers](helpers.md).
+
 ```js
 window.$docsify = {
-  crossOriginLinks:[
-      "https://example.com/cross-origin-link",
-   ],
+  crossOriginLinks: ['https://example.com/cross-origin-link'],
 };
 ```
 
@@ -629,3 +629,104 @@ window.$docsify = {
   topMargin: 90, // default: 0
 };
 ```
+
+## vueComponents
+
+- type: `Object`
+
+Registers Vue components using the component name as the key with an object containing Vue options as the value.
+
+```js
+window.$docsify = {
+  vueComponents: {
+    'button-counter': {
+      template: `
+        <button @click="count += 1">
+          You clicked me {{ count }} times
+        </button>
+      `,
+      data() {
+        return {
+          count: 0,
+        };
+      },
+    },
+  },
+};
+```
+
+```markdown
+<button-counter></button-counter>
+```
+
+<output data-lang="output">
+  <button-counter></button-counter>
+</output>
+
+## vueGlobalOptions
+
+- type: `Object`
+
+Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueOptions](#vueoptions), [vueComponents](#vuecomponents), or a markdown `<script>`.
+
+```js
+window.$docsify = {
+  vueGlobalOptions: {
+    data() {
+      return {
+        count: 0,
+      };
+    },
+  },
+};
+```
+
+```markdown
+<p>
+  <button v-on:click="count -= 1">-</button>
+  {{ count }}
+  <button v-on:click="count += 1">+</button>
+</p>
+```
+
+<output data-lang="output">
+  <p>
+    <button v-on:click="count -= 1">-</button>
+    {{ count }}
+    <button v-on:click="count += 1">+</button>
+  </p>
+</output>
+
+## vueOptions
+
+- type: `Object`
+
+Specifies Vue mount elements and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value.
+
+```js
+window.$docsify = {
+  vueOptions: {
+    '#counter': {
+      data() {
+        return {
+          count: 0,
+        };
+      },
+    },
+  },
+};
+```
+
+```markdown
+<div id="counter">
+  <button v-on:click="count -= 1">-</button>
+  {{ count }}
+  <button v-on:click="count += 1">+</button>
+</div>
+```
+
+<output id="counter">
+  <button v-on:click="count -= 1">-</button>
+  {{ count }}
+  <button v-on:click="count += 1">+</button>
+</output>
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -634,7 +634,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Registers Vue components using the component name as the key with an object containing Vue options as the value. Component `data` is unique for each instance and will not persist as users navigate the site.
+Creates and registers global [Vue components](https://vuejs.org/v2/guide/components.html). Components are specified using the component name as the key with an object containing Vue options as the value. Component `data` is unique for each instance and will not persist as users navigate the site.
 
 ```js
 window.$docsify = {
@@ -667,7 +667,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies options for use with Vue content not explicitly mounted with [vueMounts](#mounting-dom-elements), [vueComponents](#components), or a [markdown script](#markdown-script). Changes to global `data` will persist and be reflected anywhere global references are used.
+Specifies [Vue options](https://vuejs.org/v2/api/#Options-Data) for use with Vue content not explicitly mounted with [vueMounts](#mounting-dom-elements), [vueComponents](#components), or a [markdown script](#markdown-script). Changes to global `data` will persist and be reflected anywhere global references are used.
 
 ```js
 window.$docsify = {
@@ -701,7 +701,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies DOM elements to mount as Vue instances and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area each time a new page is loaded. Mount element `data` is unique for each instance and will not persist as users navigate the site.
+Specifies DOM elements to mount as [Vue instances](https://vuejs.org/v2/guide/instance.html) and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area each time a new page is loaded. Mount element `data` is unique for each instance and will not persist as users navigate the site.
 
 ```js
 window.$docsify = {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -634,7 +634,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Registers Vue components using the component name as the key with an object containing Vue options as the value. Component data is not shared and changes will not persist as users navigate the site.
+Registers Vue components using the component name as the key with an object containing Vue options as the value. Component `data` is unique for each instance and will not persist as users navigate the site.
 
 ```js
 window.$docsify = {
@@ -667,7 +667,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueMounts](#vuemounts), [vueComponents](#vuecomponents), or a markdown `<script>`.
+Specifies options for use with Vue content not explicitly mounted with [vueMounts](#mounting-dom-elements), [vueComponents](#components), or a [markdown script](#markdown-script). Changes to global `data` will persist and be reflected anywhere global references are used.
 
 ```js
 window.$docsify = {
@@ -701,7 +701,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies DOM elements to mount as Vue instances and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area each time a new page is loaded.
+Specifies DOM elements to mount as Vue instances and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area each time a new page is loaded. Mount element `data` is unique for each instance and will not persist as users navigate the site.
 
 ```js
 window.$docsify = {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -683,17 +683,17 @@ window.$docsify = {
 
 ```markdown
 <p>
-  <button v-on:click="count -= 1">-</button>
+  <button @click="count -= 1">-</button>
   {{ count }}
-  <button v-on:click="count += 1">+</button>
+  <button @click="count += 1">+</button>
 </p>
 ```
 
 <output data-lang="output">
   <p>
-    <button v-on:click="count -= 1">-</button>
+    <button @click="count -= 1">-</button>
     {{ count }}
-    <button v-on:click="count += 1">+</button>
+    <button @click="count += 1">+</button>
   </p>
 </output>
 
@@ -719,14 +719,14 @@ window.$docsify = {
 
 ```markdown
 <div id="counter">
-  <button v-on:click="count -= 1">-</button>
+  <button @click="count -= 1">-</button>
   {{ count }}
-  <button v-on:click="count += 1">+</button>
+  <button @click="count += 1">+</button>
 </div>
 ```
 
 <output id="counter">
-  <button v-on:click="count -= 1">-</button>
+  <button @click="count -= 1">-</button>
   {{ count }}
-  <button v-on:click="count += 1">+</button>
+  <button @click="count += 1">+</button>
 </output>
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -245,10 +245,10 @@ window.$docsify = {
   // Custom file name
   coverpage: 'cover.md',
 
-  // mutiple covers
+  // multiple covers
   coverpage: ['/', '/zh-cn/'],
 
-  // mutiple covers and custom file name
+  // multiple covers and custom file name
   coverpage: {
     '/': 'cover.md',
     '/zh-cn/': 'cover.md',
@@ -410,7 +410,7 @@ window.$docsify = {
 };
 ```
 
-?> If this options is `false` but you dont want to emojify some specific colons , [Refer this](https://github.com/docsifyjs/docsify/issues/742#issuecomment-586313143)
+?> If this options is `false` but you don't want to emojify some specific colons , [Refer this](https://github.com/docsifyjs/docsify/issues/742#issuecomment-586313143)
 
 ## mergeNavbar
 
@@ -604,7 +604,7 @@ window.$docsify = {
 };
 ```
 
-Load the right 404 page according to the localisation:
+Load the right 404 page according to the localization:
 
 ```js
 window.$docsify = {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -667,7 +667,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueOptions](#vueoptions), [vueComponents](#vuecomponents), or a markdown `<script>`.
+Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueMountOptions](#vuemountoptions), [vueComponents](#vuecomponents), or a markdown `<script>`.
 
 ```js
 window.$docsify = {
@@ -697,7 +697,7 @@ window.$docsify = {
   </p>
 </output>
 
-## vueOptions
+## vueMountOptions
 
 - type: `Object`
 
@@ -705,7 +705,7 @@ Specifies Vue mount elements and their associated options. Mount elements are sp
 
 ```js
 window.$docsify = {
-  vueOptions: {
+  vueMountOptions: {
     '#counter': {
       data() {
         return {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -667,7 +667,7 @@ window.$docsify = {
 
 - type: `Object`
 
-Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueMountOptions](#vuemountoptions), [vueComponents](#vuecomponents), or a markdown `<script>`.
+Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueMounts](#vuemounts), [vueComponents](#vuecomponents), or a markdown `<script>`.
 
 ```js
 window.$docsify = {
@@ -697,15 +697,15 @@ window.$docsify = {
   </p>
 </output>
 
-## vueMountOptions
+## vueMounts
 
 - type: `Object`
 
-Specifies Vue mount elements and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area (`#main, .markdown-section`) each time a new page is loaded.
+Specifies DOM elements to mount as Vue instances and their associated options. Mount elements are specified using a [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) as the key with an object containing Vue options as their value. Docsify will mount the first matching element in the main content area each time a new page is loaded.
 
 ```js
 window.$docsify = {
-  vueMountOptions: {
+  vueMounts: {
     '#counter': {
       data() {
         return {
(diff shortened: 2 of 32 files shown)


.markdownlint.json
```diff
@@ -0,0 +1,4 @@
+{
```
sy-records, line 1: What is the purpose of this file? Does it need to be submitted?

jhildenbiddle: It allows us to configure lint rules for markdown files. More information is available here: https://github.com/DavidAnson/markdownlint

trusktr: Prettier works on markdown too (including formatting code inside code blocks).

docs/configuration.md
```diff
@@ -629,3 +629,104 @@ window.$docsify = {
...
+    '#counter': {
+      data() {
+        return {
+          count: 0,
+        };
+      },
+    },
+  },
+};
+```
```
trusktr, line 718: I feel that there's explanation missing here as to what exactly this does. 

If the vueMountOptions contains 5 selectors, does that mean Vue instantiates 5 components and mounts to those five areas on the page?

docs/configuration.md
```diff
@@ -629,3 +629,104 @@ window.$docsify = {
...
+
+<output data-lang="output">
+  <button-counter></button-counter>
+</output>
+
+## vueGlobalOptions
+
+- type: `Object`
+
+Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via [vueMountOptions](#vuemountoptions), [vueComponents](#vuecomponents), or a markdown `<script>`.
```
trusktr, line 670: Do these options apply to any of the components made with `vueMountOptions`?

docs/configuration.md
```diff
@@ -629,3 +629,104 @@ window.$docsify = {
...
+      `,
+      data() {
+        return {
+          count: 0,
+        };
+      },
+    },
+  },
+};
+```
```
trusktr, line 656: Why would we want to use `vueMountOptions` with a selector, when we could specify components using `vueComponents`?

Trying to get a more clear picture.

docs/index.html
```diff
@@ -101,6 +101,62 @@
...
+            </button>
+          `,
+            data() {
+              return {
+                count: 0,
+              };
+            },
+          },
+        },
+        vueGlobalOptions: {
```
trusktr, line 118: I'm guessing these `vueGlobalOptions` get merged with each component in `vueComponents`.

docs/vue.md
```diff
@@ -20,118 +20,314 @@ To get started, load either the production or development version of Vue in your
...
-  <li v-for="i in 3">{{ i }}</li>
+  <li v-for="i in 3">Item {{ i }}</li>
 </ul>
+
+<!-- JavaScript expressions -->
+<p>2 + 2 = {{ 2 + 2 }}</p>
 ```
 
-The HTML above will render the following:
+<output data-lang="output">
```
trusktr, line 44: Why are there new `<output>` elements in the DOM?

docs/vue.md
```diff
@@ -20,118 +20,314 @@ To get started, load either the production or development version of Vue in your
...
+  },
+}
 ```
 
-#### Vue 3.x
+```markdown
+Good {{ timeOfDay }}!
+```
+
+<output data-lang="output">
```
trusktr, line 120: What does `data-lang="output"` do?

trusktr: Oooh, nevermind, I see: the `<output>` element is not the result that the user will see in their DOM, but rather it is used here in the docs to show what the result will _be like_.

docs/vue.md
```diff
@@ -20,118 +20,314 @@ To get started, load either the production or development version of Vue in your
...
+## Technical Notes
+
+- Docsify processes Vue content in the following order:
+  1. Markdown `<script>`
+  1. `vueMountOptions`
+  1. `vueGlobalOptions`
+- Docsify will not mount an existing Vue instance or an element that contains an existing Vue instance.
+- Docsify will automatically destroy/unmount all Vue instances it creates before new page content is loaded.
+- When processing `vueGlobalOptions`, docsify parses each root element of the main content area and mounts the element if Vue content is detected. Docsify does not parse each individual node within the main content area.
+- When processing `vueGlobalOptions`, docsify will only detect the full `v-` attribute syntax (e.g `v-bind:href` or `v-on:click`). For performance reasons, detection of Vue's [shorthand](https://vuejs.org/v2/guide/syntax.html#Shorthands) attribute syntax (e.g. `:href`or `@click`) is not supported.
```
trusktr, line 333: What performance reasons are these? I always use shorthand for everything in Vue and have been fine in all my cases so far (performance wise).

.markdownlint.json
trusktr: Another question is, what if we just use Prettier for markdown instead of markdownlint? Or can they complement each other?

jhildenbiddle: @trusktr Prettier and markdownlint serve different purposes but complement each other: 

* Prettier is designed to format code. By doing so, it enforces standards.
* markdownlint is designed to enforce standards. It can automatically fix some (but not all) errors which ends up formatting code.

Here's a simple example:

```markdown
## Title 1


- Item
- Item


#### Title 2


* Item
* Item
```

Prettier will remove the extra line breaks between content:

```markdown
## Title

- Item
- Item

#### Title 2

* Item
* Item
```

Markdownlint will provide the following error message:

* Document should start with an `h1` heading
* Heading levels should only increment/decrement one level at a time
* List style types should be consistent (don't mix `-` and `*` in the same document)

Regardless, I'm going to remove `.markdownlint.json` from the PR and handle markdown linting in a separate PR. :)

docs/configuration.md
jhildenbiddle: @trusktr Agreed. Good catch. This question is answered on the `vue.md` page, but I neglected to add it to the description here as well.

> Docsify will mount the first matching element in the main content area (`#main, .markdown-section`) each time a new page is loaded.

For example:

```js
window.$docsify = {
  vueMountOptions: {
    // The CSS selector "p" could match multiple elements.
    // However, only the first <p> element will be mounted.
    'p': {  
      data() {
        return {
          count: 0,
        };
      },
    },
  },
};
```

docs/configuration.md
jhildenbiddle: @trusktr --

No. That's what the following part of the description is trying to make clear: 

> These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via vueMountOptions, vueComponents, or a markdown `<script>`.

If that isn't clear, let me know and I can try to rework the description.

docs/configuration.md
jhildenbiddle: Fair question.

Here are the first few that come to mind:

1. Vue supports mounting both DOM elements and components. So should we. All of the reasons a Vue user might opt to mount a DOM element instead of creating a Vue component apply here.

1. Mounting DOM elements is the first concept found in the [Vue docs](https://vuejs.org/v2/guide/) and introductory Vue tutorials. It is the fastest and simplest way to be productive with a library like Vue and requires the least amount of expertise. Allowing code from docs and tutorials to be easily ported to docsify is an important consideration for our users--especially junior and non-developer users.

1. Components are typically presented as reusable building blocks that combine to make larger applications. Obviously this doesn't have to be the case, but it's how they are often conceptualized. Forcing users to think in "component" terms can be confusing when they are just trying to do something simple task like render data on a docsify page.

1. There are technical and UX differences between mounted DOM elements and components. For example, `vueMountOptions` allows mounting content found directly in your markdown file:

   ```markdown
   # Title

   <p id="my-content">Here is some Vue content: {{ message }}</p>
   ```
   ```js
   vueMountOptions: {
     '#my-content': {
       data() {
         return {
           message: 'Hello, World!'
         };
       }
     }
   }
   ```

   Vue components can do this as well, but doing so requires adding an `inline-template` attribute to the component tag which is not intuitive. 

   ```markdown
   # Title

   <my-component inline-template>
     <p>Here is some Vue content: {{ message }}</p>
   </my-component>
   ```

   More typically, components specify their markup using the `template` property:

   ```markdown
   # Title

   <my-component />
   ```
   ```js
   vueComponents: {
     'my-component': {
       template: `<p>Here is some Vue content: {{ message }}</p>`,
       data() {
         return {
           count: 0,
         };
       }
     }
   }
   ```

   This is not a big deal when building Vue apps outside of docsify, but with docsify it means `template` markup cannot contain markdown, [will not be processed by docsify plugins](https://github.com/jhildenbiddle/docsify-tabs/issues/31), will not have syntax highlighting or linting while authoring, will not render default content outside of docsify (e.g., GitHub), etc. You can work around some of these things, but doing requires a level of expertise that not all docsify users will have.

As for our and why you would choose one over the other, my intention was keep it light on our end and refer people to the Vue docs for more information. If you think we need to clarify why users should choose one over the other when using docsify, I can work on that.

docs/index.html
jhildenbiddle: Not shared. Here's the description:

> ## vueGlobalOptions
> 
> - type: `Object`
> 
> Specifies Vue options to be shared throughout your site. These options will be used when Docsify detects Vue content in the main content area that has not been previously mounted via `vueMountOptions`, `vueComponents`, or a markdown `<script>`.

docs/vue.md
jhildenbiddle: Nope. New styles applied to a previously unstyled HTML element: `<output>`

docs/vue.md
jhildenbiddle: Yep! 😄

docs/vue.md
jhildenbiddle: There are performance implications for docsify, not for Vue.

The issue is that I'm parsing content for Vue syntax to determine which elements to mount and pass `vueGlobalOptions` to. This is done using regular expressions:

https://github.com/docsifyjs/docsify/blob/a5a431f0c156119a6fd6fbea53b67e7eb745d550/src/core/render/index.js#L129-L131

The full syntax is the most efficient and reliable to parse (above). The shorthand syntax is less efficient and either less reliable or impossible to parse:

#### `v-bind` Shorthand

```html
<!-- full syntax -->
<a v-bind:href="url"> ... </a>

<!-- shorthand -->
<a :href="url"> ... </a>

<!-- shorthand with dynamic argument (2.6.0+) -->
<a :[key]="url"> ... </a>
```

#### `v-on` Shorthand

```html
<!-- full syntax -->
<a v-on:click="doSomething"> ... </a>

<!-- shorthand -->
<a @click="doSomething"> ... </a>

<!-- shorthand with dynamic argument (2.6.0+) -->
<a @[event]="doSomething"> ... </a>
```

I'd prefer to release something strict but efficient and reliable, then wait for feedback from users to determine if we need to support shorthand syntax as well. If we have to, we can, but doesn't feel like a requirement at this point for me.

docs/configuration.md
jhildenbiddle: @trusktr Update the description as follows:

> Specifies options for use with Vue content not explicitly mounted with `vueMounts`, `vueComponents`, or a markdown `<script>`. Changes to global `data` will persist and be reflected anywhere global references are used.

Hopefully that makes it more clear.

docs/index.html
jhildenbiddle: @trusktr Updated the description as follows:

> Specifies options for use with Vue content not explicitly mounted with `vueMounts`, `vueComponents`, or a markdown `<script>`. Changes to global `data` will persist and be reflected anywhere global references are used.

Hopefully that makes it more clear.

docs/vue.md
jhildenbiddle: Added support for Vue shorthand directive syntax in https://github.com/docsifyjs/docsify/pull/1409/commits/35ad5c559b43f6d70fae054b1812edef9d2b7e4d.

docs/configuration.md
trusktr: Thanks for that explanation! I know about the starting the root of a Vue tree and components used in the Vue tree, but these are two features additional to Vue (for Docsify).

Knowing Vue, I know I can put `new Vue` in a script that I can load with `<script src="...">` on every page.

I suppose what I'm really asking is the following.

Here are two ways to basically do the same thing with the new system. 

**Method 1:**

```html
<div id="my-content">

## {{ title }}
Here is some Vue content: {{ message }}

</div>
```

```js
vueMounts: {
  '#my-content': {
    data: () => ({
      title: 'My Title',
      message: 'Hello, World!'
    })
  }
}
```

**Method 2:**

```html
<my-content inline-template>

## {{ title }}
Here is some Vue content: {{ message }}

</my-content>
```

```js
vueComponents: {
  'my-content': {
    data: () => ({
      title: 'My Title',
      message: 'Hello, World!'
    })
  }
}
```

Both have about the same ease-of-use (learning to use `inline-template` is as easy as writing an id or class name). The question is, why have both?

Is the difference only the inline-template vs no inline-template thing? If so, is it worth having two options just for that small difference?

docs/configuration.md
trusktr: Why not have it apply to all matches? That would be convenient I imagine (one definition, but multiple instances of it across the page). Thinking `querySelectorAll` instead of `querySelector`.

Also people may wonder if the following examples will work:

**Exampe 1**

Does it render both?

```html
<p class="one">{count}</p>
<p class="two">{count}</p>
```

```js
window.$docsify = {
  vueMounts: {
    'p.one': {  
      data: () => ({ count: 1 },
    },
    'p.two': {  
      data: () => ({ count: 2 },
    },
  },
};
```

**Example 2**

What happens?

```html
<p class="one">
  {count}
  <p class="two">
    {count}
  </p>
</p>
```

```js
window.$docsify = {
  vueMounts: {
    'p.one': {  
      data: () => ({ count: 1 },
    },
    'p.two': {  
      data: () => ({ count: 2 },
    },
  },
};
```

**Example 3**

What happens?

```html
<component-one inline-template>
  {count}
  <component-two inline-template>
    {count}
  </component-two>
</component-one>
```

```js
window.$docsify = {
  vueComponents: {
    'component-one': {
      data: () => ({ count: 1 },
    },
    'component-two': {  
      data: () => ({ count: 2 },
    },
  },
};
```

**Example 4**

User opens feature request saying "can we render all of them?"

```html
<p>{count}</p>
<p>{count}</p>
<p>{count}</p>
<p>{count}</p>
<p>{count}</p>
```

```js
let pCount = 0

window.$docsify = {
  vueMounts: {
    'p': {  
      data: () => ({ count: ++pCount  },
    }
  },
};
```

**Exampe 5**

How do I make this work?

```html
<p class="one">{count}{foo}</p>
<p class="two">{count}{foo}</p>
```

```js
window.$docsify = {
  vueMounts: {
    'p.one': {
      data: () => ({ count: 1 },
    },
    'p.two': {
      data: () => ({ count: 2 },
    },
    'p': {
      data: () => ({ foo: 'bar' },
    },
  },
};
```

docs/index.html
trusktr: That is indeed a lot more clear than before.

The object looks like a Vue component definition. What if it is called `vueGlobalComponent` instead?

Also is there an example?

trusktr: Nvm, the `vue.md` file was not expanded by default here in the diff view, so searching for `vueGlobalOptions` didn't yield results but now it does.

How is it that we can change the DOM on page changes, and the data still works with each new DOM?

docs/vue.md
```diff
@@ -20,118 +20,318 @@ To get started, load either the production or development version of Vue in your
...
-  {{ counter }}
-  <button v-on:click="counter += 1">+</button>
-</div>
+## Markdown script
 
-<!-- prettier-ignore-end -->
+Vue content can mounted using a `<script>` tag in your markdown pages.
 
-!> Only the first `<script>` tag in a markdown file is executed. If you are working with multiple Vue components, all `Vue` instances must be created within this tag.
+!> Only the first `<script>` tag in a markdown file is executed. If you wish to mount multiple Vue instances using a script tag, all instances must be mounted within the first script tag in your markdown.
```
trusktr, line 307: > Only the first `<script>` tag in a markdown file is executed

This should be fixed, it is totally not intuitive. But let's save that for a separate PR.

docs/vue.md
```diff
@@ -20,118 +20,318 @@ To get started, load either the production or development version of Vue in your
...
+    // Options...
+  }).mount('#example');
+</script>
+```
+
+## Technical Notes
+
+- Docsify processes Vue content in the following order:
+  1. Execute markdown `<script>`
+  1. Register `vueComponents`
```
trusktr, line 332: It isn't specified anywhere, but I'm guessing that all components registered in `vueComponents` will be usable inside any other component (f.e. a component mounted via `vueMounts` can use a component registered with `vueComponent`).

If that's the case, it would be worth mentioning this.

docs/vue.md
```diff
@@ -20,118 +20,318 @@ To get started, load either the production or development version of Vue in your
...
+  1. Mount DOM elements via `vueMounts`
+  1. Process unmounted Vue content using `vueGlobalOptions`
+- Docsify will not mount an existing Vue instance or an element that contains an existing Vue instance.
+- Docsify will automatically destroy/unmount all Vue instances it creates before new page content is loaded.
+- When processing `vueGlobalOptions`, docsify parses the child elements within the main content area (`#main`) and mounts the element if it contains Vue content. Docsify does not parse each individual node within the main content area.
```
trusktr, line 337: I'm not clear on the last bullet point. I guess if it'll start to explain this magic part, it might just need to go into more detail.

docs/vue.md
trusktr: Ah, I see it uses `Vue.component` if Vue 2, etc, to register the components. So I guess that answers the question.

I think it'll be good to mention that they are global components that can be used in other components.

src/core/render/index.js
```diff
@@ -41,58 +43,179 @@ function formatUpdated(html, updated, fn) {
...
+              {},
+              docsifyConfig.vueGlobalOptions || {}
+            );
+
+            // Replace vueGlobalOptions data() return value with shared data object.
+            // This provides a global store for all Vue instances that receive
+            // vueGlobalOptions as their configuration.
+            if (vueGlobalData) {
+              vueConfig.data = function() {
+                return vueGlobalData;
```
trusktr, line 182: Neat

src/core/render/index.js
```diff
@@ -41,58 +43,179 @@ function formatUpdated(html, updated, fn) {
...
-
-        if (!isAlreadyVue) {
-          new window.Vue({
-            mounted: function() {
-              this.$destroy();
-            },
-          }).$mount(elm);
+    // Mount
+    for (const [mountElm, vueConfig] of vueMountData) {
+      const isVueAttr = 'data-isvue';
```
trusktr, line 193: `data-isvue` isn't documented in `vue.md`

trusktr: Oh, I see, it's a "private" attribute. What about using a more private thing like a JS property like `__docsify_is_vue`?

src/core/render/index.js
```diff
@@ -41,58 +43,179 @@ function formatUpdated(html, updated, fn) {
...
+        isMountedVue(mountElm) ||
+        // Has mounted instance(s)
+        mountElm.querySelector(`[${isVueAttr}]`);
+
+      if (!isSkipElm) {
+        mountElm.setAttribute(isVueAttr, '');
+
+        if (vueVersion === 2) {
+          vueConfig.el = undefined;
+          new window.Vue(vueConfig).$mount(mountElm);
```
trusktr, line 207: What happens if user wants to do the following?

```html
<style>
  .example { border: 3px solid cornflowerblue }
</style>

<div class="example">
  Hello {{name}}, this content should have a blue border around it.
</div>
```

```js
window.$docsify = {
  vueMounts: {
    '.example': { data: () => ({ name: 'Amadar' }) }
  }
}
```

Will the content have a blue border?

trusktr: Here's how I solved that problem:

https://github.com/trusktr/trusktr.io/blob/54664bac7a3cb3d10c0d36ad0e4d6a11ab8e610e/meteor-app/client/imports/routes/index.js#L81-L91

src/core/render/index.js
```diff
@@ -41,58 +43,179 @@ function formatUpdated(html, updated, fn) {
...
+          const app = window.Vue.createApp(vueConfig);
+
+          // Register global vueComponents
+          vueComponentNames.forEach(name => {
+            const config = docsifyConfig.vueComponents[name];
+
+            app.component(name, config);
+          });
+
+          app.mount(mountElm);
```
trusktr, line 218: I haven't tried Vue 3, not sure if it has the same issue as the comment about the blue border with Vue 2.

src/themes/vue.styl
```diff
@@ -103,28 +103,85 @@ body
...
+.markdown-section output
+  border-radius 2px
+  display block
+
+.markdown-section pre > code,
+.markdown-section output::after
+  -moz-osx-font-smoothing initial
+  -webkit-font-smoothing initial
+
+.markdown-section code
```
trusktr, line 130: Are these style changes potentially breaking (website will look different when v4 user updates)?

trusktr: Oh never mind, looks like only refactoring.

test/e2e/vue.test.js
```diff
@@ -1,141 +1,206 @@
...
-        const docsifyInitConfig = {
-          config: {
-            executeScript: false,
-          },
-          markdown: {
-            homepage: vueData.markdown,
-          },
-          scriptURLs: vueURL,
-        };
+      test(`ignores content when Vue is not present`, async () => {
```
trusktr, line 134: Perhaps this same exact test should run for the case when `Vue` is present, but `executeScript` is explicitly `false`, to ensure Vue doesn't render in that case.

trusktr: Oh wait, nevermind, it render Vue regardless of `executeScript`. That's better.

test/e2e/vue.test.js
```diff
@@ -1,141 +1,206 @@
...
+        await expect(page).toEqualText('#vuecomponent', '0');
+        await expect(page).toEqualText(
+          '#vueglobaloptions p',
+          'vueglobaloptions'
+        );
+        await expect(page).toEqualText('#vuemounts p', 'vueglobaloptions');
+        await expect(page).toEqualText('#vuescript p', 'vuescript');
+      });
+
+      test(`ignores <script> when executeScript is false`, async () => {
```
trusktr, line 195: This test doesn't seem to test that scripts were ignored, but that Vue still ran. Should the description be `renders Vue even if executeScript is false` or something similar?

docs/vue.md
jhildenbiddle: Agreed on handling this separately as this is a pre-existing limitation of docsify.

sy-records:
Thanks to your contribution. the vue content in documentation now looks great 👍🏻

docs/vue.md
jhildenbiddle: I've made a few minor to the docs to specify that `vueComponents` will be registered as _global_ components. I've also added links to related sections in the Vue docs for each of the following:

* `vueGlobalOptions`: https://vuejs.org/v2/api/#Options-Data
* `vueMounts`: https://vuejs.org/v2/guide/instance.html
* `vueComponents`: https://vuejs.org/v2/guide/components.html

I think we're better off linking to the official Vue docs to cover concepts like component composition.

docs/vue.md
jhildenbiddle: Yes, I agree this description could use some work. To be honest, I debated about even adding it. I've updated the description to (hopefully) make it more clear. At the end of the day, this text exists to help explain why this edge case won't work:

```js
window.$docsify = {
  vueGlobalOptions: {
    data() {
      return {
        globalMsg: 'Hello from vueGlobalOptions!',
      };
    },
  },
  vueMounts: {
    '#test': {
      data() {
        return {
          mountMsg: 'Hello from vueMounts!'
        };
      },
    },
  },
};
```

**Example 1:** This will work

```html
<!-- Source -->
<p>{{ globalMsg }}</p>
<p id="test">{{ mountMsg }}</p>
```
```html
<!-- Output -->
<p>Hello from vueGlobalOptions!</p>
<p id="test">Hello from vueMounts!</p>
```

In this example, docsify first mounts the top-level  `<p id="test">` from `vueMounts`. Docsify then iterates over the top-level child elements (`<p>`), skips the ones that contain or are themselves Vue instances, detects Vue template syntax in the top-level `<p>` element, and mounts it using `vueGlobalOptions`. Great.

**Example 2:** This will _not_ work

```html
<!-- Source -->
<div>
  <p>{{ globalMsg }}</p>
  <p id="test">{{ mountMsg }}</p>
</div>
```
```html
<!-- Output -->
<div>
  <p>{{ globalMsg }}</p>
  <p id="test">Hello from vueMounts!</p>
</div>
```

Like the previous example, docsify first mounts the top-level  `<p id="test">` from `vueMounts`. Docsify then iterates over the top-level child elements (`<div>`), detects that `<div>` contains a Vue instance (`<p id="test">`) and skips that element. This behavior is explained in the following "Technical Notes" bullet point:

> - Docsify will not mount an existing Vue instance or an element that contains an existing Vue instance.

I think it is highly unlikely that this will be an issue, but I added the related technical notes in hopes of making life easier for everyone if it does. The trick is how to do this succinctly without going into the same level of detail above. Hopefully the changes I've put in place are good enough for now. We can always revisit if/when needed.

src/core/render/index.js
jhildenbiddle: 😄

src/core/render/index.js
jhildenbiddle: I'm using a data attribute because it is faster and easier to detect Vue instances mounted by docsify using `querySelector('data-issue')` than it is to manually loop through a potentially large collection of nodes and test for a JS property like `__docsifyIsVue`. The decision was made easier once I noticed that Vue 3 adds its own data attribute to all Vue instances: `data-v-app`.

src/core/render/index.js
jhildenbiddle: Works fine as-is:

```js
vueMounts: {
  '#example': {
    data() {
      return {
        message: 'Looks Good!',
      };
    },
  },
},
```

```html
<style>
  #example { border: 3px solid cornflowerblue }
</style>

<div id="example">
  Test: {{message}}
</div>
```

Screenshot:

Screen Shot 2020-11-19 at 4 26 24 PM

src/core/render/index.js
jhildenbiddle: No issue. 😄

src/themes/vue.styl
jhildenbiddle: Non-breaking change because styles did not exist previously.

test/e2e/vue.test.js
jhildenbiddle: It's working as expected, but I can understand why the title could be confusing. 

Here is the relevant markup being tested:

```html
<div id="vuescript">
  <p v-text="msg">---</p>
</div>
```

* When the markdown `<script>` is executed
  * `#vuescript` is mounted directly by the `<script>`
  * The text content of `<p v-text="msg">` is changed to `"vuescript"`
* When the markdown `<script>` _not_ is executed
  * `#vuescript` is auto-mounted by docsify using `vueGlobalOptions`
  * The text content of `<p v-text="msg">` is changed to `"vueglobaloptions"`

That's why this test passes:

```js
await expect(page).toEqualText('#vuescript p', 'vueglobaloptions');
```

sy-records (approved):
LGTM.

Koooooo-7 (approved):
damnnnn good. :+1:

sy-records (approved):

sy-records (approved):

Koooooo-7 (approved):

docs/index.html
trusktr: Nvm, the source answers my question.

docs/vue.md
trusktr: @jhildenbiddle The example of what works vs what doesn't will be very helpful to some people. I think it'd be worth pasting what you wrote into the docs.

src/core/render/index.js
trusktr: @jhildenbiddle Did you test with Vue 2 or Vue 3? Or both?

PR #1414, opened 2020-11-01
We don't need to init if we're only testing marked. We probably don't…

trusktr:
… need to test marked either, as we assume that the marked library works as advertised in whichever way they tested it.


jhildenbiddle:
@trusktr FWIW, I don't think we need to test `marked` either. I only converted these tests from the old setup because I didn't want anyone to say "what happened to test X". I figured discussions about what tests we can get rid of could happen after the conversion.

trusktr:
Cool. Just testing anyway. The test didn't pass because importing marked directly does not include our custom additions.

PR #1423, opened 2020-11-10
Fix #1422

### LINKED ISSUE #1422: Embedded markdown fails due to cross site authenticaiton
#### Steps to reproduce
Exposing this issue requires the import of a markdown behind an authenticated site.  The bug was discovered when working on a site locally using the docsify-cli tool in which a markdown page had embedded another markdown from a corporate github enterprise system.   It's assumed that once the pages were uploaded to the same github instance the issue would be resolved, but this bug makes it impossible to work on such documentation on a local environment.

1. Create a markdown for embedding and store it into a github instance, which when served by pages will require authentication.
1. Create a new github pages project using the docsify embed feature attempt to embed the remote page.
1. Serve the new github pages project locally using the docsify-cli tool

When docsify attempts to fetch the protected resource instead of the markdown the request is redirected to the authentication page, which is returned and embedded into the markdown.  

When embedding a protected markdown that I have already authenticated to I should see the contents of that markdown embedded into the page.  

To solve this issue the XMLHttpRequest used to fetch the contents should set the [withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) value to true.  If the use has then successfully authenticated in the browser to the protected resource this would allow docsify to properly fetch the requested content.

- [X] Bug does still occur when all/other plugins are disabled?

- Your OS: RHEL v8.0.2
- Node.js version: v10.21.0
- npm/yarn version: 6.14.4
- Browser version: Firefox 81.0.2, Chrome 86.0.4240.183
- Docsify version: 4.11.6
- Docsify plugins: 

cmbaatz:
**Summary**

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/src/core/fetch/ajax.js
+++ b/src/core/fetch/ajax.js
@@ -30,6 +30,7 @@ export function get(url, hasBar = false, headers = {}) {
     }
   }
 
+  xhr.withCredentials = true;
   xhr.send();
 
   return {


Koooooo-7 (commented):
When the `xhr.withCredentials = true;`, the request can bring the `cookies` for authenticaiton in `headers` for the `cross-origin` request.
IMO, it's an unsafe operation/setting, maybe it could be better to be an option.

jhildenbiddle:
I agree with @Koooooo-7. We can't default to settings `withCredentials` to `true`.

My initial thought is that modifying the xhr object is best handled via a new hook instead of a configuration option. This provides greater flexibility, including the ability to modify the xhr object based on the request URL:

```js
window.$docsify = {
  plugins: [
    function(hook, vm) {
      hook.beforeSend(function(xhr, url) {
        // Domain-specific XHR settings
        if (/some-domain.com/.test(url)) {
          xhr.withCredentials = true;
          xhr.setRequestHeader("foo", "1");
          xhr.setRequestHeader("bar", "2");
        }
      });
    }
  ]
};
```

Thoughts @docsifyjs/core ?

anikethsaha:
I agree with @jhildenbiddle , Our config is getting huge lately. We need some kind of solution to limit that and customize futher. 

I agree with the hook and custom plugin.

anikethsaha:
@cmbaatz please create an issue with the details and we can have a discussion there.

PR #1432, opened 2020-11-28
feat: Support browser-level image lazy-loading

### LINKED ISSUE #1431: Support browser-level lazy-loading for images
## Feature request
Support lazy-loading images
Reference: https://web.dev/browser-level-image-lazy-loading/

For pages with many images, instead of loading all the images, using lazy-loading will allow the browser fetch images on demand.

I think we can expose an lazy loading option, or enable it by default.

Add  `loading=lazy` to `img` tags.

Not sure where to change.

sy-records:
**Summary**

Support browser-level image lazy-loading.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

resolve #1431 

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


anikethsaha (dismissed):

anikethsaha:
LGTM 
can you add a test case.

sy-records:
@anikethsaha done

src/core/render/compiler/image.js
```diff
@@ -34,6 +34,10 @@ export const imageCompiler = ({ renderer, contentBase, router }) =>
       attrs.push(`id="${config.id}"`);
     }
 
+    if (config.loading) {
+      attrs.push(`loading="${config.loading}"`);
+    }
+
```
Koooooo-7, line 40: I found that the `loading` attr is only works on specific browsers and version, we may need mention it on docs in case of users misunderstanding the behaviors in different browsers.

Koooooo-7: FYI.Chrome seems support it since Chrome75.
Firefox support it recently, it is unable by default tho (`dom.image-lazy-loading.enabled = false`).

sy-records: If the browser does not support lazy-loading then the attribute will be ignored and images will load immediately, as normal.

Koooooo-7 (approved):
LGTM.

anikethsaha (approved):

jhildenbiddle:
@sy-records, @anikethsaha, @Koooooo-7 --

This needs further discussion. Please hold off merging until I can provide detailed feedback...

jhildenbiddle:
@sy-records, @anikethsaha, @Koooooo-7 --

Two quick suggestions regarding issues and PRs moving forward:

1. We shouldn't be creating enhancement requests (#1431) immediately followed by a PR (this one) without some discussion taking place. The purpose of creating a new issue/feature request is to allow others to discuss the enhancement *before* code is written. If the implementation is critical to the discussion, a draft PR can be created an linked to in the issue.

1. The titles of issues and feature requests shouldn't dictate implementation. The title of #1431 would be better written as "Support lazy-loading of images". Native lazy-loading is just one of [many ways](https://www.sitepoint.com/five-techniques-lazy-load-images-website-performance/) to accomplish the goal. The goal of an enhancement (like lazy-loading) should be to solve a problem or provide additional value, not to "do a technical thing".

I've added detailed comments to #1431. Let's continue the discussion and make sure we've thought through the implementation and value proposition for end-users before moving forward.

Thanks!

jhildenbiddle (changes requested):
Thanks to @sy-records for the PR.

Unfortunately, I just don't see the value proposition of the implementation found in this PR. The changes allow adding an attribute to images using markdown syntax, but users familiar enough with "lazy-loading" and willing to modify their markdown to support it can just as easily do so using HTML (which is also valid markdown):

```markdown
![logo](image.jpg ':loading=lazy')
```
```html
<img src="image.jpg" loading="lazy">
 ```

If we want to support lazy-loading, let's make it a feature of docsify that all users can benefit from without technical knowledge or tedious work (like manually adding attributes to their markdown).

I've added detailed comments to #1431. Let's continue the discussion and make sure we've thought through the implementation and value proposition for end-users before moving forward.

Thanks!

trusktr (commented):
Looks good. 

Question for @jhildenbiddle: is that the recommended way to test it? Or should we test a regex instead of a snapshot?

trusktr (changes requested):
We should make config option names meaningful based on their purpose.

We should make the config option name `imgLoading` (or similar) instead of just `loading` (many things can "load").

PR #1483, opened 2021-01-29
fix: upgrade dompurify from 2.2.2 to 2.2.6

sy-records:
Replace #1470

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7233,9 +7233,9 @@
       }
     },
     "dompurify": {
-      "version": "2.2.2",
-      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-2.2.2.tgz",
-      "integrity": "sha512-BsGR4nDLaC5CNBnyT5I+d5pOeaoWvgVeg6Gq/aqmKYWMPR07131u60I80BvExLAJ0FQEIBQ1BTicw+C5+jOyrg=="
+      "version": "2.2.6",
+      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-2.2.6.tgz",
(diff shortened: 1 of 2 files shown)


Koooooo-7 (commented):
it seems there has the `security/snyk` issue on CI.
@anikethsaha  could u plz check and fix it  intead of manual upgrade？

anikethsaha:
Not sure whether I get this correctly but you are asking for a PR from snyk ? @Koooooo-7

Koooooo-7:
> Not sure whether I get this correctly but you are asking for a PR from snyk ? @Koooooo-7

yup, the PR of snyk always deletes the dependencies about `vue` in `package-lock.json`. f.e see #1470 .

Koooooo-7:
I guess we need move the `vue` from `devDependencies` to `dependencies` .

sy-records:
The reason I closed #1470 was because it was out of date, two versions short

sy-records:
Maybe snyk doesn't recognize this style

https://github.com/docsifyjs/docsify/blob/develop/package.json#L106-L107

jhildenbiddle:
> Maybe snyk doesn't recognize this style
> 
> https://github.com/docsifyjs/docsify/blob/develop/package.json#L106-L107

@sy-records -- Let's ping Snyk and let them know that their system appears unable to determine when two versions of the same library are listed as dependencies. Otherwise we're just going to run into this issue over and over again. Sounds like @anikethsaha maintains our account?

anikethsaha:
> > Maybe snyk doesn't recognize this style
> > https://github.com/docsifyjs/docsify/blob/develop/package.json#L106-L107
> 
> @sy-records -- Let's ping Snyk and let them know that their system appears unable to determine when two versions of the same library are listed as dependencies. Otherwise we're just going to run into this issue over and over again. Sounds like @anikethsaha maintains our account?

I think @sy-records does have the access, right ?
if not please tell me how to give access, I tried but it wasnt successfull.

sy-records:
I private messaged you in discord @anikethsaha

PR #1488, opened 2021-02-04
fix: Update language support

jhildenbiddle:
**Summary**

1. Dynamically set the HTML `lang` attribute based on URL hash. This attribute must be set properly in order to trigger auto-translation features on applicable Chromium-based browsers (Chrome, Edge). Note that this fix works only for the initial page load as `lang` attribute changes are not recognized on hash change. For example:

   https://docsify.js.org/
   
   ```html
   <html lang="en">
   ```

   https://docsify.js.org/#/zh-cn/
   
   ```html
   <html lang="zh-cn">
   ```

2. Remove the `write-a-plugin.md` alias as it appears to be unnecessary
3. Add Russian and Spanish search-related label translations

- [x] Feature
- [x] Docs

* **Before** (notice the lack of the translation option in the URL bar)
  Chrome1

* **After**
  Chrome2

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [x] Edge
- [ ] IE


index.html
```diff
@@ -1,99 +1,118 @@
...
+          },
+          paths: 'auto',
+          placeholder: {
+            '/es/': 'Buscar',
+            '/de-de/': 'Suche',
+            '/ru-ru/': 'Поиск',
+            '/zh-cn/': '搜索',
+            '/': 'Search',
+          },
+          // pathNamespaces: ['/es', '/de-de', '/ru-ru', '/zh-cn']
```
sy-records, line 78:
```suggestion
          pathNamespaces: ['/es', '/de-de', '/ru-ru', '/zh-cn']
```

Let's keep it

sy-records (approved):

sy-records:
It seems that the test snapshot file needs to be updated

index.html
jhildenbiddle: @sy-records Indeed. I meant to uncomment this before I checked it in. Good catch.

jhildenbiddle:
> It seems that the test snapshot file needs to be updated

@sy-records Yep. Good catch. I didn't test locally because I need that changes from #1487 to test on my ARM-based Mac. Should be all set now.

sy-records (approved):

Koooooo-7 (approved):
LGTM.

PR #1495, opened 2021-02-07
fix: Use legacy-compatible methods for IE11

jhildenbiddle:
**Summary**

Modern convenience methods like `Array.prototype.find()` and `String.prototype.startsWith()` are not compatible with legacy browsers like IE11. This PR replaces those methods with legacy-compatible approaches that accomplish the same thing.

- [x] Bugfix

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [x] IE


sy-records (approved):
LGTM. Thanks for your contribution!

Koooooo-7 (approved):
LGTM.

PR #1514, opened 2021-02-18
fix: the coverpage image display error

sy-records:
**Summary**

fix https://github.com/docsifyjs/docsify/issues/381#issuecomment-778588443

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

## Screenshots 

<details>
<summary>click</summary>

Chrome before | Chrome after
--- | ---
chrome | chrome-fix

Safari before | Safari after
--- | ---
safari | safari-fix
 </details>


Koooooo-7 (approved):

jhildenbiddle:
@sy-records How have we verified this fix? All of the screenshots appear to be on Chrome for Android.

sy-records:
I used ios safari also recurred, so the fix was made.

src/themes/basic/_coverpage.styl
```diff
@@ -15,8 +15,8 @@ section.cover
     opacity 0.8
     position absolute
     top 0
-    height 100%
-    width 100%
+    height 100vh
+    width 100vw
```
trusktr, line 19: If we do this, due to how `vw` works, we need to verify that the page does not start scrolling sideways, because `100vw` does not account for scrollbar width. See this:

https://stackoverflow.com/questions/23367345/100vw-causing-horizontal-overflow-but-only-if-more-than-one/23367686

Can you add screenshots in this PR?

trusktr (changes requested):

src/themes/basic/_coverpage.styl
sy-records: I added screenshots.

jhildenbiddle:
@sy-records 

There are a few issues with the existing CSS and the changes in this PR.

1. The `section.cover` element should be given a `min-height` value (not a fixed `height` value) because its height can exceed the viewport height.
2. The `section.cover` element's `width` does not need to be set in `vw` units. Best to leave this set to `100%`.
3. The `.mask` element needs to cover the entire `section.cover` element, so it also should not receive a fixed height. However, because this element uses `position: absolute` and contains no content of its own, we cannot use `min-height` here. Instead, it's best to remove the `height` value, set `bottom: 0`, and make sure `section.cover` is using `position: relative`.
4. BONUS: When the `section.cover` element comes close to or does exceed the viewport height, its content is often pushed against the top and bottom edges of the coverpage. By setting the top/bottom margin of the first/last elements in the coverpage, we can prevent this.

I can't edit your PR nor can I comment on unmodified lines here on GitHub so I've copy/pasted the updated CSS (Stylus) below. I'm also happy to create a separate PR to address the issue if preferred.

```stylus
section.cover
  position relative
  align-items center
  background-position center center
  background-repeat no-repeat
  background-size cover
  min-height 100vh
  width 100%
  display none

  &.show
    display flex

  &.has-mask .mask
    background-color $color-bg
    opacity 0.8
    position absolute
    top 0
    bottom 0
    width 100%

  .cover-main
    flex 1
    margin -20px 16px 0
    text-align center
    position: relative

    &:first-child
      margin-top 1em

    &:last-child
      margin-top 1em

  a
    color inherit
    text-decoration none

    &:hover
      text-decoration none

  p
    line-height 1.5rem
    margin 1em 0

  h1
    color inherit
    font-size 2.5rem
    font-weight 300
    margin 0.625rem 0 2.5rem
    position relative
    text-align center

    a
      display block

    small
      bottom -0.4375rem
      font-size 1rem
      position absolute

  blockquote
    font-size 1.5rem
    text-align center

  ul
    line-height 1.8
    list-style-type none
    margin 1em auto
    max-width 500px
    padding 0

  .cover-main > p:last-child a
    border-color var(--theme-color, $color-primary)
    border-radius 2rem
    border-style solid
    border-width 1px
    box-sizing border-box
    color var(--theme-color, $color-primary)
    display inline-block
    font-size 1.05rem
    letter-spacing 0.1rem
    margin 0.5rem 1rem
    padding 0.75em 2rem
    text-decoration none
    transition all 0.15s ease

    &:last-child
      background-color var(--theme-color, $color-primary)
      color #fff

      &:hover
        color inherit
        opacity 0.8

    &:hover
      color inherit

  blockquote > p > a
    border-bottom 2px solid var(--theme-color, $color-primary)
    transition color 0.3s

    &:hover
      color var(--theme-color, $color-primary)

```

jhildenbiddle (changes requested):
See https://github.com/docsifyjs/docsify/pull/1514#issuecomment-1027617521

sy-records:
@jhildenbiddle You can submit a PR.

PR #1521, opened 2021-02-25
fix: Fix duplicate sidebar when include audio file

sy-records:
**Summary**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

| before | after |
| --- | --- |
| image| image |


Koooooo-7 (changes requested):
IMHO, It seems not find where the issue is, we may need take more time look on it.

Koooooo-7 (changes requested):
IMHO, It seems not find where the issue is, we may need take more time look on it.

sy-records:
> 恕我直言，似乎找不到问题所在，我们可能需要花更多时间研究它。

src/core/render/compiler.js
```diff
@@ -235,7 +236,12 @@ export class Compiler {
       const slug = slugify(config.id || str);
       const url = router.toURL(router.getCurrentPath(), { id: slug });
       nextToc.slug = url;
-      _self.toc.push(nextToc);
+
+      // FIXME: Maybe we can do the de-duplication directly from _self.toc
+      if (_self.uniqueTOC.indexOf(url) === -1) {
+        _self.uniqueTOC.push(url);
+        _self.toc.push(nextToc);
+      }
```
trusktr, line 244: It is not clear what this changes or how the original problem happens. Can you explain?

sy-records: you can see https://discord.com/channels/713647066802421792/713647159882547261/813714253743128606

zenof:
I'm having the same issue. When I include more than 1 video file in the page, the sidebar menu will loop for several times if refresh the page or open the page via link. If will be correct if coming from another page. It is very annoying since I need to share the docs link to others and the menu looped like broken.

src/core/render/compiler.js
trusktr: You only pointed to some code. Can you explain?

media files should not have anything to do with the sidebar, so it **_seems_** this is not the correct fix, and that we should fix a problem with embedding media (or other stuff too maybe).

trusktr: What I mean is, that fact that this code seems entirely unrelated to media files, yet fixes the problem, seems more like a workaround than an actual fix.

firstly, why does adding media files cause duplication?

trusktr (changes requested):

PR #1571, opened 2021-05-10
Add beforeEmbed and afterEmbed hook

kevinresol:
## **Summary**

Add a "beforeEmbed", "afterEmbed" hook that will receive and return the markdown token list before and after the "embed file" mechanism takes place.

This will be useful for plugins that wish to pre-process the markdown. Currently one has to parse the markdown in `beforeEach`, work with the parsed token list and then re-stringify the tokens into markdown string for the next step, which is wasteful. With these hooks, one can just mess with the parsed token list and go on.

Feature

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE


kevinresol:
I am thinking about splitting it into beforeEmbed and afterEmbed...

trusktr:
Hello! This seems like a useful addition.

docs/write-a-plugin.md
```diff
@@ -18,9 +18,21 @@ window.$docsify = {
         return content;
       });
 
+      hook.beforeEmbed(function(tokens, next) {
+        // Invoked each time after parsing the Markdown file, but before embedding is applied to the token list.
```
trusktr, line 22: I think we need to change the wording, or otherwise describe what "embedding" is. I don't think that currently Docsify users have any concept of "embed", and the term "embed" is currently just an internal detail.

src/core/init/lifecycle.js
```diff
@@ -5,6 +5,8 @@ export function initLifecycle(vm) {
     'init',
     'mounted',
     'beforeEach',
+    'beforeEmbed',
+    'afterEmbed',
     'afterEach',
     'doneEach',
     'ready',
```
trusktr, line 12: We should list the order of hooks, otherwise people have to find this in the code right here.

src/core/render/embed.js
```diff
@@ -91,7 +91,10 @@ function walkFetchEmbed({ embedTokens, compile, fetch }, cb) {
   }
 }
 
-export function prerenderEmbed({ compiler, raw = '', fetch }, done) {
+export function prerenderEmbed(
+  { compiler, raw = '', fetch, beforeEmbed },
+  done
+) {
```
trusktr, line 97: Another thing is, this hook would depend on `marked`, thus is markdown-compiler-specific. What if we change renderers later?

Maybe it is good for now (performance boost) and if we change markdown renderer later then it would simply be a breaking change.

Arguably even without this hook, it would be a breaking change because output of the strings can change. But with this hook in place, it could be that a different markdown renderer does not have a concept of tokens that can be passed to a hook plugin.

Also, what if we wish to add the ability to specify markdown renderers? beforeEach and afterEach would continue to make sense, but maybe beforeEmbed would not make sense if a different markdown renderer does not pass tokens out.

So the question is: do we want to add this, and be coupled to `marked`? Or not?

trusktr: If we decide we want this, I think the way it is implemented looks good.

test/integration/docsify.test.js
```diff
@@ -33,6 +33,8 @@ describe('Docsify', function() {
 
             expect(hook.init).toBeInstanceOf(Function);
             expect(hook.beforeEach).toBeInstanceOf(Function);
+            expect(hook.beforeEmbed).toBeInstanceOf(Function);
+            expect(hook.afterEmbed).toBeInstanceOf(Function);
             expect(hook.afterEach).toBeInstanceOf(Function);
             expect(hook.doneEach).toBeInstanceOf(Function);
             expect(hook.mounted).toBeInstanceOf(Function);
```
trusktr, line 40: Sidenote, we need better hooks tests.

trusktr (commented):
If we decide we are okay to merge this an have hooks that are coupled to (require) `marked`, then the only thing we need to do is change or describe what "embed" means. This currently won't have any meaning to someone reading the docs.

trusktr (changes requested):
I meant to review as "Request changes", regarding the "embed" wording.

But let's chat first about if we want to have these hooks.

docs/write-a-plugin.md
kevinresol: I doubt that given there is a whole page about embedding files in the documentation: https://docsify.js.org/#/embed-files

trusktr: Oh, yeah, oops. 😄

Does marked give back tokens just for the embed? Or what exactly?

test/integration/docsify.test.js
trusktr: Would you mind adding tests so that it shows tokens being used. The other tests are bad (lacking, obviously), but moving forward we should test new features at least.

docs/write-a-plugin.md
kevinresol: "tokens" is an array of token, produced by the `compile.lexer(raw)` call. It is indeed a vendor-specific data structure produced by marked.

https://github.com/docsifyjs/docsify/blob/66303fec4c7115621e556ad742cfac9d19f26bd9/src/core/render/embed.js#L103

Docsify's embed mechanism operates on this token list as well.

docs/write-a-plugin.md
```diff
@@ -18,9 +18,21 @@ window.$docsify = {
         return content;
       });
 
+      hook.beforeEmbed(function(tokens, next) {
```
jhildenbiddle, line 21: I don't think we should pass `tokens` produced from a markdown parser to a plugin hook. As @trusktr pointed out, this makes the plugin functionality dependent on a specific markdown library which IMHO is not a good idea.

More importantly, I don't think we should pass "markdown before it becomes HTML" in two different formats (`string` vs `tokens`) to two very similar hooks:

```js
hook.beforeEach(function(content) {
  // ...
  return content;
});

hook.beforeEmbed(function(tokens) {
  // ...
  return tokens;
});
```

If plugins authors want to work with tokens, they can generate their own tokens by accessing Docsify's renderer directly within their plugin:

```js
window.$docsify = {
  plugins: [
    function(hook, vm) {
      hook.beforeEach(function(content) {
        const renderer = vm.config.markdown.renderer

        // Do renderer stuff...

        return content;
      });
    }
  ]
}
```

test/integration/docsify.test.js
jhildenbiddle: Hook tests were added in #1742 + #1770. Check out `/test/e2e/plugins.test.js`.

src/core/init/lifecycle.js
jhildenbiddle: The updates to the "Write a Plugin" page do a better job of describing when hooks are invoked. The updates aren't live yet, but they can be viewed on the `develop` branch [preview](https://docsify-preview.vercel.app/#/write-a-plugin).

jhildenbiddle (commented):
--

jhildenbiddle (changes requested):
This would make for a welcome addition. See in-line comments for code change requests.

FWIW, adding new `beforeEmbed` and `afterEmbed` hooks are a good approach for adding this feature as a non-breaking change. Long term, it may be preferable (TBD) to pass *all* markdown (cover, content, embeds, sidebar, navbar, etc.) to a single "do stuff with markdown" hook along with meta data to indicate what type of markdown it is. That's a conversation for a later date, but I wanted to mention it in case anyone was concerned with the growing number of plugin hooks.

jhildenbiddle:
@kevinresol --

Checking in. This PR kinda got lost, but if you're willing to make a few updates it would make for a nice addition to Docsify.

kevinresol:
Sorry but my memory about this has long gone and I don't right now have time to pick it up again...

Issue #1573, opened 2021-05-18
The formatUpdated config doesn't display any thing


Immortalt:
## Bug Report
You can see in this page's code:
https://github.com/immortal-blog/immortal-blog.github.io/blob/main/docs/index.html
and you can see that all pages don't have update time:
https://immortal-blog.github.io/#/golang/migrate-mgo-to-qmgo
I added the config as the doc mentioned,
```
window.$docsify = {
                formatUpdated: '{YYYY}-{MM}-{DD}',
```
but I still can't see any change in my blog. There is no time.
(I'm sure that I cleaned the cache before refreshing the page)
#### Steps to reproduce
1. Create a Github Page using docsify like https://immortal-blog.github.io/#/, or you can just fork this git.
2. Add the config option: formatUpdated: '{YYYY}-{MM}-{DD}'.
3. Clean the browser cache and refresh the Github Page.

#### What is current behaviour
Nothing changed.
There is no update time on the page.

#### What is the expected behaviour
There is an update time on the page.

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: Ubuntu 20.04 LTS
- Node.js version: v14.16.1
- npm/yarn version: 7.12.0
- Browser version: Firefox 88.0.1 x64
- Docsify version: 4.4.3
- Docsify plugins: disabled all

#### Please create a reproducible sandbox 

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/307qqv236)

sy-records:
you need add `{docsify-updated}`


```diff
    window.$docsify = {
        formatUpdated: '{YYYY}-{MM}-{DD}',
        name: 'Immortal Blog',
        repo: '',
        auto2top: true,
        coverpage: true,
        loadSidebar: true,
        maxLevel: 4,
        subMaxLevel: 2,
+        plugins: [
+            function(hook, vm) {
+                hook.beforeEach(function(html) {
+                    return (
+                        html +
+                        '\n----\n' +
+                        'Last modified {docsify-updated}'
+                    );
+                });
+            }
+        ]
+    }
```

Immortalt:
> ```diff
> +        plugins: [
> +            function(hook, vm) {
> +                hook.beforeEach(function(html) {
> +                    return (
> +                        html +
> +                        '\n----\n' +
> +                        'Last modified {docsify-updated}'
> +                    );
> +                });
> +            }
[...]

PR #1574, opened 2021-05-18
docs: Update docsify-updated to prevent replacement

sy-records:
## **Summary**

Update docsify-updated to prevent replacement

before | after
---|---
image | image

docs

## **For any code change,**

- [ ] Yes
- [x] No

## **Related issue, if any:**

#1573

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/docs/write-a-plugin.md
+++ b/docs/write-a-plugin.md
@@ -87,7 +87,7 @@ window.$docsify = {
           editHtml +
           html +
           '\n----\n' +
-          'Last modified {docsify-updated} ' +
+          'Last modified {docsify-updated}' +
           editHtml
         );
       });
@@ -108,4 +108,5 @@ Current version: <span id='tip-version'>loading</span>
 
 <script>
 document.getElementById('tip-version').innerText = Docsify.version
+document.getElementsByClassName("lang-js")[2].innerHTML = document.getElementsByClassName("lang-js")[2].innerHTML.replace(/Last modified .*'/,"Last modified {docsify-updated<span>}'</span>")
 </script>


Koooooo-7 (approved):
Haha, it reminds me back to #1322. 😅

sy-records:
yeah, This file was previously loaded using a remote, but then rolled back...

docs/write-a-plugin.md
```diff
@@ -108,4 +108,5 @@ Current version: <span id='tip-version'>loading</span>
 
 <script>
 document.getElementById('tip-version').innerText = Docsify.version
+document.getElementsByClassName("lang-js")[2].innerHTML = document.getElementsByClassName("lang-js")[2].innerHTML.replace(/Last modified .*'/,"Last modified {docsify-updated<span>}'</span>")
```
trusktr, line 111: This seems to reverse the breaking effect. Cab we instead stop the breaking in the first place? It can be easier to understand that way, instead if adding "Band-Aids" that make it more difficult to understand what is happening.

Where is the break happening?

sy-records: https://docsify.js.org/#/write-a-plugin?id=edit-button

trusktr: My "band aid" expression maybe sounded too harsh. It is a term my co-workers and I have used to describe quick fixes that don't solve the main issue, but add a workaround instead,  making the code less ideal for the short term gain.

Do you know what is causing the issue in the first place?

sy-records: Because it was replaced

https://github.com/docsifyjs/docsify/blob/407e4d4f3de78bebd639a3fdae751f8045728e57/src/core/render/index.js#L34-L43

trusktr (approved):

PR #1580, opened 2021-05-27
HTML escape included mermaid files to handle embedded tags

mgranell:
## **Summary**

Fixes an issue where included mermaid files that had embedded tags failed to render, due to the file being included verbatim into the <div> tag, and embedded html within the mermaid file "breaking out" of the div.

e.g. the following mermaid file will fail:
```
graph TB
  linkStyle default fill:#ffffff

  14["&lt;div style='font-weight: bold'&gt;This will break&lt;/div&gt;&lt;div style='font-size: 70%; margin-top: 0px'&gt;[Software System]&lt;/div&gt;&lt;div style='font-size: 80%; margin-top:10px'&gt;Description&lt;br /&gt;with HTML&lt;/div&gt;"]
```

  Bugfix

## **For any code change,**

- [X] Related documentation has been updated if needed
There are no tests that test in the include functionality

- [ ] Yes
- [X] No

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [X] Edge
- [ ] IE

### DIFF
--- a/src/core/render/embed.js
+++ b/src/core/render/embed.js
@@ -63,7 +63,13 @@ function walkFetchEmbed({ embedTokens, compile, fetch }, cb) {
             );
           } else if (token.embed.type === 'mermaid') {
             embedToken = [
-              { type: 'html', text: `<div class="mermaid">\n${text}\n</div>` },
+              {
+                type: 'html',
+                text: `<div class="mermaid">\n${text
+                  .replace(/&/g, '&amp;')
+                  .replace(/</g, '&lt;')
+                  .replace(/"/g, '&quot;')}\n</div>`,
+              },
             ];
             embedToken.links = {};
           } else {


mgranell:
Test failed due to timeout, not related to change

sy-records:
Hi, @mgranell 
Can you provide a screenshot of the before and after comparison?
I test here there is no difference.

src/core/render/embed.js
```diff
@@ -63,7 +63,13 @@ function walkFetchEmbed({ embedTokens, compile, fetch }, cb) {
             );
           } else if (token.embed.type === 'mermaid') {
             embedToken = [
-              { type: 'html', text: `<div class="mermaid">\n${text}\n</div>` },
+              {
+                type: 'html',
+                text: `<div class="mermaid">\n${text
+                  .replace(/&/g, '&amp;')
+                  .replace(/</g, '&lt;')
+                  .replace(/"/g, '&quot;')}\n</div>`,
```
trusktr, line 71: Maybe we should use a more robust solution like described in the following SO question so that it accounts for other characters, not just those specific ones?

https://stackoverflow.com/questions/18749591/encode-html-entities-in-javascript

trusktr (changes requested):

PR #1584, opened 2021-05-31
docs: add yarn global install docsify-cli

fynntang:


sy-records (approved):

Koooooo-7 (commented):
Thx for ur input !
IMO, we use the `npm` base to build this project, so we prefer to let end user fellow this way basicly .  
As same as the docs for contributions, such as `npm run dev`.
And I assume that the user who use `yarn` should know how to handle our project.

sy-records:
This is only for installing docsify-cli https://github.com/docsifyjs/docsify-cli#installation

fynntang:
> Thx for ur input !
> IMO, we use the `npm` base to build this project, so we prefer to let end user fellow this way basicly .
> As same as the docs for contributions, such as `npm run dev`.
> And I assume that the user who uses `yarn` should know how to handle our project.

@Koooooo-7  Thanks for the reminder. I'll make it up later

Koooooo-7:
> This is only for installing docsify-cli https://github.com/docsifyjs/docsify-cli#installation

yep, I knew it.
> And I assume that the user who uses yarn should know how to handle our project.

trusktr:
Thanks for the suggestion @fynntang. I don't think this is needed, because `npm` is standard, and if anyone uses an additional tool like `yarn`, `pnpm`, `ied`, etc, then they will know how to use that as an alternative to `npm`.

PR #1598, opened 2021-06-22
fix: allow also " inside of an embed

### LINKED ISSUE #1005: Embed files does not work
## Bug Report
By using the last release of docsify (from CDN https://cdn.jsdelivr.net/npm/docsify@4), the feature Embed files does not work.

#### Steps to reproduce

From the code below :
```
[filename](./test.md ":include")
```

The render will be a link name filename.

The test.md content should be rendered directly (embedded).

- Your OS: Windows
- Node.js version: v10.15.3
- npm/yarn version: npm 6.4.1
- Browser version: Chrome
- Docsify version: 4.4.0
- Docsify plugins: tried without any plugins, but I have the same issue

### LINKED ISSUE #1555: ':include' not working in Docsify 4.12.1
> image
>
image

> 
> Details:
> 
> * OS: macOS 11.2.3
> * Node.js version: 15.14.0
> * npm version: 7.8.0
> * Browser version: chrome 89.0.4389.114
> * Docsify version: 4.12.1
>

### LINKED ISSUE #1587: ":include" in double quotes doesn't recognized
#### Steps to reproduce

Try to embed external file using for example this link:
```
[readme](https://raw.githubusercontent.com/softvisio/core/main/README.md ":include")
```

Currently file will not be embedded, because `:include` tag is placed in double quotes.
This is the problem, because `prettier` markdown processor replaces single quotes with double and double quotes are legal for the link title.

I am expecting, that file will be embedded ti the parent document.

- [ x] Bug does still occur when all/other plugins are disabled?

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 4.12.1
- Docsify plugins: 

### LINKED ISSUE #1597: embed quotes bug
#### Steps to reproduce
1. Create empty docsify project
2. Add following to one markdown file
  ```markdown
  [gist: script.js](https://gist.githubusercontent.com/anikethsaha/f88893bb563bb7229d6e575db53a8c15/raw/script.js ":include :type=code")
  ```
3. See not renderd embed 

Embedded gist will not be renderd.

Embedded gist should be renderd.

This is not working because the type of quotes are not supported. The default prettier configuration changes the `''` to  `""`. And if you have auto format toggled on, you're not be able to save the file working.

- Your OS: any os
- Node.js version: any nodejs version
- npm/yarn version: any yarn/npm version
- Browser version: any browser and version
- Docsify version: 4.12.1
- Docsify plugins: none

[Edit docsify-template-forked-zlwql](https://codesandbox.io/s/docsify-template-forked-zlwql)

luwol03:
## **Summary**

I fixed a bug where `":include"` is not handled like `':include'`. This is not working because the type of quotes are not supported. The default prettier configuration changes the `''` to `""`. And if you have auto format toggled on, you're not be able to save the file with the right quotes.

I added added the `"` also the the replace regex

```js
str = str
      .replace(/^('|")/, '')
      .replace(/('|")$/, '')
```

## **For any code change,**

- [ ] Related tests have been updated or tests have been added (I cant find some tests for that already.)

- [ ] Yes
- [x] No

## **Related issue, if any:**

closes #1597 
closes #1587 
closes #1555 
closes #1005 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/test/unit/render-util.test.js
+++ b/test/unit/render-util.test.js
@@ -1,4 +1,7 @@
-const { removeAtag } = require('../../src/core/render/utils');
+const {
+  removeAtag,
+  getAndRemoveConfig,
+} = require('../../src/core/render/utils');
 
 const { tree } = require(`../../src/core/render/tpl`);
 
@@ -14,6 +17,46 @@ describe('core/render/utils', () => {
       const result = removeAtag('<a href="www.example.com">content</a>');
 
       expect(result).toEqual('content');
+    });
+  });
+
+  // getAndRemoveConfig()
+  // ---------------------------------------------------------------------------
+  describe('getAndRemoveConfig()', () => {
+    test('parse simple config', () => {
+      const result = getAndRemoveConfig(
+        `[filename](_media/example.md ':include')`
+      );
+
+      expect(result).toMatchObject({
+        config: {},
+        str: `[filename](_media/example.md ':include')`,
+      });
+    });
+
+    test('parse config with arguments', () => {
+      const result = getAndRemoveConfig(
+        `[filename](_media/example.md ':include :foo=bar :baz test')`
+      );
+
+      expect(result).toMatchObject({
+        config: {
+          foo: 'bar',
+          baz: true,
+        },
+        str: `[filename](_media/example.md ':include test')`,
+      });
+    });
+
+    test('parse config with double quotes', () => {
+      const result = getAndRemoveConfig(
+        `[filename](_media/example.md ":include")`
+      );
+
+      expect(result).toMatchObject({
+        config: {},
+        str: `[filename](_media/example.md ":include")`,
+      });
     });
   });
 });
--- a/docs/embed-files.md
+++ b/docs/embed-files.md
@@ -16,7 +16,7 @@ Then the content of `example.md` will be displayed directly here;
 
 You can check the original content for [example.md](_media/example.md ':ignore').
 
-Normally, this will compiled into a link, but in docsify, if you add `:include` it will be embedded.
+Normally, this will compiled into a link, but in docsify, if you add `:include` it will be embedded. You can use single or double quotation marks around as you like.
 
 External links can be used too - just replace the target. If you want to use a gist URL, see [Embed a gist](#embed-a-gist) section.
 
--- a/src/core/render/utils.js
+++ b/src/core/render/utils.js
@@ -23,8 +23,8 @@ export function getAndRemoveConfig(str = '') {
 
   if (str) {
     str = str
-      .replace(/^'/, '')
-      .replace(/'$/, '')
+      .replace(/^('|")/, '')
+      .replace(/('|")$/, '')
       .replace(/(?:^|\s):([\w-]+:?)=?([\w-%]+)?/g, (m, key, value) => {
         if (key.indexOf(':') === -1) {
           config[key] = (value && value.replace(/&quot;/g, '')) || true;


sy-records (commented):
Can you add a test for this?

Koooooo-7:
@sy-records 
As we all know, this func `getAndRemoveConfig ` has been used for much places.
It seems we need separate it for different functions in case of unpredicted issues.
WDTY?

Koooooo-7 (commented):
@sy-records
As we all know, this func getAndRemoveConfig has been used for much places.
It seems we need separate it for different functions in case of unpredicted issues.
WDTY?

sy-records:
Yes, you're right.

trusktr:
> It seems we need separate it for different functions in case of unpredicted issues.

@Koooooo-7 What do you mean?

Koooooo-7:
> > It seems we need separate it for different functions in case of unpredicted issues.
> 
> @Koooooo-7 What do you mean?

This func r working in multi place to get configs. such as titles ,  sidebar headings.
If we add " for embed files,  it might cause other problems in other places.
so, we may need separate this func , for sidebar, for titles and for embed files.

trusktr:
@Koooooo-7 I see what you're saying, although it seems very unlikely. My guess would be that there is no one relying on he non-working double quotes.

It seems multiple people have hit this issue, and we closed the issues. It was in fact working before, then it broke:

- https://github.com/docsifyjs/docsify/issues/1587
- https://github.com/docsifyjs/docsify/issues/1555
- https://github.com/docsifyjs/docsify/issues/1005

Koooooo-7:
@trusktr  I checked this func history, it seems never support the `"` before, the docs either.
You could check [this func](https://github.com/docsifyjs/docsify/pull/708/files) since `Dec 5, 2018` and earlier.
If we support it right now, it would remove the `"` in other places also. 
So, I prefer to separate this func only for different places.

luwol03:
> Can you add a test for this?

sure, will try that.

> > > It seems we need separate it for different functions in case of unpredicted issues.
> > 
> > 
> > @Koooooo-7 What do you mean?
> 
> This func r working in multi place to get configs. such as titles , sidebar headings.
> If we add " for embed files, it might cause other problems in other places.
> so, we may need separate this func , for sidebar, for titles and for embed files.

I dont think so. I would also like to use the double ones inside of headlines, or other stuff. Everywhere I can use this style.

And another question, do you prefere an clean commit history, so should I rebase merge upstream changes?

Koooooo-7:
@luwol03  of cuz we can support this feat for those configs. instead, how about those titles/other stuff without configs but with " ?

luwol03:
> @luwol03 of cuz we can support this feat for those configs. instead, how about those titles/other stuff without configs but with " ?
> e.g a title `this is "my title"`.

Then we should improve the function, so that it only replaces the quotes, if they are followed by a `:xxxx` and there is also a closing one.

Koooooo-7:
> > @luwol03 of cuz we can support this feat for those configs. instead, how about those titles/other stuff without configs but with " ?
> > e.g a title `this is "my title"`.
> 
> Then we should improve the function, so that it only replaces the quotes, if they are followed by a `:xxxx` and there is also a closing one.

agreed.
If possible, glad to see a good solution for this func/regx to solve those issues above.

Koooooo-7 (approved):

luwol03:
Is this normal, that your e2e tests failed?

Koooooo-7:
> Is this normal, that your e2e tests failed?

It seems not this changes broken, it's okay.
I would retry it.

trusktr:
> If we support it right now, it would remove the `"` in other places also.

@Koooooo-7 can you point them out?

trusktr:
> Is this normal, that your e2e tests failed?

Sometimes it fails for some reason despite that it shouldn't. Needs a fix.

luwol03:
Before merge, I will try that:

> Then we should improve the function, so that it only replaces the quotes, if they are followed by a :xxxx and there is also a closing one.

trusktr:
> Before merge, I will try that:
> 
> > Then we should improve the function, so that it only replaces the quotes, if they are followed by a :xxxx and there is also a closing one.

Is anyone using it without the `:`?

trusktr:
Here is an example without the colon: https://github.com/docsifyjs/docsify/issues/1587#issuecomment-854293984

test/unit/render-util.test.js
```diff
@@ -16,6 +19,46 @@ describe('core/render/utils', () => {
...
+      const result = getAndRemoveConfig(
+        `[filename](_media/example.md ":include")`
+      );
+
+      expect(result).toMatchObject({
+        config: {},
+        str: `[filename](_media/example.md ":include")`,
+      });
+    });
+  });
```
trusktr, line 61: I wonder what the behavior should actually be. Seems funky that it leaves things behind. Seems like it should handle everything.

trusktr (approved):

trusktr:
@jhildenbiddle One test or other keeps randomly failing.

trusktr:
Investigating the failures. Let's not merge until we know it isn't the change from here. Seems unlikely that it is.

trusktr (changes requested):
Looks good. I'm not requesting any changes, just blocking from merging to ensure the test is just flaky (that's what it seems like).

trusktr (approved):

trusktr:
> improve the function, so that it only replaces the quotes, if they are followed by a :xxxx and there is also a closing one.

I don't think we should do that, it could break this: `"this-is-currently-ignored :but=this-currently-works"`. If we do, it can be a new PR.

Thanks @luwol03! Merging, since tests pass except one at random, which seems unrelated.

luwol03:
Oh, A few seconds ago, I pushed my changes, before I relized, that you already merged this changes :joy:

sy-records:
@luwol03 You can open a new PR

luwol03:
> @luwol03 You can open a new PR
@sy-records done - see #1603

PR #1643, opened 2021-10-01
Build commands are needed for AWS Amplify

cnpcshangbo:
## **Summary**
Based on tests from my end and AWS Amplify. The build commands are needed to show pages.

Docs

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


sy-records:
Does docsify need to be built even if it only uses index.html as an entry point?

Koooooo-7 (commented):
It makes no sense for docsify to build dependencies when u import it as a single script.
Did u use the source code to set up your site instead ?

Issue #1695, opened 2021-12-24
How to use Google Analytics v2 API


ekoz:
#### What problem does this feature solve?
I use ga plugin, set the right ga code, but Google Analytics can not recived data, ga.min.js request param is `v=1`, but Google Analytics gtag.js request param is `v=2`

![1640327720(1)](https://user-images.githubusercontent.com/3312512/147325318-adeec282-4b6e-4dda-9492-5736198f2762.png)

#### Are you willing to work on this yourself?
Yes

trusktr:
Hello, not sure what issue you're having. Did you try https://docsify.js.org/#/plugins?id=google-analytics ?

ekoz:
@trusktr  yes, I use ga plugin like this:
```
<script src="//cdn.jsdelivr.net/npm/docsify/lib/docsify.min.js" data-ga="UA-XXXXX-Y"></script>
<script src="//cdn.jsdelivr.net/npm/docsify/lib/plugins/ga.min.js"></script>
```
but in did, the http request param is v1 version, but Google Analytics api need v2 version

trusktr:
Do you mean we need a `ga2.js` plugin, that calls v2 APIs? It should be fairly simple if you want to make it. The code is short: 

https://github.com/docsifyjs/docsify/blob/develop/src/plugins/ga.js

If you can submit a new ga2.js, that'd be awesome.

trusktr:
Also if you post a live example, that'd help. I'm not familiar with ga.

ekoz:
@trusktr Thanks for your reply, maybe my description is not accurate enough. I checked the code script carefully, the v1 code script src is `https://www.google-analytics.com/analytics.js`, the v2 code script src is `https://www.googletagmanager.com/gtag/js?id=xxx`, and I use ga plugin, so it is not work. I will try to write the gtag plugin, and create merge request to u.

idmarjr:
Hello folks!

I can see #1702 have a test failing and it does not see any action for few months. 
Can I help somehow?

Also, there's any documented work around to make Google Analytics work meanwhile?

PR #1702, opened 2021-12-29
feat: add google analytics gtag.js plugin

ekoz:
## **Summary**

Google Universal Analytics (gtag.js) plugin, usage is the same as `src/plugins/ga.js`

Feature

## **For any code change,**

- [ ] Yes
- [ ] No

## **Related issue, if any:**
#1695

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


trusktr:
Nice, thanks! Can you add a similar doc section as GA? Also it needs lint fix.

ekoz:
@trusktr Hi, I update this merge request again, pls check it when you are free, thanks ~

ekoz:
@trusktr Hi, there is an error,  `ci/codesandbox — Building packages failed.`, Is there anything I can do?

trusktr:
Hey @ekoz. Updated and re-running the checks.

trusktr (dismissed):

src/plugins/gtag.js
```diff
@@ -0,0 +1,47 @@
...
+  window.gtag('event', 'page_view', {
+    page_title: document.title,
+    page_location: location.href,
+    page_path: location.pathname,
+  });
+}
+
+const install = function (hook) {
+  if (!$docsify.ga) {
+    console.error('[Docsify] ga is required.');
```
sy-records, line 40: Should it be gtag?

```suggestion
  if (!window.gtag) {
    init($docsify.gtag);
  }

  // usage: https://developers.google.com/analytics/devguides/collection/gtagjs/pages
  window.gtag('event', 'page_view', {
    page_title: document.title,
    page_location: location.href,
    page_path: location.pathname,
  });
}

const install = function (hook) {
  if (!$docsify.gtag) {
    console.error('[Docsify] gtag is required.');
```

sy-records (changes requested):

sy-records (approved):

jhildenbiddle:
@ekoz --

**TL;DR:** I believe we need to update the existing `ga.js` plugin with the changes from this PR and allow the `ga` config option to accept both a string *and* an array of property IDs. Details below.

I am by no means a Google Analytics expert, so forgive me if I'm off base here.

My understanding is that there are three main analytics products offered by Google:

1. [Google Analytics](https://developers.google.com/analytics/devguides/collection/analyticsjs#the_javascript_measurement_snippet) (a.k.a. "Classic Analytics")
   This is the original analytics product launched in 2008. It was deprecated ~2014 in favor of Google Universal Analytics, and according to [this site](https://www.analyticsmania.com/post/difference-between-google-analytics-and-universal-analytics/) all "Classic" properties were transitioned to "Universal" properties in 2016. Support is provided via Google's `analytics.js` snippet.
1. [Google Universal Analytics](https://developers.google.com/analytics/devguides/collection/gtagjs#install_the_global_site_tag) (a.k.a. "GA3")
   This is the successor to "Classic Analytics" which became Google's recommended/default product in 2013. Google [recently announced](https://blog.google/products/marketingplatform/analytics/prepare-for-future-with-google-analytics-4/) that they will be "sunsetting" this product on July 1, 2023 in favor of Google Analytics 4 (below). Support is provided via Google's `gtag.js` snippet.
1. [Google Analytics 4](https://support.google.com/analytics/answer/10089681?hl=en) (a.k.a. "GA4")
   This is the successor to "GA3" and has been Google's recommended/default product since October 14, 2020. Support is also provided via Google's `gtag.js` snippet.

Summary:

- "Classic" analytics is dead. There are no sites using this product today.
- "GA3" is in use today but will be replaced soon(ish).
- "GA4" is in use today and is what Google recommends.

Summary:

- If there are no "Classic" analytics sites, there is no reason for Docsify's `ga.js` plugin to use the "Classic" `analytics.js` snippet.
- If all Google analytics properties are either GA3 or GA4 and both GA3 and GA4 use Google's `gtag.js` snippet, we can simply update the `ga.js` plugin to use the `gtag.js` snippet without introducing a breaking change and everyone will be better off for it.

Based on Google documentation [here](https://support.google.com/analytics/answer/9310895?hl=en#zippy=%2Cin-this-article) and [here](https://support.google.com/analytics/answer/10271001?hl=en#zippy=%2Cin-this-article), it looks like the `gtag.js` snippet is configured for GA3 or GA4 based on the ID passed to it:

> Universal Analytics property IDs have the format "UA-XXXXXXXX". This is sometimes referred to as the Tracking ID. In this guide, we'll refer to it as the UA Property ID. Google Analytics 4 property web data streams use a Measurement ID with the format "G-XXXXXXXX".

```js
gtag('config', '<Some Property ID>');
```

This means there is no need for separate GA3/GA4 plugins. Users pass their property ID to the plugin via the `ga` config option, the plugin passes that to the `gtag.js` snippet, then Google handles the rest.

```js
window.$docsify = {
  ga: 'UA-XXXXXXXX' // Google Universal Analytics (GA3)
}
```

```js
window.$docsify = {
  ga: 'G-XXXXXXXX' // Google Analytics 4 (GA4)
}
```

It's worth noting that it is possible to use multiple IDs with the `gtag.js` snippet. We could extend the plugin to accept an Array as well a string. Each array item would then be set using `gtag('config', '<Some Property ID A>')`:

```js
window.$docsify = {
  ga: [
    'UA-XXXXXXXX', // Google Universal Analytics (GA3)
    'G-XXXXXXXX', // Google Analytics 4 (GA4)
    'AW-XXXXXXXX', // Google Ads
    'DC-XXXXXXXX' // Floodlight
  ]
}
```

This would have the same result as the following:

```js
gtag('config', 'UA-XXXXXXXX');
gtag('config', 'G-XXXXXXXX');
gtag('config', 'AW-XXXXXXXX');
gtag('config', 'DC-XXXXXXXX');
```

All of this boils down to two change requests:

1. Update the existing `ga.js` plugin with the changes made in this PR
2. Update the `ga.js` plugin to accept an array of property IDs instead of just a string

We'll need a GA3 and a GA4 site to test on. If you don't have access to those, I have some public docsify sites I can test on.

jhildenbiddle (changes requested):
See comment: https://github.com/docsifyjs/docsify/pull/1702#issuecomment-1071393522

jhildenbiddle:
Hi @ekoz. Checking in to see if/when the requested changes can be made so we can move forward with this PR.

Thanks!

ekoz:
@jhildenbiddle I want to design `$docsify.ga` to support both strings and arrays, but there are relatively few data for Google ads and floodlight. Can the current version only support GA3 and GA4?

jhildenbiddle:
@ekoz --

It is my understanding that the only thing required to support multiple properties ("G-", "AW-", "DC-", ...) or multiple identifiers for the same property ("G-XX1", "G-XX2", ...) is to render a separate `gtag()` statement for each:

```js
// GA3
gtag('config', 'UA-XXXXXXXX');

// Multiple GA4
gtag('config', 'G-XXXXXXX1');
gtag('config', 'G-XXXXXXX2');

// Other Services
gtag('config', 'AW-XXXXXXXX');
gtag('config', 'DC-XXXXXXXX');
```

This is based on the following documentation:

* https://developers.google.com/tag-platform/gtagjs/install
* https://developers.google.com/analytics/devguides/collection/gtagjs/#configure_additional_google_analytics_properties

So the implementation should be simple: just iterate over the array of IDs and render a `gtag('${id}')` for each.

It's worth nothing that [this post](https://stackoverflow.com/questions/46789189/google-analytics-gtag-multiple-analytics-account-tracking-ids) on StackOverflow seems to imply that this method of supporting multiple IDs for the same service does not always work. I can't verify this, but since but since Google's official documentation states that it this is how it is supposed to work, I think we're okay just rendering a `gtag()` for each ID and calling it a wrap.

ekoz:
@jhildenbiddle Hi, The idea you pointed out should be achievable, and the first ID in arrays must be a global site tag id, each page should have only one global site tag instance.

docs/plugins.md
```diff
@@ -78,6 +78,16 @@ Install the plugin and configure the track id.
   window.$docsify = {
     ga: 'UA-XXXXX-Y',
   };
+  
+  // or multi gtag, the first element of the array must be a global site gtag id, each page should have only one global site tag instance.
```
jhildenbiddle, line 82: This is confusing. Google's own terminology is confusing, so this isn't your fault. Let's simplify this for end users so they don't have to deal with Google's mess.

First, let's make sure we're using the correct terminology so people can easily locate and find more information on the IDs we're asking for. I'm not familiar with the term "global site gtag id", and when I search for it I do not find any results that match this  exact term (Google instead asks if I meant "global site tag id"). Are you referring to a GA3 "tracking ID" or a GA4 "measurement ID" here? If so, let's say this explicitly. I believe both of these IDs can be referred to as a "Google Property ID", but I'm not 100% clear on this.

Another idea is to remove the need for users to properly order their IDs and instead do the work for them to determine which ID should be used to initialize the `gtag` script. For example, instead of requiring users to list a GA4 or GA3 ID as the first item in the array, the plugin could search the array for the first instance of a preferred ID type. For example:

```js
window.$docsify = {
  ga: [
    'AW-XXXXXXXX', // Google Ads
    'DC-XXXXXXXX', // Floodlight
    'G-XXXXXXXX', // Google Analytics 4 (GA4)
    'UA-XXXXXXXX', // Google Universal Analytics (GA3)
  ],
};
```

1. Plugin searches array
2. Plugin finds GA4 "measurement ID" (`'G-XXXXXXXX'`)
3. Plugin passes GA4 ID to `initGlobalSiteTag()`
4. Plugin passes remaining IDs to `initAdditionalTag()`

The result is the same as if the end user made the GA4 ID the first item in the array, but by doing this work for them the user did not have to spend time trying to figure out the difference between a "Google Property ID", "Tracking ID", "Measurement ID", etc.

Thoughts?

jhildenbiddle (changes requested):

jhildenbiddle:
One other request: Please add tests for the plugin. Basic tests that verify that the script is being added to the DOM properly, that IDs match, and that the `gtag` calls are being made as expected is all we need.

Thanks @ekoz!

jhildenbiddle:
@ekoz Ping! 😄

docs/plugins.md
```diff
@@ -91,6 +91,27 @@ Configure by `data-ga`.
...
+  }
+</script>
+<script src="//cdn.jsdelivr.net/npm/docsify/lib/docsify.min.js"></script>
+<script src="//cdn.jsdelivr.net/npm/docsify/lib/plugins/gtag.min.js"></script>
+```
+
+Configure by `data-ga`.
+
+```html
+<script src="//cdn.jsdelivr.net/npm/docsify/lib/docsify.min.js" data-ga="UA-XXXXX-Y"></script>
```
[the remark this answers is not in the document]
ekoz: yes, it also works, only support single gtag id

ekoz:
@jhildenbiddle  I use an HTML page under `docsify` directory to test
```
<html>
  <head>
    <title>ga test</title>
    <link rel="stylesheet" href="lib/themes/vue.css" />
  </head>
  <body>
    <div id="app"></div>
    <script>
      window.$docsify = {
        // ga: [
        //   'AW-XXXXXXXX', // Google Ads
        //   'DC-XXXXXXXX', // Floodlight
        //   'G-XXXXXXXX', // Google Analytics 4 (GA4)
        //   'UA-XXXXXXXX', // Google Universal Analytics (GA3)
        // ],
      };
    </script>
    <script src="lib/docsify.min.js" data-ga="G-XXXXXXXX"></script>
    <script src="lib/plugins/ga.min.js"></script>
  </body>
</html>
```

docs/plugins.md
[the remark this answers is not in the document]
jhildenbiddle: Ahh. Got it. Assigning configuration values via data attributes is handled elsewhere. Thx. :)

docs/plugins.md
jhildenbiddle: First, where does the requirement that "the first element of the array must be a global site gtag id" come from? I don't see this listed anywhere in Google's documentation or in places like StackOverflow:

   - https://developers.google.com/analytics/devguides/collection/gtagjs/
   - https://developers.google.com/tag-platform/gtagjs/install
   - https://stackoverflow.com/questions/46789189/google-analytics-gtag-multiple-analytics-account-tracking-ids

If this isn't a requirement, then we can remove the comment and process the IDs in the order they are listed (as you are now).

If there *is* a requirement for specific ID types to be listed first in the array, it's probably best to just identify those ID types specifically in the comment so it's easier for everyone to understand. For example:

```js
<script>
  // Single property
  window.$docsify = {
    ga: 'UA-XXXXX-Y',
  };

  // Multiple properties
  // IDs are processed in order the order listed
  // Analytics IDs should ordered before other IDs
  window.$docsify = {
    ga: [
      'G-XXXXXXXX', // Google Analytics 4 (GA4)
      'UA-XXXXXXXX', // Google Universal Analytics (GA3)
      'AW-XXXXXXXX', // Google Ads
      'DC-XXXXXXXX', // Floodlight
    ],
  };
</script>
```

This seems like an easier and more flexible solution that searching the array for specific ID types (which is what I recommended previously).

Thoughts?

ekoz:
@jhildenbiddle thanks, I read the docs [https://playwright.dev/docs](https://playwright.dev/docs) and have a try, run with `npm run test:e2e ga.test.js`. 

How to test `data-ga` with playwright?

docs/plugins.md
ekoz: I understand, the current way is searching the array for specific ID types, which begin with `G-`.

jhildenbiddle: @ekoz --

Understood. But is this necessary? Where does the requirement come from that states specific types of IDs need to  go first?

I am not doubting you, but I'd like to understand it so we can make sure our code and documentation are as they should be.

ekoz: [https://developers.google.com/tag-platform/gtagjs/install](https://developers.google.com/tag-platform/gtagjs/install)

image

Maybe I misunderstood this paragraph. I think a `global site tag` is needed here. Look at the Google demo, it may not be the `G-`tag. So, how to make the user specify this `global site tag`?

jhildenbiddle: @ekoz --

I believe the "global site tag" referred to in the documentation is the `<script>` tag that loads the `gtag` snippet:

```html
<script async src="https://www.googletagmanager.com/gtag/js?id=DC-ZZZZZZ"></script>
```

The documentation is saying that if you have multiple IDs, users **_SHOULD NOT_** configure them doing this:

```html
<script async src="https://www.googletagmanager.com/gtag/js?id=G-XXXXXX"></script>
<script async src="https://www.googletagmanager.com/gtag/js?id=AW-YYYYYY"></script>
<script async src="https://www.googletagmanager.com/gtag/js?id=DC-ZZZZZZ"></script>
```

Instead, users should load the `gtag` snippet _once_, pass the first ID via a URL parameter to the snippet (e.g, `?id=G-XXXXXX`), then pass additional IDs using `gtag('configure', /* ID */)`.

Based on this, I don't think it matter which type of ID is passed as a URL parameter or via `gtag('configure')`. This will simplify things. We can remove the logic that searches for specific ID types and just process them in the order the appear in the array. The documentation can also be simplified:

```js
// Single ID
window.$docsify = {
  ga: 'UA-XXXXX-Y',
};

// Multiple IDs
window.$docsify = {
  ga: [
    'G-XXXXXXXX',  // Google Analytics 4 (GA4)
    'UA-XXXXXXXX', // Google Universal Analytics (GA3)
    'AW-XXXXXXXX', // Google Ads
    'DC-XXXXXXXX', // Floodlight
  ],
};
```

ekoz: ok, I think the data collector can be custom in `window.$docsify`, so that the user can define the data to be collected. If the user does not specify it, the default collector will be used. Do you think so?
```

window.$docsify = {
  ga: [
    'G-XXXXXXXX',  // Google Analytics 4 (GA4)
    'UA-XXXXXXXX', // Google Universal Analytics (GA3)
    'AW-XXXXXXXX', // Google Ads
    'DC-XXXXXXXX', // Floodlight
  ],

  // custom gaCollector
  gaCollector: function () {
    window.gtag('event', 'conversion', {
      send_to: 'AW-YYYYYY/AbC-D_efG-h12_34-567',
      value: 1.0,
      currency: 'USD',
    });

    // Measure Floodlight conversions
    window.gtag('event', 'conversion', {
      allow_custom_scripts: true,
      send_to: 'DC-ZZZZZZ/actions/locat304+standard',
    });
  },
};

// default ga collector
window.gtag('event', 'page_view', {
    page_title: document.title,
    page_location: location.href,
    page_path: location.pathname,
});
```

ekoz:
@jhildenbiddle Ping! 😄

src/plugins/ga.js
```diff
@@ -1,31 +1,61 @@
...
 
-function init(id) {
-  appendScript();
-  window.ga =
-    window.ga ||
+// global site tag instance initialized
+function initGlobalSiteTag(id) {
+  appendScript(id);
+
+  window.dataLayer = window.dataLayer || [];
```
jhildenbiddle, line 15: Let's move the `window.dataLayer` and `window.gtag` setup *outside* of the `initGlobalSiteTag` function. This will allow users to call `gtag()` so long as it is done after the plugin's `<script>` tag:

**Plugin:**

```js
window.dataLayer = window.dataLayer || [];
window.gtag =
  window.gtag ||
  function () {
    window.dataLayer.push(arguments);
  };

$docsify.plugins = [].concat(install, $docsify.plugins);
```

**HTML:**

```html
<script src="//cdn.jsdelivr.net/npm/docsify@4/lib/docsify.min.js"></script>
<script src="//cdn.jsdelivr.net/npm/docsify@4/lib/plugins/gtag.min.js"></script>
<script>
// Advanced users can call `gtag()` on their own
window.gtag('event', 'conversion', {
  allow_custom_scripts: true,
  send_to: 'DC-ZZZZZZ/actions/locat304+standard',
});
</script>
```

test/e2e/ga.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+    page.on('response', response => {
+      const request = response.request();
+      // googleads.g.doubleclick.net
+      // www.google-analytics.com
+      // www.googletagmanager.com
+      const reg =
+        /googleads\.g\.doubleclick\.net|www\.google-analytics\.com|www\.googletagmanager\.com/g;
+      if (request.url().match(reg)) {
+        console.log(request.url(), response.status());
```
jhildenbiddle, line 33: Let's remove or comment out the `console.log` tests so we don't see them in every test run.

jhildenbiddle (changes requested):
Hey @ekoz.

Apologies for the delayed response. Let's see if we can wrap this up and get it shipped. :smile:

- Now that I better understand the purpose of the `gtag` snippet, I think we are better off creating a new `gtag` plugin (as you originally proposed). I initially thought the snippet GA4-specific, but my understanding is that it is more of a universal snippet for Google services. Knowing this, it would be inappropriate to imply that this is a "Google Analytics" plugin ("ga.js") and instead make it clear that this is for Google's universal gtag plugin ("gtag.js").
- If we create a new "gtag" plugin, we'll need to update the `$docsify` property names as well (e.g., `gtag` instead of `ga`).
- I don't think we need to worry about the more complex scenarios handled by `gaCollector` as proposed in previous comments. Advanced users that require these configurations can simply call `gtag()` in a `<script>` on their own (see below). We should call this out in the docs as well.

What this all boils down to:

1. Rename plugin to `gtag.js` and make necessary updates to accommodate name change
2. Fix issues outlined below

Thanks for your patience with all of this back and forth, @ekoz. As I said early on, I'm not a Google Analytics expert so much of the churn has been caused by me learning more about GA and the `gtag` snippet as we go.

src/plugins/ga.js
```diff
@@ -1,31 +1,61 @@
...
 function collect() {
   if (!window.ga) {
     init($docsify.ga);
   }
 
-  window.ga('set', 'page', location.hash);
-  window.ga('send', 'pageview');
+  // usage: https://developers.google.com/analytics/devguides/collection/gtagjs/pages
+  window.gtag('event', 'page_view', {
+    page_title: document.title,
```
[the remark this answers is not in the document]
ekoz: Hi, there is some questions:
1. Do we still need to keep `ga.js` ?
2. `docs/plugins.md` is necessary to keep the instructions for `ga` and add the instructions for `gtag` ?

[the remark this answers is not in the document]
ekoz: Great~

src/plugins/ga.js
ekoz: This is not work, I think we can ignore this function for the time being, and then pass the custom collector in through parameters when we need it later.

src/plugins/ga.js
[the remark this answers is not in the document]
ekoz: I just pushed the code, there is an exception in ci/codesandbox. I'm not sure what the reason is, please check it for me.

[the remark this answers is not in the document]
jhildenbiddle: I took a look and tried a few things in #1851 but I was unable to find the issue. This happened once before with Codesandbox it is it kinda of a PITA to troublehoot because the failures don't happen locally or in CI tests.

As much as I'd love to wrap this up, I don't have the spare cycles to devote to it right now. I will circle back as soon as I can devote some time to it.

In the meantime, thanks for being patient with this, @ekoz. Very much appreciated. :smile:

JobaDiniz:
can we merge this one?

sy-records (approved):

PR #1712, opened 2022-01-05
allow the `el` option to accept an `HTMLElement` instance

trusktr:
## **Summary**

The [`el` docs]() say

> It can be a CSS selector string or an actual `HTMLElement`.

which wasn't true. This change updates the code to match the docs.

Additionally this allows Docsify to be embedded in custom elements or components systems like React/Vue/Solid/etc where normally people have a reference to an element.

Bugfix?
Feature?

## **For any code change,**

- [ ] Yes
- [x] No

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE
- [x] jsdom


trusktr:
I realized that `dom.getNode()` has similar logic that I implemented here. Using that would make this PR shorter.

- [ ] use dom.getNode instead

PR #1716, opened 2022-01-07
convert some DOM creation code to Solid JSX

trusktr:
This updates Rollup and replaces Buble with Babel (we're already using Babel anyway) to add babel-preset-solid for compiling Solid JSX expressions.

At the moment this breaks the current SSR feature.

## **Summary**

Converts string-based DOM creation code to Solid JSX expressions. See commit description for more info.

Refactor

## **For any code change,**

- [x] Yes, but the plan is not to merge this until it does not. (Draft PR until then, or maybe never)
- [ ] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


package.json
```diff
@@ -69,19 +69,27 @@
...
+    "@rollup/plugin-replace": "^3.0.1",
     "autoprefixer-stylus": "^1.0.0",
     "axios": "^0.21.1",
     "babel-eslint": "^10.0.3",
-    "babel-jest": "^26.3.0",
+    "babel-jest": "^27.0.0",
+    "babel-preset-solid": "^1.3.1",
     "browser-sync": "^2.26.12",
     "chokidar": "^3.4.2",
+    "colors": "^1.4.0",
```
trusktr, line 92: Pinned because `live-server` depends on libraries using `@latest` which is a good way to get breaking changes into projects, including this one: https://github.com/Marak/colors.js/issues/285

package.json
```diff
@@ -93,9 +101,9 @@
     "eslint-plugin-jest-playwright": "^0.2.1",
     "eslint-plugin-prettier": "^3.1.2",
     "husky": "^3.1.0",
-    "jest": "^26.4.2",
+    "jest": "^27.0.0",
     "jest-image-snapshot": "^4.2.0",
-    "jest-playwright-preset": "^1.4.1",
+    "jest-playwright-preset": "^1.7.0",
```
trusktr, line 106: updating Jest dependencies seems to have somehow causes `jest-image-snapshot` to produce slightly different font rendering, requiring an update to one of our visual snapshots.

I verified locally that both develop and this branch look identical by alternating between tabs in Chrome, and we can also verify this using Vercel's preview.

test/integration/__snapshots__/docs.test.js.snap
```diff
@@ -1,7 +1,7 @@
...
-"<section class=\\"cover show\\" style=\\"background: linear-gradient(to left bottom, hsl(127, 100%, 85%) 0%,hsl(127, 100%, 85%) 100%)\\"><div class=\\"mask\\"></div><div class=\\"cover-main\\"><p><img src=\\"http://127.0.0.1:3001/_media/icon.svg\\" data-origin=\\"_media/icon.svg\\" alt=\\"logo\\"></p><h1 id=\\"docsify-4122\\"><a href=\\"#/?id=docsify-4122\\" data-id=\\"docsify-4122\\" class=\\"anchor\\"><span>docsify <small>4.12.2</small></span></a></h1><blockquote>
+"<section class=\\"cover show\\" style=\\"\\"><div class=\\"mask\\"></div><div class=\\"cover-main\\"><p><img src=\\"http://127.0.0.1:3001/_media/icon.svg\\" data-origin=\\"_media/icon.svg\\" alt=\\"logo\\"></p><h1 id=\\"docsify-4122\\"><a href=\\"#/?id=docsify-4122\\" data-id=\\"docsify-4122\\" class=\\"anchor\\"><span>docsify <small>4.12.2</small></span></a></h1><blockquote>
 <p>A magical documentation site generator.</p></blockquote>
 <ul><li>Simple and lightweight</li><li>No statically built html files</li><li>Multiple themes</li></ul><p><a href=\\"https://github.com/docsifyjs/docsify/\\" target=\\"_blank\\" rel=\\"noopener\\">GitHub</a>
 <a href=\\"#/?id=docsify\\">Getting Started</a></p></div></section>"
```
trusktr, line 7: As noted in some comments in the test specs, this snapshot had to be updated because jsdom presents an empty `style` attribute despite that it is not (bug in jsdom).

Generally we must avoid updating snapshots at all cost, because those can be breaking changes to someone's CSS code. In this case no harm is done because I verified that the style attributes work as intended in a real browser environment.

packages/docsify-server-renderer/index.js
```diff
@@ -41,21 +42,22 @@ function isExternal(url) {
...
+      {config.repo && (
+        <GithubCorner
+          githubUrl={config.repo}
+          cornerExternalLinkTarget={config.cornerExternalLinkTarge}
+        />
+      )}
+      {config.coverpage && <Cover />}
+      <Main {...config} />
+    </>
+  );
```
trusktr, line 60: By converting to Solid components, like this example, we will eventually be able to allow end users to provide their own component in order to override Docsify's, allowing them to make their own Docsify templates.

This will vastly improve Docsify's theme-ability: not only can they override CSS like currently (f.e. docsify-themeable being the most notable example), but they will also be able to replace the DOM structure that wraps the markdown output.

src/core/util/polyfill/css-vars.js
```diff
@@ -26,7 +28,7 @@ export default function(color) {
       }
 
       get(href).then(res => {
-        const style = dom.create('style', res);
+        const style = <style>{res}</style>;
```
trusktr, line 31: 👀

jhildenbiddle:
@trusktr --

Before too much time is spent integrating SolidJS, I think it's worth getting buy-in from @docsifyjs/core. Some maintainers have already expressed concerns and a desire to consider and better understand the pros and cons of SolidJS and alternatives.

My goal is to avoid potential issues down the road if/when this work is completed and you receive pushback instead of appreciation for the efforts you've made. It's the reason why we recommend in our [PR request template](https://raw.githubusercontent.com/docsifyjs/docsify/develop/.github/PULL_REQUEST_TEMPLATE.md) that contributors create an issue, discuss, and wait for approval before creating a PR:

```
```

I also want to make sure smaller, seemingly non-invasive changes aren't used as justification for larger and more significant changes down the road. I'm not saying that is the goal here, but it's easy to imagine how "but we're already using SolidJS" could be used as an argument in future discussions.

As I've mentioned before, I have no issues with SolidJS specifically or the improvements your proposing (framework discussions aside). I just want to avoid conflicts down the road and make sure we're making the best choice for Docsify long-term.

Thx.

Issue #1721, opened 2022-01-12
default behavoir should not be to emojify all, including non-emojis


ChoKaPeek:
## Bug Report

Any couple of colons gets emojified.

#### Steps to reproduce

- have a md containing the following:

`Namespace::SubNameSpace::Class hh:mm:ss :one::two:`

#### What is current behaviour

It will  create a string like this:

> Namespace:<img class="emoji" src="https://github.githubassets.com/images/icons/emoji/SubNameSpace.png alt="SubNameSpace" />:Class hh<img class="emoji" src="https://github.githubassets.com/images/icons/emoji/mm.png alt="mm" />ss <img class="emoji" src="https://github.githubassets.com/images/icons/emoji/unicode/0031-20e3.png?v8" alt="one" /><img class="emoji" src="https://github.githubassets.com/images/icons/emoji/unicode/0032-20e3.png?v8" alt="two" />

#### What is the expected behaviour

This:

> Namespace::SubNameSpace::Class hh:mm:ss <img class="emoji" src="https://github.githubassets.com/images/icons/emoji/unicode/0031-20e3.png?v8" alt="one" /><img class="emoji" src="https://github.githubassets.com/images/icons/emoji/unicode/0032-20e3.png?v8" alt="two" />

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 4.12.2
- Docsify plugins: emoji.js

sy-records:
see https://docsify.js.org/#/configuration?id=noemoji

Issue #1380, opened 2020-10-02
Avoiding emoji transform of colons using __colon__ is undocumented and buggy


atuttle:
## Bug Report
Today I saw in #742 that  supposedly you can use `__colon__` to insert colons that won't be used to transform into an emoji, as when e.g. `:smile:` would become `😄` .

Firstly, if this is intended to be a supported work-around, it should be documented. It is not mentioned at all in the docsify documentation. Secondly, it is not working correctly.

#### Steps to reproduce

```
$ echo '# foo:bar:baz\nfoo bar baz!\n\n\n# foo__colon__bar__colon__baz\nhere we go again' > src/test.md
$ docsify serve src
$ open http://localhost:3000/#/test
```

#### What is current behaviour

https://codesandbox.io/s/wizardly-noyce-44rr6?file=/README.md

Note that the first example displays correctly in the sidebar but incorrectly in the content, and the second example displays incorrectly in the sidebar but correctly in the content. 🤦 

![image](https://user-images.githubusercontent.com/46990/94882419-f9767580-0435-11eb-900a-b2c6cecbd46f.png)

#### What is the expected behaviour

If `__colon__` is the preferred way of avoiding the emoji transform, it should be rendered as a `:` in both the sidebar and the content consistently.

As I was typing up this bug report I thought to try `&#58;` and that does seem to work. You might want to consider adopting that as the preferred approach. I included it in my code sandbox.

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: Mac OSX 10.14.5
- Node.js version: 14.10.1
- npm/yarn version: 6.14.8
- Browser version: N/A
- Docsify version: 4.4.1 (I see more recent versions on github, but none later than 4.4.1 published to npm)
- Docsify plugins: none

#### Please create a reproducible sandbox 

https://codesandbox.io/s/wizardly-noyce-44rr6?file=/README.md

#### Mention the docsify version in which this bug was not present (if any)

n/a only tried latest from npm

sy-records:
First, you need to load the emoji plugin. https://docsify.js.org/#/plugins?id=emoji

atuttle:
That does seem to work, thank you.

Issue #853, opened 2019-05-21
Emoji killing datetime


VagnerDomingues:
#### Steps to reproduce

Just Write a datetime with `:01:` in hourly period anywhere in the documentation.

#### What is current behaviour

An emoji is appearing in the middle of datetime like the image below:

![image](https://user-images.githubusercontent.com/5666881/57528040-9c8d9300-7307-11e9-8760-021f2ca95eb0.png)

#### What is the expected behaviour

People should be able to write hourly periods with seconds in datetime objects representation in their documents, which means writing twice double dots between a number like this...    ` :01: `

#### Other relevant information
There should be an option to disable these emojis.

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: 
- Node.js version: v7.5.0
- npm version: 4.1.2

- Browser version: Chrome  74.0.3729.157 (official) 64 bits
- Docsify version: latest from cdn
- Docsify plugins: docsify-themeable, docsify-pagination

Koooooo-7:
Hi , 
 the bug is a little bit funny :rofl: 。
it seems wrong to parse datetime  as emoji  syntax  `:emoji:`
I think  it should work that   making   datetime  in  the code syntax , add  \` 。
\` 2019-05-06 10:11:12\`

VagnerDomingues:
i can't escape it like that because it'll loose original css styling which is made for ease readings,
i'm using it like this:

![image](https://user-images.githubusercontent.com/5666881/59214952-7b31f800-8b8e-11e9-90e2-4c3258e74081.png)

i just wanted to disable emoji feature because it will never be used in this documentation and it's already on production... cheers!

VagnerDomingues:
please if anyone could help i'd be very thankfull.. 

you can check it on production here:
https://fairlay.com/api/#/public/?id=examples

VagnerDomingues:
so... no way to get rid of emojis?

Mkots:
> so... no way to get rid of emojis?

You can try: 
` window.$docsify = {
        noEmoji: true
      };`

Issue #742, opened 2019-01-12
emoji in comment should not be processed


upupming:
The following md:
```md
So this: <!-- begin-inline-katex
\def\arraystretch{1.5}
   \begin{array}{c:c:c}
   a & b & c \\ \hline
   d & e & f \\
   \hdashline
   g & h & i
\end{array}
end-inline-katex-->
```

will be rendered as:

```html
<p>So this: <!-- begin-inline-katex
\def\arraystretch{1.5}
   \begin{array}{c<img class="emoji" src="https://assets-cdn.github.com/images/icons/emoji/c.png" alt="c" />c}
   a & b & c \\ \hline
   d & e & f \\
   \hdashline
   g & h & i
\end{array}
end-inline-katex--></p>
```

It is expected to leave the content in comments unchanged.

anikethsaha:
Please use `__colon__` instead of colon

```
So this: <!-- begin-inline-katex
\def\arraystretch{1.5}
   \begin{array}{c__colon__c__colon__c}
   a & b & c \\ \hline
   d & e & f \\
   \hdashline
   g & h & i
\end{array}
end-inline-katex-->

```

Hope it helps

sy-records:
![image](https://user-images.githubusercontent.com/33931153/94878274-7aae2680-048f-11eb-869c-b222f2190a63.png)
@atuttle I tested normal, what's wrong with you?

atuttle:
@sy-records I'm not saying that it doesn't work, I'm saying that it's undocumented. If this is the proper way to avoid unintentional conversion to emoji, it should be documented.

PR #1726, opened 2022-01-17
Either enable emoji plugin or not

ChoKaPeek:
## **Summary**

_Disclaimer: All the issues I have referenced were directly affected by the placeholder function. I am not pushing for emoji parsing to be disabled by default, but there are big issues with the actual placeholder / three-way system. It might be nice to add the emoji plugin to the default html_

- Remove the placeholder emojify function dealing more bad than good.
  - _replaces any non-empty couple of colons_
  - _makes the user think emojis are enabled_
  - _duplicates management of the same problem_
  - _one uses the unicode value, the other the word_
  - _pushes the user to use `__colon__` or `&#58;` "quick-and-dirty" fixes_
- Remove the noEmoji configuration option that can unintuitively be set at the same time as the emoji plugin.
- Update documentation.
- Keep replacing `__colon__` to `:` to stay compatible with most quick-and-dirty fixes out there.

This does not seem to need a three-way gate. As long as the plugin is installed, emojis are enabled. Otherwise they aren't. Saves time, avoids confusion and configuration, skips some computation.

If I am not wasting my time, ping me so I can fix tests.

Feature

## **For any code change,**

- [x] Yes
- [ ] No

- install the emoji plugin, if needed, if not already done
- remove the now implicit noEmoji option [optional]

## **Related issue, if any:**
#1721 
#1380 
#853 
#742 
and so on

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -398,20 +398,6 @@ window.$docsify = {
 
 Note that if you are running an external script, e.g. an embedded jsfiddle demo, make sure to include the [external-script](plugins.md?id=external-script) plugin.
 
-## noEmoji
-
-- type: `Boolean`
-
-Disabled emoji parse.
-
-```js
-window.$docsify = {
-  noEmoji: true,
-};
-```
-
-?> If this option is `false` but you don't want to emojify some specific colons, [refer to this](https://github.com/docsifyjs/docsify/issues/742#issuecomment-586313143)
-
 ## mergeNavbar
 
 - type: `Boolean`
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -91,13 +91,18 @@ Configure by `data-ga`.
 
 ## emoji
 
-The default is to support parsing emoji. For example `:100:` will be parsed to :100:. But it is not precise because there is no matching non-emoji string. If you need to correctly parse the emoji string, you need install this plugin.
+Emoji parsing is disabled by default.
+To enable emoji parsing, you need to add the following plugin.
 
 ```html
 <script src="//cdn.jsdelivr.net/npm/docsify/lib/plugins/emoji.min.js"></script>
 ```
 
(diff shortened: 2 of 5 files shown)


trusktr:
Hello @ChoKaPeek ! Thanks for submitting this. I will circle back soon.

trusktr (commented):
Personally I like this change, but it is a breaking change. This would require us to:

- bump to v5
- create the new migration log we've been chatting about, and describe the step to migrate from the previous version to the new one
- there are a high number of people not using version numbers in script tags that load Docsify from script tag. We need to either
  - keep the v4 docsify{.min}.js in the same place, and output a new v5 script to a new location in the output folder, that way people can opt into the new version
  - or just let people's emojies break, and when they ask, we point them to migration log.

Ultimately, for health of the project, I think breaking emoji's is better for the long term, but I can understand if we go with the non-breaking approach.

Related is PR https://github.com/docsifyjs/docsify/pull/1733 (prevent people from starting Docsify sites with version-less links)

trusktr:
Here's a shorter way to put it:

I think people making stable sites by default (stable as in it works exactly how they tested it) is better as a default. People can use `@X` instead of `@X.X.X` is they wish to opt into automatic updates. But IMO better if people update their versions when they feel is a good time rather than us doing it.

I've never written a web app where it automatically updates itself and publishes itself with new dependencies. In my opinion, that is a recipe for periodical breakage, and more work (from years of experience).

jhildenbiddle:
Agreed that both the use of the `emoji.js` plugin and the emoji parsing code need some attention.

See #779 for proposals on rending emoji and the the use of the emoji plugin moving forward. Much of what you've outlined here is covered there.

In the meantime, there are simple changes that can be made to the emoji parser to address many of the issues you've linked to here. I'm going to take look at what non-breaking changes can be made short-term, then I'll circle back.

jhildenbiddle:
These issues have been addressed as a non-breaking change with #1746.

As for version locks in URLs, this is a completely separate topic that has been covered at great length elsewhere.

PR #1225, opened 2020-06-16
fix: Updated docs with instructions for installing specific version #780

aemmadi:
**Summary**
Updated docs with example of using a specific version of docsify while getting started, instead of the default version. 

Issue #780 requested changes in the CLI, opened a PR there as well with the changes [#107](https://github.com/docsifyjs/docsify-cli/pull/107)

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

**Other information:**

---

* [x] DO NOT include files inside `lib` directory.

Issue #780, opened 2019-02-21
Specify major version number in URLs


jhildenbiddle:
The HTML rendered by [docsify-cli](https://github.com/docsifyjs/docsify-cli#readme) and listed in the [Manual Installation](https://docsify.js.org/#/quickstart?id=manual-initialization) section of the official docs results in the very latest version of docsify being loaded by default. This is not only bad practice generally speaking, but it also means that breaking changes introduced in a new major release could break *a lot* of live docsify sites.

The suggested fix is to specify a major version number in all docsify URLs:

```html
<link rel="stylesheet" href="//unpkg.com/docsify@4/themes/vue.css">
<script src="//unpkg.com/docsify@4/lib/docsify.min.js"></script>
```

This means that when a new *major* version is released, site admins will need to manually update the URLs with the new major version number (e.g. `@5`). The benefit of this approach is that breaking changes in the new release will not affect existing docsify sites since they are loading the previous version.

If URLs are being updated, it's also worth considering the following:

* Use CDN redirect URLs instead of specifying a path for the main lib file
* Specify the URL scheme to avoid unnecessary redirects (i.e use `https://` instead of `//`)
* Switch to jsDelivr (#756) for better CDN performance

```html
<script src="https://cdn.jsdelivr.net/npm/docsify@4/themes/vue.css"></script>
<script src="https://cdn.jsdelivr.net/npm/docsify@4"></script>
```

timaschew:
So this issue should be then migrated (or at least linked) to the CLI project, right?

jhildenbiddle:
It's a valid issue for this repo since the [Manual Installation](https://docsify.js.org/#/quickstart?id=manual-initialization) section of the docs needs to be updated, but yes--work will also need to be done in the CLI repo.

I've made a similar issue in the CLI repo.

trusktr:
Shouldn't this be in the 4.x project board?

jhildenbiddle:
Yep. Done. 😄

PR #1733, opened 2022-01-24
Prevent people from importing Docsify without versions in URLs.

trusktr:
## **Summary**

Every new person that starts using Docsify copies the examples, which have no versions in CDN URLs, and makes websites that can easily break. This is not good, and instead we should encourage everyone to use versions by example.

This adds versions to all CDN link examples to prevent people from making websites that are at risk of breaking changes to Docsify, makes it clear that the recommended approach is to use full versions in URLs, and fixes formatting in any files that were modified.

  Docs

- [ ] Yes
- [x] No


docs/index.html
```diff
@@ -20,24 +20,24 @@
     />
     <link
       rel="stylesheet"
-      href="//cdn.jsdelivr.net/npm/docsify@4/lib/themes/vue.css"
+      href="//cdn.jsdelivr.net/npm/docsify@4.12.2/lib/themes/vue.css"
```
sy-records, line 23: Why not `@4`?

Koooooo-7 (approved):
LGTM.
But It seems we should update `release.sh` as well, otherwise, there should hv change the version manually in docs at each new release.

trusktr:
> LGTM. But It seems we should update `release.sh` as well, otherwise, there should hv change the version manually in docs at each new release.

I thought the same. I think it can be a new PR. How should we do it?

1. If we place `{{docsifyVersion}}` (or something) everywhere, we can find/replace it during publish for the website. The downside is that when viewing the docs on GitHub, version numbers will not be there.
2. Another idea is we could have `.docs/` folder, and a build step to output to `docs` folder. People who edit docs would edit files in `.docs/`, but people who read docs would view `docs/`. Downside is double content in the repo.

Any other idea?

I like option 1 better so far. People can just see the website for docs.

docs/index.html
trusktr: I think @Koooooo-7's idea below, to replace the version automatically, would be great. Let's go with that approach because then it will behave like `@4` for us (for Docsify site), but users will see the full version and will copy/paste that into their projects which is good.

jhildenbiddle:
I do not think this is a good idea, and there is a precedence for not doing this.

Which is a better / more likely scenario:

1) We ship multiple versions of docsify with zero bugs, which users are locked into to save them from *future* bugs.
2) We ship multiple versions of docsify with one-or-more bugs, for which users can receive fixes with zero work and without breaking their site.

Since there is a 100% chance we will continue to ship versions of Docsify with bugs, I think the second option is better for everyone. Having the ability to deliver new minor versions (with non-breaking bug fixes and enhancements) and patch versions (with non-breaking bug fixes only) is important. This is the whole purpose of [semantic versioning](https://semver.org). Beginners should *prefer* major version locks for exactly this reason. Had we done this years ago, think of all of the sites that would still be stuck on less-capable, less-stable versions of docsify v4.

The template in the docs and the output of `docsify-cli` already include `@4` version locks:

```html
<link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify@4/themes/vue.css" />
```

```html
<script src="//cdn.jsdelivr.net/npm/docsify@4"></script>
```

I vote we leave it this way. If we want to help beginning understand how CDN version locks work, we can modify the docs accordingly. Here is a screenshot of the block of text I've added to some of my projects available on CDNs:

---

CleanShot 2022-01-28 at 16 54 31@2x

---

Not saying wave to use that exact text, but something similar would suffice.

jhildenbiddle (changes requested):
See comment: https://github.com/docsifyjs/docsify/pull/1733#issuecomment-1024725618

trusktr:
By encouraging full version numbers, when someone encounters that bug situation, they can

- implement an alternative or workaround, staying on the same version, and the site _stays in a known condition_
- or they can _manually_ update the version number to receive a fix, updating to a new _known condition_

Like you said, there will always be bugs (all software has bugs), so automatically fixing bugs does not mean we won't introduce new ones. People are going to get annoyed when things randomly break (of course I hope it will happen as little as possible).

The worst thing (from experience) is in-range breaking changes. Those are annoying, and those situations happen too often even though people try hard to breaking releases. What do we do in those cases to avoid the most work? Pin the version. :smiley:  Might as well do it from the beginning for an important project that wishes to _remain in a known working condition_.

From another angle, in theory, if there are no breaking changes, then a developer can also update the version once in a while. Better that they do that, perform some testing, then release, in my opinion.

---

Whatever the case, we need to eliminate the fact that too many people use Docsify with no version numbers in their URLs.

So the alternative to these changes is to update all the URLs to having `@x` instead of `@x.x.x`.

@docsifyjs/core we need votes.

jhildenbiddle:
@truskr --

> So the alternative to these changes is to update all the URLs to having @x instead of @x.x.x.

This is what we have today. Major version locks (`@4`) have been in place for ~2.5 years (see #780, #1225).

> Whatever the case, we need to eliminate the fact that too many people use Docsify with no version numbers in their URLs.

How are you determining that "too many people use Docsify with no version number"?

Here are the weekly stats from [jsdelivr](https://www.jsdelivr.com/package/npm/docsify):

CleanShot 2022-01-30 at 16 39 55@2x


We have no way of knowing how those sites are loading the latest version of docsify because all of the following URLs will resolve to the latest version of docsify:

* https://cdn.jsdelivr.net/npm/docsify
* https://cdn.jsdelivr.net/npm/docsify@4
* https://cdn.jsdelivr.net/npm/docsify@4.12
* https://cdn.jsdelivr.net/npm/docsify@4.12.2
* https://cdn.jsdelivr.net/npm/docsify@latest



Unfortunately, jsDelivr only provides stat for specific versions like `@4.12.2`. It does not provide stats for redirects.

Based on those numbers, 97% people visiting sites that load docsify via jselivr are receiving the latest version (4.12.2). If we include the people that it *appears* have manually locked themselves to the last version (4.12.1) but have yet to update their URLs, that percentage jumps to 99%. I'd say this this is great! It means most site visitors are using the latest-and-greatest version we have which include all of the non-breaking enhancements and bug fixes we've added up until today.

sy-records (changes requested):
I agree with @jhildenbiddle

jhildenbiddle:
@trusktr --

> when someone encounters that bug situation...

This is the critical difference between locking to a major version vs. a specific version: *when* a site owner encounters a bug, which also means *if* a site owner encounters a bug.

Locking to a major version has risks and benefits. Doing so accepts bugs may exist that affect site visitors and it's optimistic about Docsify's ability to make fixes available responsibly. You run the risk of your site going down if a bug or unintentional breaking change is published, but you'll get fixes and enhancements automatically without having to manually check for updates or test on a variety of environments (browsers, operating systems, screen sizes, etc). If the possibility of downtime isn't a deal-breaker and you want your docsify site to remain up-to-date without any effort, major version locks work great.

Locking to a specific version also has risks and benefits. It locks your site into a known state -- for better or worse -- that you as the site owner have determined is good enough to meet your needs at that time. It is pessimistic about Docsify's ability to make fixes available responsibly, which is the right approach if the possibility of downtime is a deal-breaker for you. The risk is that there may be bugs that affect site visitors that you are unaware of, and the way to mitigate that risk is by spending time manually checking change logs and testing in all environments to make sure your site works as expected.

Which option a site owner chooses is a balance between things like requirements, convenience, testing ability, and confidence in the ecosystem's authors/contributors. For better or worse, we decided a major version lock as the default lock type with a [message](https://docsify.js.org/#/quickstart?id=specifying-docsify-versions) explaining the benefits of locking to a specific version with instructions for doing so was the right way to go (#780, #1225, https://github.com/docsifyjs/docsify/pull/1225#issuecomment-647159979, https://github.com/docsifyjs/docsify/pull/1225#issuecomment-647197742). We can't revisit this decision, but this feels like a topic that has been discussed and resolved already, and it is unclear why it is resurfacing now.

There's room for creative solutions here though. Both the website and the CLI can be updated to ask users to choose between the convenience of major version locks and the stability of specific version locks, which would update the template accordingly. We could extend this to also ask users which plugins they want to install and include those CDN URLs as well (also adhering to the version lock type they choose). This would be much more helpful for end users instead of our current system which requires them to copy/paste a template and/or CDN URLs and manually modify markup if they prefer a version lock type that differs from the one we use by default.

Can we close this PR and move on to discussing other solutions like the one mentioned above? I believe we already have two maintainers in favor of doing so (myself and @sy-records).

@docsifyjs/core -- Thoughts?

sy-records:
The document has already been modified in v5.

Issue #1209, opened 2020-06-10
Wrapper function for creating plugins


anikethsaha:
Another v5 feature we should consider is tweaking the docisfy plugin architecture to be more resilient. This warrants further discussion, but here are two ideas worth considering:

1. Provide a docisfy plugin "starter" or a sample repo to use as a starting point.

   This would make life much easier for the community and help enforce acceptance criters (below).

1. Define acceptance criteria for all docsify plugins.

   This is critical for docsify because unlike static sites, a docsify-powered site requires Javascript to render content. If a docsify plugin throws an error, it is almost certain that site functionality will be broken. 

   Here are a few idea for acceptance criteria:

   - Specify the version of ECMAScript plugin developers should target. For docsify v4.x, ES5 is the required target to ensure IE11 support. This means dev must either author their plugin with ES5 syntax or transpile their code from ES6+ to ES5. For docsify v5, we'll have to look at our supported browser list to determine the appropriate ECMAScript version to target.
   - Plugin code must be wrapped in a `try { ... } catch(err) { ...}` statement to prevent a plugin error from breaking a site. 
   - Plugins must make their version number accessible to other plugins for interoperability. Docsify addons like [docsify-themeable](https://jhildenbiddle.github.io/docsify-themeable/) and [docsify-tabs](https://jhildenbiddle.github.io/docsify-tabs/) already do this by adding an object to the global `$doscify` object (`$docsify.tabs` and `$docsify.themeable`) with a `value` key and value. The goal is to allow both docsify and other plugins to determine which plugins are active and, if necessary, modify their behavior as needed.
   - Plugins must specify a compatible docsify version to prevent breaking sites when a new version of docsify is released. For example, consider a v4 site that is configured to load the latest version of docsify and multiple plugins. If docsify v5 is released using the same URLs, the v4 site will automatically be "upgraded" to v5 but will likely break as a result of trying to load v4 plugins. This safeguard would allow docsify (or the plugin itself) to determine if it is safe to run the plugin before doing so, which would at least allow the site to load and function properly (albeit without incompatible plugin functionality). This wouldn't be necessary if users locked their docsify, theme, and plugin CDN URLs to a major version (see #780), but since they don't a safeguard like this is necessary.

Just some ideas to get started.

_Originally posted by @jhildenbiddle in https://github.com/docsifyjs/docsify/issues/1061#issuecomment-641731543_

anikethsaha:
[...]

PR #1742, opened 2022-02-01
feat: Plugin error handling

### LINKED ISSUE #1741: Handle uncaught plugin errors to prevent live sites from breaking
## Feature request

Docsify should handle uncaught plugin errors to prevent JavaScript execution from halting and breaking docsify sites.

Docsify sites are rendered on the client using JavaScript. When a plugin error occurs, JavaScript execution is halted. The result for site visitors is a "broken" site typically rending only a "Loading..." message or an unresponsive UI (depending on when the plugin was invoked and what type of error occurs).

The problem with uncaught plugin errors is that they often go unnoticed by site owners and plugin authors and end up breaking live sites. For example, here are just a few uncaught plugin error issues that I have personally worked on:

- https://github.com/docsifyjs/docsify/issues/390
- https://github.com/docsifyjs/docsify/issues/686
- https://github.com/francoischalifour/medium-zoom/issues/34
- https://github.com/waruqi/docsify-plugin-carbon/issues/1
- https://github.com/imyelo/docsify-pagination/issues/5
- https://github.com/jperasmus/docsify-copy-code/issues/1
- https://github.com/Plugin-contrib/docsify-plugin/issues/16

Every issue above was the result of a plugin author using an unsupported method or ES6+ syntax, and each one prevented docsify sites from rendering properly in IE10/11. These bugs went undetected until I started testing [docsify-themeable](https://jhildenbiddle.github.io/docsify-themeable/) with various plugins in all supported platforms (which at the time included IE10 as well as IE11). If Docsify handled those uncaught plugin errors, the sites would have rendered properly (albeit without the expected plugin functionality).

The examples above are all "legacy" browser related, but this is not a legacy browser issue. The same issue exists for evergreen browsers since browser vendors choose if and when they will adopt new ES202X methods and syntax changes. For example, consider ES2022's [Array.prototype.at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) feature which is currently [supported by all evergreen browsers except Safari](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility):

```js
window.$docsify = {
  // ...
  plugins: [
    function (hook, vm) {
      hook.beforeEach(function (content) {
        console.log('Plugin 1');
        console.log(['fail', 'success'].at(1));
      });
    },
    function (hook, vm) {
      hook.beforeEach(function (content) {
        console.log('Plugin 2');
        console.log(['fail', 'success'].at(1));
      });
    }
  ]
}
```

Chrome 99:

CleanShot 2022-03-09 at 15 09 57@2x

Safari 15.3 (latest):

CleanShot 2022-03-09 at 15 10 06@2x

This seems like an obvious error that one might expect to be caught by either the site owner or the plugin author, but it is not that simple:

- What if the site owner and/or plugin author does not have the knowledge or resources to test on all supported platforms?
- What if the plugin error is caused by a conflict with another plugin?
- What if the plugin error occurs only when a specific configuration option is set?
- What if the plugin error occurs only when an infrequent event occurs (e.g., malformed markdown or HTML, local storage limits exceeded, interaction on mobile device in landscape mode, etc.)?

These are not theoretical examples. We have real-world examples of each of them: #1057, #1392, #1526, #1538, ... etc. Obvious bugs are easy to detect, but not all bugs are obvious.

Ideally, plugin authors would conscientiously handle their own errors and thoroughly test on all platforms supported by Docsify. We know from years of docsify plugin development that this doesn't happen. The best we can do to prevent these issues is to make Docsify more resilient when possible and provide some best practiceson our [Write a Plugin](http://localhost:3002/docs/#/write-a-plugin) section.

**Note:** _Asynchronous_ code will require plugin authors to implement their own `try/catch/finally` blocks directly in their code. An update to the official documentation that explains this would be helpful.

```js
hook.afterEach(function (html, next) {
  try {
    // Async tasks...
    console.log("Plugin 1");
    console.log(["fail", "success"].at(1));
  } catch (err) {
    // ...
  } finally {
    next();
  }
});
```

A `catchPluginErrors` configuration option can be added to allow uncaught plugin errors to throw as usual. This may be useful for docsify contributors, plugin authors, and site owners who rely on uncaught errors for their development or site monitoring tools.

- Type: `boolean`
- Default: `true`

```js
window.$docsify = {
  catchPluginErrors: true // default
}
```

As a non-breaking change with documentation updated accordingly.

Yes

jhildenbiddle:
## **Summary**

Fix #1741 (see for details)

- Add new `catchPluginErrors` configuration option to enable/disable handling uncaught plugin errors (default: `true`). This feature prevents uncaught plugin errors from halting JavaScript execution and breaking Docsify sites.
- Add e2e tests for `catchPluginErrors` configuration option
- Add e2e tests for all plugin hooks (data handling and invocation order)
- Update `Write a Plugin` new new Setup, Template, Lifecycle Hooks, and Tips sections (see Vercel preview below)

## Screenshots

Assuming the following plugin code:

```js
window.$docsify = {
  // ...
  plugins: [
    function (hook, vm) {
      hook.beforeEach(function (content) {
        console.log('Plugin 1');
        console.log(['fail', 'success'].at(1)); // <= ERROR: at() not supported by Safari
      });
    },
    function (hook, vm) {
      hook.beforeEach(function (content) {
        console.log('Plugin 2');
        console.log(['fail', 'success'][1]);
      });
    }
  ]
}
```

### Before

Safari 15.3: Uncaught "Plugin 1" error halts JavaScript execution, breaking docsify site.

CleanShot 2022-03-14 at 15 33 35@2x

### After

Safari 15.3 w/ `catchPluginErrors: true` (default): Uncaught "Plugin 1" error is caught by docsify and logged to the console, allowing Docsify to continue with "Plugin 2" and rendering site content.

CleanShot 2022-03-14 at 15 33 01@2x

Note that error details are provided in the console as they were before. Site owners and plugin authors that use the console to troubleshoot and debug can continue to do so. Safari collapses these details by default. Chrome does not.

CleanShot 2022-03-14 at 15 33 21@2x

Safari 15.3 w/ `catchPluginErrors: false`: Same as "Before".

CleanShot 2022-03-14 at 15 33 35@2x

## **For any code change,**

- [ ] Yes
- [x] No

## **Related issue, if any:**

#1741
#1209

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE


trusktr (commented):
Is the error navigable the same way as an uncaught error, in console?

Now that errors are caught, they will not be pausable by default in debugger, and pause on caught errors option generally pauses too much that it can become inconvenient.

Personally I would prefer to see my actual errors in console, and in the console I can clearly see which plugin I wrote bad code in. 

The error stack trace if an error shown in console tells us exactly where the error came from (hence we already know if it in some plugin)

jhildenbiddle:
@trusktr --

Details available in #1741...

> Is the error navigable the same way as an uncaught error, in console?

Yes.

> Now that errors are caught, they will not be pausable by default in debugger, and the pause on caught errors option generally pauses too much that it can become inconvenient.
> 
> Personally I would prefer to see my actual errors in console, and in the console I can clearly see which plugin I wrote bad code in.

Set `catchPluginErrors` to `false` during plugin development if you wish to use a debugger as you do today.

> The error stack trace of an error that is shown in console tells us exactly where the error came from (hence we already know if it comes from a plugin)

Still there. See screenshots in #1741.

Koooooo-7 (commented):
should we console all hooks directly or we should have a switch to open this console output? i.e.
`debugEnable: true.`

jhildenbiddle:
@Koooooo-7 --

> should we console all hooks directly or we should have a switch to open this console output? i.e.`debugEnable: true.`

I'm not sure I understand the question. 

It sounds like you're asking if we should add a new configuration option to either A) prevent uncaught plugin errors from being displayed in the console or B) open the console automatically when an error occurs (presumably to make people aware that the error occurred). 

I don't think we should consider either option.

For option "A": Errors should continue to be sent to the console as they are today because the console is where developers and end-users expect to find error-related details. This PR does not change that, and that is by design. Developers and end-users that use the console to debug and/or troubleshoot can continue to do so exactly as they do today. Site visitors that don't know or care about the console are free to ignore it exactly as they do today.

It's worth mentioning that if a plugin author or end user *really* wanted to suppress error messages in the console, they can do so already by implementing their own `try/catch` block(s) and/or using the console message filters provided by the browser:

**Chrome 99**

CleanShot 2022-03-14 at 13 08 49@2x

**Safari 15.3**

CleanShot 2022-03-14 at 13 07 03@2x

For option "B": I don't think it is possible to programmatically open the browser console, but even if was we should not assume that every site visitor 4) needs to be notified that a plugin error has occurred, 2) knows what the console is, 3) will understand why the console is being displayed, or 4) knows how to dismiss the console. 

All that being said, your question does hint at two separate-but-related topics:

1. How will people know when an error has occurred if the console isn't open?
2. What should happen when an error occurs?

These are valid questions, but I believe they are outside the scope of this PR.  If we want to address them, let's create a new issue where we can discuss the details. I have some ideas regarding a new error-related hook that would address both of these conners and provide some interesting plugin opportunities.

Koooooo-7 (approved):
LGTM.

sy-records (approved):
LGTM.

Thank you for your contribution.

jhildenbiddle:
Thanks @Koooooo-7 and @sy-records for the review! 🥳

PR #1743, opened 2022-02-03
fix: Legacy bugs (styles, site plugin error, and dev server error)

jhildenbiddle:
## **Summary**

Fix a collection of legacy browsers bugs.

**NOTE:** Obviously, legacy browsers are not a huge priority but the fact remains that Docsify v4.x does officially support legacy browsers like IE11. These non-breaking fixes address issues I bump into most often when testing IE11 from time to time.

* Fix CSS styles in legacy browsers without support for CSS custom properties
* Fix presentation of search field
* Fix an error on the website caused by linking to the non-transpired version of `docsify-plugin-carbon`
* Fix dev server error in IE (BrowserSync snippet not compatible / now loaded conditionally)
* Refactor `/docs` site conditional JS loading (simplification)

**Additional changes**

* Fix ESLint errors found in `/docs/index.html`
* Fix `.gitignore` paths that blocked committing changes to `/src/themes`

**Screenshots - Before**

CleanShot 2022-02-02 at 23 48 53@2x

CleanShot 2022-02-02 at 23 49 53@2x

**Screenshots - After**

CleanShot 2022-02-02 at 23 43 39@2x

CleanShot 2022-02-03 at 11 00 01@2x

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [x] IE


Koooooo-7 (dismissed):

sy-records:
There seems to be some problems here.

image

jhildenbiddle:
@sy-records 

> There seems to be some problems here.

Fixed.

FYI, the issue you highlighted in your screenshot wasn't introduced with this PR. The search field has always been like that in IE11. It was easy enough to fix, so we may as well while we're addressing (what will hopefully be the last) IE issues.

src/core/init/lifecycle.js
```diff
@@ -36,14 +36,27 @@ export function Lifecycle(Base) {
...
-            hookFn(data, result => {
-              data = result;
-              step(index + 1);
-            });
+            try {
+              hookFn(data, result => {
+                data = result;
+              });
+            } catch (err) {
+              console.error(errTitle, err);
```
trusktr, line 47: Can we keep this in the separate PR?

trusktr (changes requested):
Introduces the same change as in https://github.com/docsifyjs/docsify/pull/1742. Needs to be split? (or maybe this one should merge onto that one)

src/core/init/lifecycle.js
jhildenbiddle: Yes. This change should not have been merged into this PR. Sorry about that. Good catch. I'll remove it.

jhildenbiddle: @trusktr -- Reverted. Thanks again for the catch.

docs/index.html
```diff
@@ -55,7 +55,7 @@
 
   <body>
     <div id="app">Loading ...</div>
-    <script src="//cdn.jsdelivr.net/npm/docsify-plugin-carbon@1/index.js"></script>
+    <script src="//cdn.jsdelivr.net/npm/docsify-plugin-carbon@1"></script>
```
sy-records, line 58: Do you need to update index.html?

https://github.com/docsifyjs/docsify/blob/533193fd2d27f7edae051ea4de80514c09545248/index.html#L26

jhildenbiddle: Yep. :)

Done.

sy-records (approved):
LGTM.

Koooooo-7 (approved):
LGTM.

jhildenbiddle:
Thanks @sy-records and @Koooooo-7! 🙏

Issue #779, opened 2019-02-21
Render unicode emoji with image fallback


jhildenbiddle:
(Moving discussion of rendering unicode emoji from #768 to a separate thread)

From #768:

> According to [caniemoji.com](http://caniemoji.com), unicode emoji are supported all the way back to Windows 7, macOS 10.7, Android 4.4, and iOS 5. Switching to unicode emoji would remove emoji `<img>` requests and render emoji native to each platform. Seems like a better route to go, so long as the unsupported OS+Browser combinations aren't a concern. [This article](https://about.gitlab.com/2018/05/30/journey-in-native-unicode-emoji/)  will likely be helpful if/when the switch to unicode happens.

timaschew:
What is your idea to support GitHub based aliases? I don't see any unique short names on this page: https://unicode.org/emoji/charts/full-emoji-list.html 
On another I saw some [other project](https://github.com/amio/emoji.json) which use the same source, but there are ambiguous keywords which are assigned to multiple emojis sometimes.

jhildenbiddle:
My initial thought was to just leverage the [GitHub emoji API](https://api.github.com/emojis) to get both the alias and the unicode value:

```javascript
// API
{
  "+1": "https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png?v8",
  ...
  "thumbsup": "https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png?v8",
  ...
}

// Alias  : "+1" or "thumbsup"
// Unicode: "1f44d"
// URL    : "https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png?v8"
```

GitHub has already done the work of mapping aliases to unicode values and images (the file name is the unicode value), as well as adding logical duplicates like the example above.

As for the implementation:

- Emoji data should be embedded in docsify.js, not requested from the GitHub API by the client.
- GitHub data should be transformed to better fit docsify's usage and reduce client-side parsing. For example, instead of embedding GitHub's API data as-is:

  ```javascript
  {
    "+1": "https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png?v8",
    ...
  }

  // const emojiURL  = emoji["+1"];
  // const emojiCode = emojiURL.slice(emojiURL.lastIndexOf('/') + 1, emojiURL.lastIndexOf('.'));

  ```

  The data should be transformed to allow fast unicode and URL lookups:

  ```javascript
  {
    "+1": {
      unicode: "1f44d",
[...]

PR #1745, opened 2022-02-03
chore: sync emojis

sy-records:
## **Summary**

Sync GitHub emojis.

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


Koooooo-7 (approved):

trusktr (approved):
is there another way to do this? Perhaps get them from an API?

jhildenbiddle:
@sy-records -- 

Looks good, but I think it's worth revisiting #779 instead of continuing to manually update this list. No issue with merging this until we resolve #779.

@trusktr --

I've reopened #779 because the PR you reference as addressing the issue are unrelated.

jhildenbiddle (approved):

sy-records:
I wrote a code using PHP, not a manual update, and I will add a CI later.

jhildenbiddle:
@sy-records --

I think we should avoid making PHP a dependency for our project. This seems like a very easy thing to do in JS/Node, no?

sy-records:
It will not become dependent, just a CI.

I will try to use js or shell for this.

PR #1746, opened 2022-02-04
feat: Native emoji w/ image-based fallbacks and improved parsing

### LINKED ISSUE #742: emoji in comment should not be processed

### LINKED ISSUE #829: Emoji is killing datetime
emoji :01:

emoji :01: is breaking formatted datetime :cry:  i couldn't find an option to ignore them.

### LINKED ISSUE #853: Emoji killing datetime

### LINKED ISSUE #1009: emojis in titles not working correctly
## Bug Report

Processing the emojis in titles isn't working correctly. Sometimes just turning it into a single ":" character. Seems to work correctly if no leading space before the emoji code.

#### Steps to reproduce

Insert this content in a page
```
#### :arrow_forward:  :arrow_forward:
#### :arrow_forward::arrow_forward:
#### &nbsp;:arrow_forward: :arrow_forward:
#### &nbsp;:arrow_forward::arrow_forward:
#### &nbsp;:arrow_forward:&nbsp;:arrow_forward:
```

image

All emojis should appear

- Your OS: mac
- Node.js version: 11.15
- npm/yarn version: 6.7
- Browser version: chrome latest
- Docsify version: cli is 4.4, using latest from cdn
- Docsify plugins: all disabled

### LINKED ISSUE #1721: default behavoir should not be to emojify all, including non-emojis

jhildenbiddle:
## **Summary**

Add native emoji support and fix docsify core's built-in emoji parser. See related issues below for details.

1. Fix multiple issues related to Docsify's built-in emoji parser
2. Resolve confusion about the need for a separate emoji plugin
3. Add option to render emoji `:shorthand:` codes using platform-native emoji characters or images
4. Automate the process of updating the emoji data uses to render emoji `:shorthand:` codes
5. Add tests for the built-in emoji parser

- [x] Feature: Add native emoji support
- [x] Feature: Add `nativeEmoji` configuration option (`true` = native, `false` = image-based)
- [x] Feature: Merge emoji plugin data and shorthand verification for better parsing into core
- [x] Feature: Add deprecation notice to emoji plugin (no longer used internally by docsify)
- [x] Fix: Incorrect parsing of unmatched emoji shorthand codes w/o separate emoji plugin
- [x] Fix: Incorrect parsing of emoji shorthand code in HTML comments
- [x] Fix: Incorrect parsing of emoji shorthand code in `<script>` tags
- [x] Docs: Add `nativeEmoji` configuration option details
- [x] Docs: Add deprecation notice and explanation for emoji plugin
- [x] Docs: Remove link to discussion of `__colon__` (confusing and no longer necessary). 
- [x] Docs: Add recommendation to use the `&colon;` HTML entity to render emoji shortcakes as text
- [x] Build: Add emoji-related tests
- [x] Build: Auto-generate emoji data from GitHub API
- [x] Build: Update codesandbox configurations to use same version of node 
- [x] Build/Docs: Auto-generate emoji markdown for website
- [x] Other: Visually test on major platforms to verify proper rendering of emoji

To update emoji data (`emojify-data.js`) and markdown demo on website (`emoji.md`):

```
npm run build:emoji
```

---

### Screenshots - Native vs. Image Emoji

Markdown:

```markdown
:smile:
:poop:
:100:
:+1:
:1st_place_medal:
:fire:
```

Image Emoji (`nativeEmoji:false`)

CleanShot 2022-02-05 at 02 05 03@2x

HTML:

```html
<img src="https://github.githubassets.com/images/icons/emoji/unicode/1f604.png?v8.png" alt="smile" class="emoji" loading="lazy">
<img src="https://github.githubassets.com/images/icons/emoji/unicode/1f4a9.png?v8.png" alt="poop" class="emoji" loading="lazy">
<img src="https://github.githubassets.com/images/icons/emoji/unicode/1f4af.png?v8.png" alt="100" class="emoji" loading="lazy">
<img src="https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png?v8.png" alt="+1" class="emoji" loading="lazy">
<img src="https://github.githubassets.com/images/icons/emoji/unicode/1f947.png?v8.png" alt="1st_place_medal" class="emoji" loading="lazy">
<img src="https://github.githubassets.com/images/icons/emoji/unicode/1f525.png?v8.png" alt="fire" class="emoji" loading="lazy">
```

Native Emoji (`nativeEmoji:true`, macOS emojis shown)

CleanShot 2022-02-05 at 02 04 45@2x

HTML:

```html
<span class="emoji">😄︎</span>
<span class="emoji">💩︎</span>
<span class="emoji">💯︎</span>
<span class="emoji">👍︎</span>
<span class="emoji">🥇︎</span>
<span class="emoji">🔥︎</span>
```

### Screenshots - Emoji Parsing

Markdown:

```markdown
hh:mm

hh:mm:

hh:mm:ss

:mm:s

Namespace::SubNameSpace

Namespace::SubNameSpace::Class

2014-12-29T16:11:20+00:00

2020-12-03T15:05:57+00:00 (extended)

2020-12-03T15:05:57Z (extended)

:01:
```

Before:

CleanShot 2022-02-04 at 01 11 40@2x

After: 

CleanShot 2022-02-04 at 01 39 48@2x

### HTML Comments

Markdown:

```markdown
<!-- :smile: -->
```

Output before:

```markdown
<!-- <img class="emoji" src="https://github.githubassets.com/images/icons/emoji/smile.png" alt="smile"> -->
```

Output after:

```markdown
<!-- :smile: -->
```

### Script Tags

Markdown:

```markdown
<script>
  var test = ':smile:';
</script>
```

Output before:

```markdown
<script>
  var test = '<img class="emoji" src="https://github.githubassets.com/images/icons/emoji/smile.png" alt="smile">';
</script>
```

Output after:

```markdown
<script>
  var test = ':smile:';
</script>
```

### Screenshot - Auto-Generated "Emoji"page

CleanShot 2022-02-05 at 00 30 22@2x

## **For any code change,**

- [ ] Yes
- [x] No

* Emoji are rendered as they are today by default (using images via GitHub's CDN)
* Improved emoji parsing comes by integrating the emoji data and shorthand verification logic used in the Docsify's emoji plugin. The core parsing functionality is the same with this change as it is for sites using the emoji plugin. The only enhancements are preventing emoji parsing within HTML comments and `<script>` tags.
* Deprecating the emoji plugin has the added benefit of removing confusion surrounding the emoji plugin and its purpose. Since Docsify supports some-but-not-all emoji without the plugin, it was difficult for users to determine if they needed to use it.
* For native emoji, users can opt-in using `nativeEmoji:true` (currently `false` by default). When the next major release is ready, maintainers can determine if native emoji should be set as the default option.

## **Related issue, if any:**

#779 

Also:

Fix #742
Fix #829
Fix #853
Fix #1009
Fix #1721
Fix #1726

... and probably others. Emoji parsing has been an issue for various reasons and for quite some time.

- [x] Android (Chrome)
- [x] iOS (Safari)
- [ ] Linux (Chrome, Firefox)
- [x] macOS (Chrome, Firefox, Safari)
- [x] Windows (Chrome, Edge, Firefox)


build/emoji.js
```diff
@@ -0,0 +1,91 @@
...
+  const response = await axios.get(emojiDataURL);
+  const baseURL = Object.values(response.data)
+    .find(url => /unicode\//)
+    .split('unicode/')[0];
+  const data = { ...response.data };
+
+  // Remove base URL from emoji URLs
+  Object.entries(data).forEach(
+    ([key, value]) => (data[key] = value.replace(baseURL, ''))
+  );
```
sy-records, line 27: There are some URLs without `unicode/`, will it affect?

like https://github.githubassets.com/images/icons/emoji/electron.png?v8
https://github.githubassets.com/images/icons/emoji/octocat.png?v8

jhildenbiddle: When rendering image-based emoji, these URLs work fine. When rendering native emoji, the code falls back to rendering image-based emoji when it encounters one of the URLs.

sy-records:
> Other: Deprecate emoji plugin (file remains to facilitate non-breaking change)

Should the contents of this file be retained?
There is an error in the CI. `ci/codesandbox — Building packages failed.`

build/emoji.js
jhildenbiddle: FYI, these are the emoji shorthand code that do not have unicode values associated with them in GitHub's API response:

* atom
* basecamp
* basecampy
* bowtie
* electron
* feelsgood
* finnadie
* goberserk
* godmode
* hurtrealbad
* neckbeard
* octocat
* rage1
* rage2
* rage3
* rage4
* shipit
* suspect
* trollface

Some of these (like `:atom:`) have [unicode values](https://emojipedia.org/atom-symbol/) associated with them but GitHub is not using them in their URLs for some reason. Others (like `:basecamp:`) are GitHub-specific, so the only way we can render them is using an image (as far as I know).

docs/plugins.md
```diff
@@ -75,29 +76,24 @@ Install the plugin and configure the track id.
...
+## Emoji
 
-The default is to support parsing emoji. For example `:100:` will be parsed to :100:. But it is not precise because there is no matching non-emoji string. If you need to correctly parse the emoji string, you need install this plugin.
-
-```html
-<script src="//cdn.jsdelivr.net/npm/docsify/lib/plugins/emoji.min.js"></script>
-```
-
-?> If you don't want to parse to emoji, you can use __colon_<span>_</span> or `&#58;`. If you need to use in the title, we recommend using `&#58;`. For example, `&#58;100:`
+!> This plugin has been deprecated. Docsify is able to render all GitHub-style emoji without the use of this plugin as of v4.13.
```
sy-records, line 96: I think this should be kept and the deprecated warning moved to the front?

Koooooo-7:
I m trying to understand the `native emoji` support.
If I write docs wanna use Native Emoji without net cost, I would use the emoji directly instead of markdown synatx （😀😬😁😂😃). It is a feat that we need support it ?

trusktr:
> I m trying to understand the `native emoji` support. If I write docs wanna use Native Emoji without net cost, I would use the emoji directly instead of markdown synatx （😀😬😁😂😃). is it a feat that we need support it ?

I could imagine someone not wanting to update some GitHub markdown file to actually use native emojis, and maybe that same person may not want extra network overhead from loading non-native emojis.

Koooooo-7:
> > I m trying to understand the `native emoji` support. If I write docs wanna use Native Emoji without net cost, I would use the emoji directly instead of markdown synatx （😀😬😁😂😃). is it a feat that we need support it ?
> 
> I could imagine someone not wanting to update some GitHub markdown file to actually use native emojis, and maybe that same person may not want extra network overhead from loading non-native emojis.

Make sense, maybe someone wanna have an auto translation between Native emoji and non-native. If so, I prefer this should be a `translator plugin` instead.

trusktr:
For the the fix for ci/codesandbox, seeing that it is an `uglify` error, I recommend updating to `terser` instead. The more time passes, the more outdated uglify will be, and the more likely it will break.

docs/plugins.md
jhildenbiddle: I can do that. FWIW, here was my rational for removing the description and the CDN URL:

* The plugin's description didn't do a very good job of explaining if/when people should use it (see #1726 as an example). I can rewrite the description, but it seemed strange to spend time writing a better description for a plugin that we are no longer telling people to use.
* Since we are no longer recommending the plugin, I didn't see value in providing a CDN URL that will load an empty JS file.

If you still prefer to display a description and CDN URL, I am happy to restore them. I would like to update the description to something a bit more user-friendly though.

@sy-records -- Thoughts?

jhildenbiddle:
@Koooooo-7 

> I m trying to understand the native emoji support. If I write docs wanna use Native Emoji without net cost, I would use the emoji directly instead of markdown synatx （😀😬😁😂😃). is it a feat that we need support it ?

The native emoji support in this PR is about converting shorthand codes like `:smile:` and `:+1:` to native emoji instead of images.

Users can add native emoji characters to their markdown today, but some users may not know how to do this, may find the process burdensome (depending on their OS' emoji picker implementation), or simply prefer the convenience of shortcodes that work in other apps they are familiar with (GitHub, Slack, Teams, etc.). 

Here is the process for adding a smile emoji using macOS:

1. Place the cursor where you want an emoji to appear in a document
2. Press CTRL+CMD+SPACE (how many macOS users know this?)
   CleanShot 2022-02-05 at 12 38 17@2x
3. Type the name of the emoji your looking for or scroll to find it
4. Click the emoji

If you accidentally click the wrong emoji (as I have countless times), you start the process all over again. 

Now, compare that to simply typing:

```markdown
:smile:
```

jhildenbiddle:
@trusktr --

> For the the fix for ci/codesandbox, seeing that it is an `uglify` error, I recommend updating to `terser` instead. The more time passes, the more outdated uglify will be, and the more likely it will break.

Yep. Agreed. The project has a lot of outdated dependencies that will continue to cause problems down the road. Let's take care of those updates in a separate PR though.

I think the ci/codesandbox fix should be an easy fix. Just haven't gotten around to it yet. I'll try to take care of it today.

jhildenbiddle:
FYI: There are other options besides the GitHub API for emoji data an images: [JoyPixels](https://github.com/joypixels/emoji-toolkit) (formerly [emojione](https://github.com/joypixels/emojione)) provides JSON files ([emoji.json](https://github.com/joypixels/emoji-toolkit/blob/master/emoji.json) and [emoji_strategy.json](https://github.com/joypixels/emoji-toolkit/blob/master/emoji_strategy.json)) which we could use as our emoji data source and their [emoji-assets](https://github.com/joypixels/emoji-assets/tree/master/png/32) are available on jsDelivr here: https://www.jsdelivr.com/package/npm/emoji-assets.

I'm not proposing we update this PR. Just wanted to add this in case an alternative is needed in the future. :)

Koooooo-7 (commented):
@jhildenbiddle  I see. I prefer to make it as a plugin instead.
I think we should make core more lite and plugins more richer to make it extend easily.

jhildenbiddle:
@Koooooo-7 --

I agree that docsify core should be kept light weight. That said, I think there may be a misunderstanding about what's included in this PR and why I'm proposing it be merged with core. I'll explain what's happening under the hood a bit (below) in hopes of clarifying.

TL;DR:

* Docsify ships with emoji support (i.e., emoji renders without a plugin) but the implementation is naive and has led to multiple issues created by users. This PR fixes these issues
* There is a documented history of confusion related to the emoji plugin. This PR resolves this confusion.
* The new native emoji features is ~10 lines of code. It adds *bytes* (not kilobytes) to the Docsify bundle sent to the client, so there is no need for concern regarding increased bundle size as a result of this feature.
* Removing the existing emoji parser from Docsify core and requiring a separate emoji plugin to render emoji `:shorthand:` codes is a perfectly reasonable suggestion (see #1726), however this would be a breaking change and would not address the existing issues without a major version bump.

I'd like to see us merge this PR, fix the existing issues, and revisit whether Docsify's emoji support should exist in core or as a separate plugin in the future when we are ready to release a new major version.

---

Docsify ships with emoji `:shorthand:`support, meaning it can render emoji shorthand codes *without* using the emoji plugin. For example, we do not use the emoji plugin on the Docsify website, yet we are still able to render a document emoji next to the "Edit Document" link at the top of each page:

CleanShot 2022-02-06 at 08 27 30@2x

This works because the emoji parser that ships with Docsify parses every page looking for emoji shorthand code (`:` + `emojicode` + `:`) and then replaces each one with an `` tag that uses the `emojicode` as the filename.

So this:

```markdown
:memo:
```

Gets converted to this (not the filename, `memo.png`):

```html
<img class="emoji" src="https://github.githubassets.com/images/icons/emoji/memo.png" alt="memo">
```

There are two problems with this approach:

1. It naively assumes every instance of `:` + `some text` + `:` is intended to be rendered as an emoji. This leads to non-emoji strings that match the emoji shorthand pattern to be replaced by broken images. For example:

   ```markdown
   hh:mm:ss
   ```

   Becomes:

   ```html
   hh<img class="emoji" src="https://github.githubassets.com/images/icons/emoji/mm.png" alt="mm">ss
   ```

   Which incorrect injects a broken image into the string (`mm` is not a valid emoji shorthand code):
   152634565-28cfa353-8344-4b35-94dd-696df6cc54d2

2. Many GitHub emoji images use filenames based on their unicode character and not their shortcode. For example:
   *  `1f44d.png` (<https://github.githubassets.com/images/icons/emoji/unicode/1f44d.png>)
   *  `thumbsup.png` (<https://github.githubassets.com/images/icons/emoji/unicode/thumbsup.png>)

These issues have been repeatedly called out by Docsify users (see #742, #829, #853, #1009, etc). The good news is that there is an easy fix: just use Docsify's emoji plugin! The bad news:

* People expect docsify's built-in emoji parser to be smart enough to properly identify valid emoji shorthand codes and not break things like their `date:hour:minute` or `Namespace::SubNamespace::Class` strings. Users see this as a bug, not an indication that they need to load a separate emoji plugin to fix the emoji parser that ships with Docsify.
* People unaware of the emoji plugin don't understand why some valid emoji shortcodes render (`:memo:`) but others don't (`:thumbsup:`),
* People aware of the emoji plugin are confused about its purpose given that Docsify is able to render emoji without it.

Bottom line: Emoji functionality feels broken without the plugin, but people are unaware that the plugin exists, unaware that the plugin will fix their problem, or unsure if they need the plugin at all. This is why we have issues/PRs created by users like #1721 and #1726.

If we want to address these issues, this PR is a non-controversial way to do it. The approach just rolls the emoji plugin's "smarts" into Docsify core, which just means emoji data from GitHub's API is used to ensure we're using valid image URLs and to test shorthand codes before replacing text with emoji. That's it.

This approach has the added benefit of also addressing the confusion surrounding the emoji plugin. Yes, the payload increases slightly (~15Kb w/ gzip), but we're fixing known issues that have surface multiple times and will likely surface again all without a breaking change.

As for the new native emoji feature, this was a trivial change that amounted to ~10 lines of code. This amounts to *bytes* (not kilobytes) added to Docsify's payload, so it wouldn't make much sense to offer this as a separate plugin on its own (if that's what you were proposing). This feature should live wherever Docsify's ability to render emoji lives (in core or the emoji plugin, but not on its own).

Finally...

To your point about keeping Docsify light weight and extending it via plugins, I agree. Removing the emoji parser from Docsify core and requiring the use of an emoji plugin to render emoji `:shorthand:` codes is a perfectly reasonable position to take. Doing so would be a breaking change that would require a major version bump, and we know the bump from v4 => v5 is a unique one because of Docsify's history of using CDN URLs with a `@latest` version lock. For this reason, I propose we postpone this discussion until we are ready to release a new major version of Docsify.

---

Hope this helps. I'm happy to continue to conversation and address any questions or concerns you have.

src/core/render/emojify.js
```diff
@@ -1,20 +1,45 @@
...
+    text
+      // Mark colons in tags
+      .replace(
+        /<(code|pre|script|template)[^>]*?>[\s\S]+?<\/(code|pre|script|template)>/g,
+        m => m.replace(/:/g, '__colon__')
+      )
+      // Mark colons in comments
+      .replace(/<!--[\s\S]+?-->/g, m => m.replace(/:/g, '__colon__'))
+      // Replace emoji shorthand codes
+      .replace(/:([\w\-+]+?):/g, (m, $1) =>
```
ChoKaPeek, line 39: About '\w': the idea between 'a-z0-9_' was to reduce false-positives, a word being 'a-zA-Z0-9_' and github emojis being lowercase.

Koooooo-7:
@jhildenbiddle  I 100% agree the fix about emoji parser. And after ur concerns about the native emoji feat, I think you convince me. :)

src/core/render/emojify.js
Koooooo-7: > 

Actually, I think we support the supper case in preview versions as well.

docs/plugins.md
sy-records: > I would like to update the description to something a bit more user-friendly though.

I agree.

jhildenbiddle: @sy-records --

I have updated the description, restored the `<script>` markup, and moved the deprecation warning.

src/core/render/emojify.js
jhildenbiddle: @ChoKaPeek: Good point. Fixed.

@Koooooo-7: Docsify's emoji parser wasn't case-sensitive but GitHub's servers are. That meant that Docsify would treat `:smile:` and `:SMILE:` the same but only the first would render properly. It therefore makes sense to restore @ChoKaPeek's RegEx change.

Example:
* https://github.githubassets.com/images/icons/emoji/smile.png
* https://github.githubassets.com/images/icons/emoji/SMILE.png

jhildenbiddle:
Alright, @docsifyjs/reviewers. This is ready for review. The outstanding issues have been addressed:

1. The codesandbox CI error is fixed
2. The emoji parser now only matches lowercase shortcodes (see reply to @ChoKaPeek & @Koooooo-7 above)
3. The emoji plugin docs have an updated descriptions and deprecation warning
4. The emoji plugin code has been restored. The plugin won't do anything if loaded along with these changes, but it is fully functional for older versions of docsify. This means an older version of docsify (< 4.13) could load these changesand enjoy updated emoji shortcode support. The end result is that the plugin is not needed but harmless if loaded with docsify v4.13+

Koooooo-7 (approved):
LGTM. :+1:

sy-records (approved):

jhildenbiddle:
Thanks for the reviews, @Koooooo-7 and @sy-records!

Issue #1221, opened 2020-06-14
Fix issues with docsify.js.org and IE11


jhildenbiddle:
## Bug Report

Docsify lists IE11 as an officially supported browser. While docsify itself works fine with IE11, docsify.js.org does not load properly in IE11 due to issues with several plugins. These issues should be addressed so that the docsify website loads properly in IE11.

#### Steps to reproduce

1. Visit https://docsify.js.org using IE11.
2. Open IE's developer tools
3. Click on the Console tab
4. Refresh the page

#### What is current behaviour

Multiple JS errors displayed in console.

#### What is the expected behaviour

No JS errors displayed in console.

#### Other relevant information

Errors caused by ES6+ syntax in the following plugins:

* [docsify-plugin-codefund](https://github.com/njleonzhang/docsify-plugin-codefund) (Issue: https://github.com/njleonzhang/docsify-plugin-codefund/issues/6)
* [docsify-dark-mode](https://github.com/Plugin-contrib/docsify-plugin/tree/master/packages/docsify-dark-mode) (Issue: https://github.com/Plugin-contrib/docsify-plugin/issues/16)

Issues have been filed in repos listed above.

jhildenbiddle:
The plugin-related issues have been resolved:

* `docsify-plugin-codefund` has been fixed: https://github.com/njleonzhang/docsify-plugin-codefund/issues/6#issuecomment-648252642
* `docsify-dark-mode` has been removed from https://docsify.js.org

The remaining issues are minor and cosmetic only:

1. Button background missing on coverpage
1. Theme color not being applied to elements (like links)
1. Search field placeholder text visibility is clipped

jhildenbiddle:
Fixed in #1743

Issue #707, opened 2018-11-26
Theme Color IE issue (improve bundling, etc, for older browsers)


sohrabtaee:
When I add `themeColor` it breaks in IE11 with this error:
```
SCRIPT8: The download of the specified resource has failed.
docsify.min.js (1,2752)
```

timaschew:
Please show a code snippet.

sohrabtaee:
The problem is that IE11 does not support CSS variables: https://caniuse.com/#feat=css-variables

I added the proper polyfills and it works fine now.

Since Docsify is advertised as "Compatible with IE10+", I guess a polyfill should be added to support CSS variables in IE.

anikethsaha:
shipping polyfills may increase the package size. I think we should update the docs with the guided steps ?

cc @jhildenbiddle @trusktr thoughts?


also, @jhildenbiddle if we need to ship polyfill for css variables from core, thoughts about [css-vars-ponyfill ?](https://github.com/jhildenbiddle/css-vars-ponyfill)

jhildenbiddle:
Best to have docsify handle loading all necessarily libraries (like poly/ponyfill) as needed via dynamic imports. This keeps the bundle size down, prevents users from having to manage multiple libraries just to get docsify working, and allows the docsify team to easily swap libraries for all users in future versions (e.g. switching from ponyfill "A" to ponyfill "B").

FWIW, [docsify-themeable](https://github.com/jhildenbiddle/docsify-themeable) bundles [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill), but this is only because times were different back when docisfy-themeable was created. If I were doing it today (and this is a change I will likely make in the future), I'd load css-vars-ponyfill via dynamic import instead of bundling:

```js
const hasCSSCustomPropSupport = window?.CSS?.supports?.('(--a: 0)');

if (!hasCSSCustomPropSupport) {
  import('https://cdn.jsdelivr.net/npm/css-vars-ponyfill@2')
    .then((module) => {
      // Do stuff...
    });
}
```

(If you unfamiliar with the `window?.css ...` syntax, see [@babel/plugin-proposal-optional-chaining](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining) for details)

anikethsaha:
We need to ship polyfill for dynamic import as well right ?

jhildenbiddle:
[...]

Issue #657, opened 2018-10-28
docsify 5.0 roadmap


QingWei-Li:
# New Features

- [x] New Feature: **Generate static html** #136
- [ ] New Markdown Syntax: GitHub(or GitLab) Flavored Markdown [ref](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown)
- [x] New Markdown Syntax: Friendly doc Helpers, #413
- [ ] New Option: For relative file paths rather than relative root, fix #593

# Breaking Changes
- [x] Remove `lib` and `themes` folder from the repo

# Deadline

~~December<sup>maybe</sup>~~ ~~February~~ Soon

MythRen:
what about generate diagrams and flowcharts from text using [Mermaid](https://mermaidjs.github.io/) ?

sansnom:
@MythRen there is a paragrah about mermaid support: https://docsify.js.org/#/markdown?id=supports-mermaid

timaschew:
My 2 cents:

- generating static HTML files is a important topic, because most link checkers won't work without static HTML
- global variables are very useful. Currently I don't see any way of defining global variables, across multiple mages and to use them in markdown files

timaschew:
I think it makes sense to add some test and fix some bugs which already lead to breaking changes by affecting the HTML output. Tests should be too hard since we can just have md as input and expect some HTMl as output. Is it correct that currently there are no automatic tests at all?

--- 

Another suggestion for 5.0 is to merge some features of [doczify](https://github.com/timaschew/doczify) to the docsify CLI. 

BTW: I've just pushed it, but I've already written the code a week ago.

timaschew:
I have another proposal which is not affecting the features itself, it's more the deployment.

Currently the lib directory is part of the master branch. Committing generated files into the source repository is actually an anti pattern. IMHO we should delete the lib directory from the repo and generate these files only to publish them to npm.
This makes the whole process less error-prone. This for instance would not happen again: https://github.com/docsifyjs/docsify/issues/717 The issue is about a breaking change, but if you look to the diff of source files between version 4.8.2 and 4.8.3 you don't see any relevant changes. It seems that the srouce (package.json) was not in sync with the generated files (files in lib/)

sansnom:
@timaschew it's already in their plan if you check the "Breaking changes" part inside the first post: "Remove lib and themes folder from the repo". :)

jhildenbiddle:
A few additional proposals for 5.x:

[...]

PR #1747, opened 2022-02-05
Drop IE11 support.

trusktr:
## **Summary**

Microsoft is about to do the same thing: https://docs.microsoft.com/en-us/lifecycle/products/internet-explorer-11

Plus, Microsoft already has "IE Mode" in Edge which it will support until 2029: https://www.microsoft.com/en-us/edge/business/ie-mode

There will be no reason to keep IE support soon.

  Docs

## **For any code change,**

- [ ] Yes
- [x] No

## **Related issue, if any:**

closes https://github.com/docsifyjs/docsify/issues/1221
closes https://github.com/docsifyjs/docsify/issues/707

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] IE 😄


jhildenbiddle:
We have discussed and agreed to a plan for dropping support for IE11 several times in the past.  See https://github.com/docsifyjs/docsify/issues/657#issuecomment-458341229 for one of those conversations.

TL;DR: We can drop support for IE11 with v5, but we should not bump to v5 just to drop IE11 support.

We're all ready to put IE11 behind us. Given the date on the comment linked above, we've been ready for over three years to move on. That said, we can't drop support for IE11 without publishing a new major version, and because of Docsify's history of linking to CDN URLs locked to the `@latest` version the jump from v4 to v5 is not as straightforward as a typical major version bump. There are ways to work around this, but the launch of v5 will require a conscientious effort to not break what could be a significant number of live Docsify sites.

Equally important, we should not be so quick to assume legacy browsers (like IE11) are not being used anywhere. Here's just one data point to consider: I created [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill) to add support for CSS custom properties (aka "CSS variables") to legacy browsers which is downloaded ~130 million times per month from one CDN ([jsDelivr](https://www.jsdelivr.com/package/npm/css-vars-ponyfill)) and ~500 thousand times per month from [npm](https://www.npmjs.com/package/css-vars-ponyfill). Here are live stats:

jsDelivr hits (npm) npm

We can't derive the number of IE11 users from those stats, but the stats do indicate that many developers and site visitors continue to use tools designed specifically to support IE11 even into 2022 (as sad as that may be). I continue to get questions from developers to this day asking things like how to integrate the ponyfill into their app which is using the latest version of Angular.

For comparison, here are the same live stats for docsify (~7M/mo for [jsDelivr](https://www.jsdelivr.com/package/npm/docsify), 117K/mo for [npm](https://www.npmjs.com/package/docsify) when written):

jsDelivr hits (npm) npm

FYI: The linked issues above were fixed by #1743. I've marked them as resolved and closed them.

@docsifyjs/core @docsifyjs/reviewers: Vote to close this PR and drop IE11 support with the next major version.

trusktr:
What if we move the a new version (whether it is 4.x, 5.x, or etc) with build output at a new path, and leave the last IE-compatible version at the same path so `@latest` with the current URLs will never break?

Koooooo-7 (commented):
I agree with @jhildenbiddle  that we need not drop this IE support for now. 
If user has new request on IE 11 we could mention them we will remove IE support in the future.

jhildenbiddle:
> What if we move the a new version (whether it is 4.x, 5.x, or etc) build output to a new path, and leave the last IE-compatible version at the same path so that `@latest` with the current URLs will never break?

That's an option. Users would lose the ability to load the latest version of Docsify using the `@latest` URL version lock or by omitting a version lock from the URL entirely, but maybe that isn't such a bad thing.

But for now...

> The TL;DR is that we can do this with a major version bump (v4 => v5), but we should not publish a new major version just to drop IE11 support.

Let's close this PR and open where we can discuss approaches for releasing v5 (hopefully with a "how should we do it" focus instead of a "what should it contain" debate).

@trusktr -- I generally prefer to let authors close their own issues/PRs, so I'm leaving this open until you're ready.

jhildenbiddle:
Closing due to reviewer consensus and lack of inactivity. We'll revisit legacy support in v5.

PR #1751, opened 2022-02-14
feat: support relative path with target config.

### LINKED ISSUE #1750: Support relativePath link set linkTarget
Hi
When I read the  docs of "Set target attribute for link" https://docsify.js.org/#/helpers?id=set-target-attribute-for-link 
I try a relative path to jump

```
[demo](/docs/front_end/summery_of_experience/data_use/assets/demo/index.html ':target=_blank')
```

but It be compiled as
```
<a href="/docs/front_end/summery_of_experience/data_use/assets/demo/index.html">demo</a>
```
It is not my expectation.

After then, I read the source code: "src\core\render\compiler\link.js" line 22
I discover that `docsifyjs` didnt support relative path linkTarget custom

so I make the issue in order to discuss the problem

thx

Koooooo-7:
## **Summary**

Support relative path set `target` config.
close #1750 

Bugfix

## **For any code change,**

- [ ] Yes
- [x] No

## **Related issue, if any:**
#1750 

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


src/core/render/compiler/link.js
```diff
@@ -28,6 +28,12 @@ export const linkCompiler = ({
       }
 
       href = router.toURL(href, null, router.getCurrentPath());
+
+      if (config.target) {
+        attrs.push(
+          href.indexOf('mailto:') === 0 ? '' : `target="${linkTarget}"`
```
jhildenbiddle, line 34: Does the empty string in the `attrs` array serve a purpose?

I'm not familiar with this part of the code so perhaps this is what needs to be done, but if the empty string does not serve a purpose (i.e., it's skipped when `attrs` is processed) then perhaps we should avoid it using a simple `if` statement:

```js
if (config.target) {
  if (href.indexOf('mailto:') !== 0) {
    attrs.push(`target="${linkTarget}"`)
  }
}
```

Or a single-line conditional like this:

```js
if (config.target) {
  href.indexOf('mailto:') !== 0 && attrs.push(`target="${linkTarget}"`);
}
```

jhildenbiddle (changes requested):
One minor stylistic question, otherwise looks good. :)

src/core/render/compiler/link.js
Koooooo-7: Got it !

jhildenbiddle (dismissed):
LGTM!

src/core/render/compiler/link.js
trusktr: Just my preference, but:

```js
if (config.target && href.indexOf('mailto:') !== 0) {
  attrs.push(`target="${linkTarget}"`);
}
```

trusktr:
Why do we check if it doesn't start with `mailto:`? Does that break mailto links? 

I'm of the opinion that we shouldn't get in the way, should not change intent. So if someone writes a mailto link with a target, just like they can in regular HTML (`<a href="mailto:..." target="blank">`), we should just let it be.

It in other words, it shouldn't be our goal to change DOM conventions, but only to provide the interface to it as much as we can.

jhildenbiddle:
@trusktr --

> Why do we check if it doesn't start with mailto:? Does that break mailto links?

I don't know the history, but my guess is that the original contributor was unaware of the usefulness of using the `target` attribute on `mailto:` links. The `target` attribute is ignored when the default email client is an external app (Apple Mail, Outlook, Thunderbird, etc) but it is honored when the default email client is web-based (Gmail, iCloud, Outlook.com, etc).

> I'm of the opinion that we shouldn't get in the way, should not change intent. So if someone writes a mailto link with a target, just like they can in regular HTML (`<a href="mailto:..." target="blank">`), we should just let it be.

Agreed, but this is a separate issue. My vote would be to review this PR for its intended purpose and create another PR to "fix" the target attribute on `mailto:` links.

jhildenbiddle (approved):

Koooooo-7:
@trusktr  Hi, as @jhildenbiddle  said, it is the original contributor did, I wanna keep it incase of any unpredicted changes.
If we could change the `mailto:` issue, I agree that we can have a new one  to fix that .

jhildenbiddle:
@Koooooo-7 --

Since this has been targeted for v5, can we update this branch and get one more review/approval? @sy-records? @trusktr?

jhildenbiddle:
👆 @sy-records

sy-records (approved):

PR #1798, opened 2022-05-12
Upgrade marked

tgarciaalcala-godaddy:
## **Summary**
https://github.com/docsifyjs/docsify/pull/1724
Docsify depends on an ancient version of marked which has some CVEs. The fix it to upgrade to marked v4.

Bugfix

## **For any code change,**

- [X ] Yes
- [ ] No

**marked** node dependency upgraded to version 4 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,4 +125,3 @@ This project exists thanks to all the people who contribute. [[Contribute](CONTR
 ## Special Thanks
 
 A preview of Docsify's PR and develop branch is <a href="https://vercel.com/?utm_source=docsifyjs&utm_campaign=oss" target="_blank">Powered by <img src="https://cdn.jsdelivr.net/gh/docsifyjs/docsify/docs/_media/vercel_logo.svg" alt="Vercel" width="133px"></a>
-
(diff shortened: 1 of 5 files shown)


kinetifex:
Great to see this - I'm curious and concerned about the tests though. It seems like the changes around using playwright may be yielding false positives.

For example, I had started similar work against an older commit [here](https://github.com/docsifyjs/docsify/compare/develop...kinetifex:upgrade-marked), but integration tests fail locally catching the `marked` [breaking changes from v3](https://github.com/markedjs/marked/releases/tag/v3.0.0). So, I'm really surprised to see these test passing. I would expect more code changes to be required. For a specific example, [this code](https://github.com/tgarciaalcala-godaddy/docsify/blob/feat/SYNTH-843/src/plugins/search/search.js#L57-L67).

Any insights here docsify team? It would be great to get `marked` updated to address the CVEs.

tgarciaalcala-godaddy:
@kinetifex

Issue #1821, opened 2022-06-17
Support specifying short-titles for the sidebar from content pages?


Jacco-V:
#### What problem does this feature solve?

This feature would allow long titles for the documentation headers, with short titles in the sidebar, when needed. Right now, this can be done by manually creating each and every entry for the sidebar, but this:

a) takes *a lot* of time,
b) is very prone to errors, and
c) forces you to write each and every sidebar entry yourself, even if short-titles are only required for a couple of entries.

#### How should this be implemented in your opinion?

Right now, Docsify supports [Customise ID for headings](https://docsify.js.org/#/helpers?id=customise-id-for-headings). It would be great if it would also support something like:

```
## How would I write an "hello, world" example? :sidebar=hello-world
```

#### Are you willing to work on this yourself?

Willing? yes, but realistically, I don't have the time to learn frontend development.

Jacco-V:
I would be willing to try and implement this, but I have not idea where to start?

julienw:
My understanding is that the title is generated here:
https://github.com/docsifyjs/docsify/blob/54cc5f939b9499ae56604f589eef4e3f1c13cdc5/src/core/render/compiler.js#L210

As you see it's taken from `str` quite directly. But you'd like that it's taken from the `config` object instead.

Maybe you can try doing something like this in line 238:
https://github.com/docsifyjs/docsify/blob/54cc5f939b9499ae56604f589eef4e3f1c13cdc5/src/core/render/compiler.js#L236-L239
and add:
```js
if (config.sidebar) {
  nextToc.title = config.sidebar;
}
```

Jacco-V:
Please see PR [#1825](https://github.com/docsifyjs/docsify/pull/1825)

Jacco-V:
#1825

PR #1825, opened 2022-06-22
feature: support customized sidebar item name from content

Jacco-V:
Add support for setting a customized sidebar item name from the documentation markdown content

See also issue: [#1821](/docsifyjs/docsify/issues/1821)

## **Summary**

This feature add support for specifying custom sidebar item names (especially for long titles) in the documentation markdown files, so that a the custom (typically shortened) item name is used in the sidebar:

```md
## How would I write an "hello, world" example? :sidebar="hello world?"
```

image

Feature

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] IE

### DIFF
--- a/src/core/util/str.js
+++ b/src/core/util/str.js
@@ -5,3 +5,12 @@ export function startsWith(str, prefix) {
 export function endsWith(str, suffix) {
   return str.indexOf(suffix, str.length - suffix.length) !== -1;
 }
+
+export function removeDocsifyIgnoreTag(str) {
+  return str
+    .replace(/<!-- {docsify-ignore} -->/, '')
+    .replace(/{docsify-ignore}/, '')
+    .replace(/<!-- {docsify-ignore-all} -->/, '')
+    .replace(/{docsify-ignore-all}/, '')
+    .trim();
+}
--- a/src/plugins/search/search.js
+++ b/src/plugins/search/search.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-unused-vars */
 import { getAndRemoveConfig } from '../../core/render/utils';
+import { removeDocsifyIgnoreTag } from '../../core/util/str';
 
 let INDEXS = {};
 
@@ -89,19 +90,16 @@ export function genIndex(path, content = '', router, depth) {
     if (token.type === 'heading' && token.depth <= depth) {
       const { str, config } = getAndRemoveConfig(token.text);
 
+      const text = removeDocsifyIgnoreTag(token.text);
+
       if (config.id) {
         slug = router.toURL(path, { id: slugify(config.id) });
       } else {
-        slug = router.toURL(path, { id: slugify(escapeHtml(token.text)) });
+        slug = router.toURL(path, { id: slugify(escapeHtml(text)) });
       }
 
       if (str) {
-        title = str
-          .replace(/<!-- {docsify-ignore} -->/, '')
-          .replace(/{docsify-ignore}/, '')
-          .replace(/<!-- {docsify-ignore-all} -->/, '')
-          .replace(/{docsify-ignore-all}/, '')
-          .trim();
+        title = removeDocsifyIgnoreTag(str);
       }
 
       index[slug] = { slug, title: title, body: '' };


src/core/render/compiler.js
```diff
@@ -207,38 +207,36 @@ export class Compiler {
...
 
+      const title = config.sidebar || str;
       const slug = slugify(config.id || str);
       const url = router.toURL(router.getCurrentPath(), { id: slug });
+      nextToc.title = removeAtag(title);
       nextToc.slug = url;
       _self.toc.push(nextToc);
 
-      return `<h${level} id="${slug}"><a href="${url}" data-id="${slug}" class="anchor"><span>${str}</span></a></h${level}>`;
+      return `<h${level} id="${slug}"><a href="${url}" data-id="${slug}" class="anchor"><span>${title}</span></a></h${level}>`;
```
julienw, line 239: I haven't tried the patch, but looking at this I would think that the sidebar text would also be used in the rendered page, but I don't think this is correct. We want the sidebar text in the sidebar but not in the page.

Have you tried this? :-)

Jacco-V: No, I have not tried the changes, because I do not have a working frontend development environment. (I'm a backend dev giving this a try).

sy-records: create a _sidebar.md https://docsify.js.org/#/more-pages?id=sidebar

Jacco-V:
I pushed some changes, but it does not work as expected :(
It seems that the `:sidebar="hello world?"` does not get parsed as `config`?

```md
## How would I write an "hello, world" example? :sidebar="hello world?"
```

Jacco-V:
I've added quoted string support to the [getAndRemoveConfig](https://github.com/Jacco-V/docsify/blob/feature-customizableSidebarItemName/src/core/render/utils.js#L21=L44) function and updated the tests.

Jacco-V:
Any chance anyone of the reviewers is going to review this?

Jacco-V:
**Code has been tested and is working as intended**, with the sole caveat that:

* we cannot support *escaped characters in quoted-string config-values*, because by the time the `str` is passed to the [`getAndRemoveConfig()`](https://github.com/docsifyjs/docsify/blob/da7937e340ec44b514fc397b3c1b1734eb938952/src/core/render/utils.js#L21=L44) method, it has been html-escaped, and backslashes have been stripped. I expect his will not be an issue for the large majority of use cases.

src/core/render/utils.js
```diff
@@ -23,16 +23,18 @@ export function getAndRemoveConfig(str = '') {
...
-      .replace(/^('|")/, '')
-      .replace(/('|")$/, '')
-      .replace(/(?:^|\s):([\w-]+:?)=?([\w-%]+)?/g, (m, key, value) => {
-        if (key.indexOf(':') === -1) {
-          config[key] = (value && value.replace(/&quot;/g, '')) || true;
-          return '';
-        }
+      .replace(
+        /(?:^|\s):([\w-]+:?)=?([\w-%]+|&quot;((?!&quot;).)*&quot;|[“”][^“”]*[“”])?/g, // Note: because the provided `str` argument has been html-escaped, with backslashes stripped, we cannot support escaped characters in quoted strings :-(
+        function (m, key, value) {
```
Koooooo-7, line 28: FYI, there is a bit complicated regex including `negative lookahead`, we may need more tests here.

Jacco-V: Yes, the regex became more complicated because the incomming `str` argument has been transformed from markdown to html. This is also the reason we cannot support escaped characters in the quoted-string config-values. It would have been a lot cleaner if the markdown to html transformation would have been applied after the config strings would have been read.

Jacco-V:
bump

Jacco-V:
Hi,
Is anyone going to review this PR? or is this functionality somehow not good enough?

Cheers,

Koooooo-7 (commented):
Hi @Jacco-V , Custom sidecar item name is a good purpose.
Personally, I m considering what is the good `mark` to define a config content , whether we should use the `"` here or not . therefore I may not approve this immediately.

For the instance, based on this PR , using the `mark` to include content means this `mark` can not be used. i.e.
If using `"` as  the mark ( `:sidebar="content"`), we can not use the `"` such as `:sidebar="the include "  content "` like you mentioned.
Instead, if we change the `mark` to `:sidebar={#"any content {in here}"}` like:
```js
    str = str
      .replace(
         /(?:^|\s):([\w-]+:?)=?([\w-%]+|{#.*})?/g,
        function (m, key, value) {
          if (key.indexOf(':') === -1) {
            config[key] = (value && value.replace(/{#|}$/g, '')) || true;
            return '';
          }

          return m;
        }
      )
      .replace(/^('|")|('|")$/g, '')
      .trim();
```
we can get this thing :
image

The point is what the `mark` should be the common rule/guidline for those exist/future config and it should not be very common use in user content.

Any ideas?

Jacco-V:
@Koooooo-7 Yes, I can see what you mean. Architecturally speaking, it would be best if the `str` argument would be passed to the `getAndRemoveConfig()` function _before_ the markdown has been transformed into html.

because, if this was the case, we could fully support quoted strings, including escape characters. This would mean you get the benefit of a using a syntax everybody knows, and knows what to expect from. 

```
:sidebar="In quoted strings we can use escaped \" characters"
```

However, I'm not sure it is possible to do this, without significant changes to the code base. Having said that, from an architectural point of view, it would still be the best option.

Another option would be to replace the regular-expression based 'parsing' with an actual parser, which would be able to handle escape charters even in the html formatted `str` argument. This option would get the quoted strings working, without addressing the architectural issue, but the end user won't see the difference.

The other option could be as you propose: go for a different set of 'start of string' and 'end of string' markers. This would sidestep the problem. The code-changes would be more simple, at the cost introducing a custom format instead of using quoted strings.

In the end, it is not up to me to choose which option is best suited for this project: I'm not one of the project members; I'm just a back-end guy who contributed a pull-request.

Koooooo-7:
@Jacco-V  Thx for your input !
About change the `regex` to `parser` for configs is a good idea but we have not moved forward to it yet.
And exposing the `getAndRemoveConfig` as a hook is on the road. I suppose this hook could let user to custom config tpl more flexible, we could only provide a happy path one by default.

jhildenbiddle:
@Jacco-V @Koooooo-7 --

I haven't tested this PR, but perhaps we can work with the escaped character limitations using one of the following approaches:

1. Allow titles to be wrapped in single or double quotes. This allows for either quote type to be used within the string without needing to be escaped:
   ```
   :sidebar="Single 'quotes' here"
   ```
   ```
   :sidebar='Double "quotes" here'
   ```
2. Use HTML entities
   ```
   :sidebar='Double &quot;quotes&quot; here'
   ```


FWIW, I would also change the "Customize..." sidebar links to the following:

- Helpers > Heading IDs
- Helpers > Sidebar Titles

Koooooo-7:
@jhildenbiddle How you doing ~ 

About the `Customize ` changes, I think it is a nice change to make the label more plain and understandable.

The `"/'` approach it sounds both fine. The more importance thing to me is the now and further enrichment configs, I think we should refactor `getAndRemoveConfig` to be  more flexible parser(s) for configs, otherwise , the regex would be more and more complex and messy.

Jacco-V:
@jhildenbiddle This is a possibility, and it could be an intermediate solution, but it would also be another incarnation of 'sidestepping the problem', although with a better solution than introducing custom `start of string` and `end of string` markers (in my opinion).

However, by sidestepping the problem, you continue building on non-robust legacy choice and will need to add increasingly complex regular expression(s) to hammer the non-ideal setup into a working solution. So, I agree with @Koooooo-7: Going the  `getAndRemoveConfig` hook, with a parser way, you get more flexible, more robust, and easier to maintain code.

jhildenbiddle:
@Jacco-V, @Koooooo-7 --

No arguments from me about the issues with building upon problematic legacy code. My comment was intended only to propose a way to make the stated drawback of this PR -- the inability to support  escaped characters in quoted-string config-values -- less of an issue so that we can delivery a solution sooner. If someone wants to take a crack at doing things the "better" way, I'm all for it.

Issue #1839, opened 2022-07-18
Question: Parsing URL Links that have characters wrapped in colons (ie :w:)


andrefox333:
Hi team, quick question on how to properly manage URL links that has `:v:` or any character that is wrapped in a colon. Right now it's trying to convert these `:v:` into emoji characters... Any help is deeply appreciated. Thanks!

jhildenbiddle:
Duplicate of #1823

Issue #1823, opened 2022-06-22
Emoji parsing should exclude URLs


socsieng:
## Bug Report

Parsing of `:emojis:` should exclude URLs as this causes URLs in the following format render invalid HTML:

```
[link](https://link.com/:smile:/index)
```

Note that in my use case, I am also using other `:emojis:` so I can't set `noEmoji: true` in config.

Happy to contribute a fix if this bug is accepted.

#### Steps to reproduce

Example: https://codesandbox.io/s/brave-margulis-tw5t2b?file=/index.html

#### What is current behaviour

Invalid HTML is rendered:

```
<a alt="smile" src="https://github.githubassets.com/images/icons/emoji/smile.png" href="https://link.com/&lt;img class=">/index" target="_blank"  rel="noopener"&gt;link</a>
```

<img width="324" alt="image" src="https://user-images.githubusercontent.com/2647062/174975972-36e2ca3e-1fc8-4fb7-a100-a9180cd3b377.png">

#### What is the expected behaviour

Rendered HTML should produce something like:

```
<a href="https://link.com/:smile:/index" target="_blank"  rel="noopener">link</a>
```

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: macOS
- Node.js version:
- npm/yarn version: 
- Browser version: Safari 15.5
- Docsify version: 4
- Docsify plugins: 

#### Please create a reproducible sandbox 

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/brave-margulis-tw5t2b?file=/index.html)

sy-records:
Thank you for your feedback.

Have you tried #1746?

socsieng:
I've tried with `nativeEmoji: true` and it still suffers from the same problem: https://codesandbox.io/s/practical-borg-nb3vdl?file=/index.html

jhildenbiddle:
@socsieng --

Ideally, docsify should handle this for you. If you're up for creating the PR, that would be great as I'm not sure I'll have the bandwidth anytime soon.

If you need a quick fix, you can just manually encode any portion of a URL that requires it (like a string matching the double-colon emoji pattern). You can do this using an online [encode/decode tool](https://www.urlencoder.org) or with the console in your browser's dev tools:

```js
encodeURIComponent(':smile:')
```

```markdown
[link](https://link.com/%3Asmile%3A/index)
```

PR #1847, opened 2022-07-30
fix: Ignore emoji shorthand codes in URIs

socsieng:
Fixes: #1823,
Fixes: #1839

## **Summary**

Fixes an [issue](https://github.com/docsifyjs/docsify/issues/1823) related to unintentional processing of emoji shorthand codes when included in a URI

Bugfix

## **For any code change,**

- [ ] Yes
- [x] No

Using: https://codesandbox.io/s/thirsty-fog-inwfl8

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE


src/core/render/emojify.js
```diff
@@ -35,6 +35,8 @@ export function emojify(text, useNativeEmoji) {
       )
       // Mark colons in comments
       .replace(/<!--[\s\S]+?-->/g, m => m.replace(/:/g, '__colon__'))
+      // Mark colons in URIs
+      .replace(/[a-z]{2,}:\/\/[^\s]+/gi, m => m.replace(/:/g, '__colon__'))
```
jhildenbiddle, line 39: This regex will miss URLs that omit the protocol:

```html
<a href="//domain.com/:smile/">Smile</a>
```

jhildenbiddle (changes requested):
Thanks for submitting the PR, @socsieng.

There is a much easier fix: simply add `a` to the existing list of tags that are being marked:

https://github.com/docsifyjs/docsify/blob/682bf967aed9bc557c688360d9ca61b81b4d2db0/src/core/render/emojify.js#L31-L35

Change this line:

https://github.com/docsifyjs/docsify/blob/682bf967aed9bc557c688360d9ca61b81b4d2db0/src/core/render/emojify.js#L33

To this:

```js
/<(a|code|pre|script|template)[^>]*?>[\s\S]+?<\/(a|code|pre|script|template)>/g,
```

This fix also addresses an issue with the regex you were using in your initial PR.

socsieng:
Wouldn't marking the `a` tags be too broad? I can imagine emojis being used in many `a` tags today.

> This fix also addresses an issue with the regex you were using in your initial PR.

What was the issue with the regex?

src/core/render/emojify.js
socsieng: I was trying to be a strict as possible with the regex to reduce the number of false positives. Happy to update the expression to make `scheme:` optional.

The root cause of my issue was related to absolute SharePoint urls being pasted into markdown which were all absolute and included the scheme.

socsieng: Note that //domain.com/:smile:/ doesn't https://domain.com/:smile:/ appear to be automatically parsed as a URI like a fully qualified URI does by the markdown parser.

What do you prefer? Optional versus required `scheme:`?

socsieng, line 39: Can change to the following if you would like it to be optional:

```suggestion
      .replace(/([a-z]{2,}:)?\/\/[^\s]+/gi, m => m.replace(/:/g, '__colon__'))
```

jhildenbiddle: You're right about ignoring everything inside `<a>` tags. That will miss some shortcodes that I believe people will expect to be processed.

There are a number of scenarios to consider. We should probably start by defining the behavior. For example:

```html
<style>
  @import url("http://domain.com/:smile/file.css");
  @import url('http://domain.com/:smile/file.css');
  @import url(http://domain.com/:smile/file.css);
  @import "http://domain.com/:smile/file.css";
  @import 'http://domain.com/:smile/file.css';

  .myclass::before {
    content: ":smile:";
    background: url("http://domain.com/:smile/");
    background: url('http://domain.com/:smile/');
    background: url(http://domain.com/:smile/);
  }
</style>

<a href="http://domain.com/:smile:/">:smile: http://domain.com/:smile:/ :smile:</a>
<p>:smile: http://domain.com/:smile:/ :smile:</p>

<img src="http://domain.com/:smile:/file.png" alt=":smile:">
<img srcset="http://domain.com/:smile:/file.jpg 320w" src="http://domain.com/:smile:/file.png">

<script src="http://domain.com/:smile:/file.js"></script>
```

Here's how I would assume we would want to final output to look like:

```html
<style>
  @import url("http://domain.com/:smile/file.css");
  @import url('http://domain.com/:smile/file.css');
  @import url(http://domain.com/:smile/file.css);
  @import "http://domain.com/:smile/file.css";
  @import 'http://domain.com/:smile/file.css';

  .myclass::before {
    content: "😄";
    background: url("http://domain.com/:smile/");
    background: url('http://domain.com/:smile/');
    background: url(http://domain.com/:smile/);
  }
</style>

<a href="http://domain.com/:smile:/">😄 http://domain.com/:smile:/ 😄</a>
<p>😄 http://domain.com/:smile:/ 😄</p>

<img src="http://domain.com/:smile:/file.png" title="😄">
<img srcset="http://domain.com/:smile:/file.jpg 320w" src="http://domain.com/:smile:/file.png">

<script src="http://domain.com/:smile:/file.js"></script>
```

I'm probably missing something, but the rules for the above scenarios would be:

1. Ignore emoji shortcodes in `href`, `src`, and `srcset` attributes
1. Ignore emoji shortcodes in `url()` values
1. Ignore emoji shortcodes in text URLs within tags

Thoughts?

socsieng: I think that looks about right. 3 is a useful _catch all_ for URLs anywhere in the content, while 1 and 2 are useful for handling partial URLs (e.g. relative).

For my use case, 3 is sufficient and what this pull request is meant to fix. 1 and 2 are nice to have.

jhildenbiddle: I believe we can handle all three scenarios by making a small change to your suggested change above:

```js
.replace(/([a-z]{2,}:)?\/\/[^\s'">)]+/gi, m => m.replace(/:/g, '__colon__'))
```

Tested here: https://regex101.com/r/Paa6KG/1

jhildenbiddle (changes requested):

src/core/render/emojify.js
socsieng: Fixed

socsieng: Fixed

jhildenbiddle (approved):
LGTM! Thanks again, @socsieng.

sy-records (approved):
LGTM. Thank you!

socsieng:
Any ideas about when these changes will be available in npm?

sy-records:
Not sure when npm will be released.

But you can try https://github.com/sy-records/docsify-nightly

trusktr:
You can also try installing from github using `npm`. To install from `develop`:

```sh
npm install "docsify@docsifyjs/docsify#develop"
```

Or to install from a commit:


```sh
npm install "docsify@docsifyjs/docsify#3c9b3d9702bb05a5ff45a4ce4233e144cf1ebecb"
```

Or any git ref:


```sh
npm install "docsify@docsifyjs/docsify#<GIT_REF_GOES_HERE>"
```

which can be a tag, branch, etc.

socsieng:
Thanks. Appreciate the help.

I was more wondering what the timelines were to determine if I should wait for a formal package update or if I should use a github based version.

jhildenbiddle:
@socsieng --

I think we're close to pushing an update. There's one other PR that I'd like to see merged before we do so, but otherwise I don't think anyone would have an issue pushing a new minor or patch release (have to check history to determine which would be appropriate). I'll see if I can wrap that up today or tomorrow.

PR #1850, opened 2022-08-02
Docs: Fix plugin insertion order in docs

jhildenbiddle:
## **Summary**

This simple change to the plugin template provided in the docs ensures that plugins are inserted into the `$docsify.plugins` array in the order that the `<script>` tags appear in the DOM. Before this change, plugins were always inserted at the beginning of the plugins array which can (and has) caused issues when the order of plugin execution is important.

Docs

- [ ] Yes
- [x] No


Koooooo-7 (approved):
LGTM.

sy-records (approved):

sy-records (approved):

PR #1855, opened 2022-08-12
fix: When alias contains parameters, append ext error

### LINKED ISSUE #1854: .md getting appended while using alias
#### Steps to reproduce
Hi
I am trying to use alias to resolve the files in other repo. While doing that I need to add to token to get access.
'/changlog':'https://some/raw/url/README.md?token=Mytoken,

When I try to have the above alias, the value is getting resolved to https://some/raw/url/README.md?token= Mytoken.md.

The result is suffixing with ".md" . 
Please help me resolve this issue.

`https://some/raw/url/README.md?token= Mytoken.md`
.md gets appended at the end of token url while using alias 

`https://some/raw/url/README.md?token= Mytoken`

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 
- Docsify plugins: 

Here are files to reproduce the issue
[test with vs code](https://github.com/docsifyjs/docsify/files/9308204/ezyzip.zip)

sy-records:
## **Summary**

As mentioned in #1854, if the URL used by the alias contains parameters, we will add the 'ext' configuration item incorrectly

Bugfix

## **For any code change,**

- [ ] Yes
- [x] No

## **Related issue, if any:**

Fix #1854

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/src/core/router/history/base.js
+++ b/src/core/router/history/base.js
@@ -26,7 +26,9 @@ function getFileName(path, ext) {
     ? path
     : /\/$/g.test(path)
     ? `${path}README${ext}`
-    : `${path}${ext}`;
+    : path.indexOf('?') === -1
+    ? `${path}${ext}`
+    : path;
 }
 
 export class History {


Koooooo-7 (approved):
LGTM.
In line 85, it seems a type `bug`.

test/unit/router-history-base.test.js
```diff
@@ -66,4 +66,35 @@ describe('router/history/base', () => {
...
+      const file = history.getFile('https://some/raw/url/README.md');
+
+      expect(file).toBe('https://some/raw/url/README.md');
+    });
+    test('path is url, but ext is .html', () => {
+      const file = history.getFile('https://foo.com/index.html');
+
+      expect(file).toBe('https://foo.com/index.html');
+    });
+    test('path is url, bug with parameters', () => {
```
Koooooo-7, line 85:
```suggestion
    test('path is url, but with parameters', () => {
```

Koooooo-7 (dismissed):

Koooooo-7 (approved):

Koooooo-7 (dismissed):

src/core/router/history/base.js
```diff
@@ -26,7 +26,9 @@ function getFileName(path, ext) {
     ? path
     : /\/$/g.test(path)
     ? `${path}README${ext}`
-    : `${path}${ext}`;
+    : path.indexOf('?') === -1
+    ? `${path}${ext}`
+    : path;
```
trusktr, line 31: We need to handle all URLs the same with or without the query, just like regular HTML in a browser.

For example, `/path/to/folder/?foo=123` has a query, and we still need to look for `/path/to/folder/README.md?foo=123`, but this skips anything with a query.

In the end, we need to operate on the path separately, and always send the query with the final result.

trusktr (changes requested):
See comment. Basically this still has other problems. The solution is to handle paths and queries separately

SumithChandran:
> See comment. Basically this still has other problems. The solution is to handle paths and queries separately
Hi,
Any new updates on this issue.
We were working on federating multiple repo's readme.md files to a single documentation page.

src/core/router/history/base.js
vergilcw: @trusktr, is there anything that could help push this along? I'm not that familiar with  js but I am good with regular expressions. If you can provide a little more context on the issue perhaps I could put a little time into it, as fixing this would open up a lot of doors for those of us needing to federate markdown files from private repos/orgs on GitHub.

trusktr: @vergilcw sorry for late reply. I think the solution should determine the fully qualified URL (f.e. `anchor.href = partialUrlString; const fullyQualifiedUrl = anchor.href`) and then use a `new URL` object to get all the pieces and use them correctly.

trusktr:
> > See comment. Basically this still has other problems. The solution is to handle paths and queries separately
> > Hi,
> > Any new updates on this issue.
> > We were working on federating multiple repo's readme.md files to a single documentation page.
> > Fixing this error would help a lot.

@SumithChandran hi, sorry for taking long. Did you find a solution? I [commented](https://github.com/docsifyjs/docsify/pull/1855#discussion_r1087479706) on a direction we should look at. We just need to handle all the URL parts correctly, f.e. get all the parts of the URL, do what we need with only the path part, then put everything back together properly.

SumithChandran:
> > > See comment. Basically this still has other problems. The solution is to handle paths and queries separately
> > > Hi,
> > > Any new updates on this issue.
> > > We were working on federating multiple repo's readme.md files to a single documentation page.
> > > Fixing this error would help a lot.
> 
> @SumithChandran hi, sorry for taking long. Did you find a solution? I [commented](https://github.com/docsifyjs/docsify/pull/1855#discussion_r1087479706) on a direction we should look at. We just need to handle all the URL parts correctly, f.e. get all the parts of the URL, do what we need with only the path part, then put everything back together properly.

@trusktr , no I haven't found anything solution.

trusktr:
> Any chance this will get done soon? I've actually never committed to a public repo, but this looks pretty straightforward. I'd be willing to try.

@zomgbre willing to take a stab at it? We haven't circled back yet

src/core/router/history/base.js
trusktr: @sy-records 

> I think the solution should determine the fully qualified URL (f.e. `anchor.href = partialUrlString; const fullyQualifiedUrl = anchor.href`) and then use a new URL object to get all the pieces and use them correctly.

Have a chance to try this out?

sy-records: Use `new URL`?

sy-records: In most cases, `path` is an Invalid URL, so I don't use it.

image

Issue #850, opened 2019-05-15
How to add images from root?


rilc0n:
For example, I have 2 folders in the root directory. There is a picture in the first folder (images).

The second folder (docs) contains the markdown file.

How do I insert an image into this markdown file that is located in the folder (images).

If I put it this way:

`![](/images/testimage.jpg)`

He tries to add it like this:

`/docs/images/testimage.jpg`

timaschew:
Use https://docsify.js.org/#/configuration?id=relativepath

rilc0n:
This changes the links but not the path to pictures. Am I doing something wrong?

timaschew:
Hmm, it’s maybe a bug

LemarkNova:
I confirm this is a bug. When you use 
`![](/img/img.png)`
it doesn't work, but when you use 
`<img src='/img/img.png'></img>`
it works

TianYongwei:
Hmm .. 
this bug is fixed?

LemarkNova:
Yes, now it works

TianYongwei:
Hey @LemarkNova , Thanks for reply.

I am confused about the img tag compile result.

my markdown content is :
```md
![my img](/myimg.png)
``` 

The docsify compiled result is :
```html
<img src="/foldername/subfoldername/myimg.png" alt.../>
```

my expected result is :
```html
<img src="/myimg.png" alt.../>
```
because my all static img file is just in a single folder.

TianYongwei:
I found that this 「long path」is not caused by "marked" —— docsify compile markdown file by it, 

It seems caused docsify ? 

something could be configed ?

LemarkNova:
No finally it isn't fixed, I remember that I just use a workaround to use image in markdown:
I use `../picturefolder/picture.jpg`. So the real problem is that docsify doesn't compile correctly the markdown image. I don't know how you compile the markdown, but the problem is in the compilation of the markdown like in your previous comment.

LemarkNova:
Does Docsify use the latest version of marked? But I just try with marked and it work.

TianYongwei:
Thanks for your help. I note the workaround way.

I do nothing about the compilation of markdown files, just do some steps from the official documentation.

I am not sure the version of marked, but the docsify I installed is the latest version.

ShusenTang:
> No finally it isn't fixed, I remember that I just use a workaround to use image in markdown:
[...]

Issue #415, opened 2018-03-06
Version 4.6.8 broke images (404)


augustogoncalves:
The following code for showing images is working fine on 4.6.7, but fails (404) on 4.6.8

`![](_media/account/activate_sub.png)`

QingWei-Li:
https://github.com/QingWei-Li/docsify/releases/tag/v4.6.8

Yes, this is a small change. The relative path of this picture is expected behavior like github.

yzhang921:
How to use absolute path link to image. After this change, looks like all the path works as related path, even a standard absolute path string start with "/" will be parser to relative path.

willin:
粉转黑

MisterFili:
I too would like to know more about this update. I noticed some .img files are loading and others are not. is this something you can further explain on the main docsify site?

QingWei-Li:
@yzhang921 Starting with `/` is the absolute path

MisterFili:
Thanks for responding to my previous question but I'm still having issues displaying images from outside of my current working directory. I've updated to the most recent build, running localy, set and unset `basePath`, and also forked the docsify site to see where I could be going wrong.

 If I place the file in the current working dir I can display the image. however, when I choose the path of `_imagefolder` which is not in my current working directory, the image does not render like so ![foo](bar.png)

Prior to the update `![foo](_media/bar.png)` this was working vs having to use `![foo](bar.png)` which means placing the image file in the current document file location. 

I've read over all the bug updates users have created and still couldn't understand the reasoning as to why my images are not showing up. if you have any suggestions or examples, do share. tried to duplicate [this](https://github.com/QingWei-Li/docsify/blame/master/docs/deploy.md#L15), your own image solution but failed to render image outside of current working dir.

knightknight:
I think this is a similar issue.

I thought my sub-directory example/_coverpage.md would show the image from the root _images/example/example-logo.png directory. I even tried using `../_images/example/example-logo.png`.

<pre>
docsify
├── _images
│   └── example
│       └── example-logo.png
└── example
    ├── _coverpage.md | ![example-logo](_images/example/example-logo.png)
    ├── _sidebar.md
    └── README.md
_coverpage.md
_sidebar.md
index.html
README.md
</pre>

But it didn't work as @QingWei-Li has pointed out;

> path is relative to the current document file

What I needed to do was have an _images directory inside my /example directory and load them from there.

<pre>
docsify
└── example
    ├── _images
    │   └── example
[...]

PR #1868, opened 2022-08-23
fix: render image in absolute path

dandrade-meli:
## **Summary**

This PR documents support for including images with absolute paths and fixes a small bug that made it impossible to use.
This avoids the need to use ../../ to get to the root path.
Enable to fix bugs #850, #415

Example of use:

```md
![example-logo](:_images/example/example-logo.png)
![example-logo](//_images/example/example-logo.png)
```

```md
docsify
├── _images
│   └── example
│       └── example-logo.png
└── example
    ├── _coverpage.md | ![example-logo](:_images/example/example-logo.png)
    ├── _sidebar.md
    └── README.md
_coverpage.md
_sidebar.md
index.html
README.md
```

in /core/render/compiler/image.js which is responsible for generating the html `img`, there is a check if the href is not absolute.

```js
if (!isAbsolutePath(href)) {
   url = getPath(contentBase, getParentPath(router.getCurrentPath()), href);
}
```

the `isAbsolutePath` function uses the following regex to do this check

```js
/(:|(\/{2}))/g
```

what this regex does:
if the url contains `:` or `//`, then this path is absolute.

but when we used this strategy, adding the `:` or `//`, the image was not rendered because in the generated img tag the path kept the `:` or `//`

```html
<img src=":_images/example/example-logo.png" data-origin=":_images/example/example-logo.png" alt="" title="">
<img src="//_images/example/example-logo.png" data-origin="//_images/example/example-logo.png" alt="" title="">
```

and to fix this bug I applied a simple replace to remove `//` and `:` by ``

```html
<img src="_images/example/example-logo.png" data-origin="//_images/example/example-logo.png" alt="" title="">
<img src="_images/example/example-logo.png" data-origin="//_images/example/example-logo.png" alt="" title="">
```

Bugfix
Feature
Docs

## **For any code change,**

- [ ] Yes
- [x] No

## **Related issue, if any:**

#850
#415

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/docs/helpers.md
+++ b/docs/helpers.md
@@ -95,6 +95,7 @@ Only when you set both the `routerMode: 'history'` and `externalLinkTarget: '_se
 
 ### Resizing
 
+
 ```md
 ![logo](https://docsify.js.org/_media/icon.svg ':size=WIDTHxHEIGHT')
 ![logo](https://docsify.js.org/_media/icon.svg ':size=50x100')
@@ -120,6 +121,19 @@ Only when you set both the `routerMode: 'history'` and `externalLinkTarget: '_se
 ```md
 ![logo](https://docsify.js.org/_media/icon.svg ':id=someCssId')
 ```
+
+### Absolute Path
+
+```md
+
+![logo](:docs/_media/icon.svg 'with :')
+![logo](//docs/_media/icon.svg 'with //')
+
+```
+
+![logo](:docs/_media/icon.svg 'with :')
+
+![logo](//docs/_media/icon.svg 'with //')
 
 ## Customise ID for headings
 
--- a/src/core/render/compiler/image.js
+++ b/src/core/render/compiler/image.js
@@ -3,7 +3,7 @@ import { isAbsolutePath, getPath, getParentPath } from '../../router/util';
 
 export const imageCompiler = ({ renderer, contentBase, router }) =>
   (renderer.image = (href, title, text) => {
-    let url = href;
+    let url = href.replace(/^\/+|^:/, '');
     let attrs = [];
 
     const { str, config } = getAndRemoveConfig(title);


Koooooo-7 (commented):
Hi @dandrade-meli  , thx for ur investigation and input for this !
This change seems trick to match the regex of absolute condition tho. 

I think maybe we should have a `config` such as `:absolute` to indicate an `absolute` path, it can get rid of our own `isAbsolutePath` check. and I could have a patch on it asap.
cc @docsifyjs/reviewers  wdyt.

jhildenbiddle:
Let's identify the behavior we want first without worrying about implementation details. I think we also need to consider if/how Docsify's [`relativePath`](https://docsify.js.org/#/configuration?id=relativepath) configuration option weighs into this discussion.

Given the following Docsify site structure:

```
[Root]
└── dir
    └── FILE.md
_sidebar.md
index.html
README.md
```

I believe Docsify's HTML output for `/dir/FILE.md` image URLs should be as follows:

| Markdown URL   | Proposed output<br>relativePath:false | Proposed output<br>relativePath:true | current output      |
|----------------|--------------------------------|-------------------------------|---------------------|
| /img.png       | /img.png                       | dir/img.png                   | /dir/img.png        |
| img.png        | img.png                        | dir/img.png                   | /dir/img.png        |
| ../img.png     | ../img.png                     | ../img.png                    | /dir/../img.png     |
| /dir/img.png   | /dir/img.png                   | dir/img.png                   | /dir/img.png        |
| ../dir/img.png | ../dir/img.png                 | dir/img.png                   | /dir/../dir/img.png |

The table can be a bit confusing but the rules required to match the proposed output are simple:

1. When `relativePath:false` (default), do not modify the markdown URL
2. When `relativePath:true`...
   - Make URLs relative to the file context *if* the URL resolves to target within the file context. If the URL resolves to a target outside of the file context, do not modify the URL
   - Output relative URLs instead of absolute URLs. This allows Docsify sites contained in subdirectories to function properly

Thoughts?

Koooooo-7:
> Let's identify the behavior we want first without worrying about implementation details. I think we also need to consider if/how Docsify's [`relativePath`](https://docsify.js.org/#/configuration?id=relativepath) configuration option weighs into this discussion.

Yea, If the `relativepath` config is the highest rule, what the users do on paths should follow this config, then things should work well. Unfortunately, users wanna out of the rule and they can manually set one or more specific paths to absolute.
So, about this solution is based on :
A. Force user abide by the rules of the game.
B. Change the rule and left a back door.

Personally, I think it is fine to give users more sugars and tricks if the change would not break a big thing.

dandrade-meli:
> > Vamos identificar o comportamento que queremos primeiro sem nos preocupar com detalhes de implementação. Acho que também precisamos considerar se/como a [`relativePath`](https://docsify.js.org/#/configuration?id=relativepath)opção de configuração do Docsify pesa nessa discussão.
> 
> Sim, se a `relativepath`configuração for a regra mais alta, o que os usuários fazem nos caminhos deve seguir essa configuração, então as coisas devem funcionar bem. Infelizmente, os usuários querem sair da regra e podem definir manualmente um ou mais caminhos específicos como absolutos. Então, sobre esta solução é baseada em: A. Forçar o usuário a cumprir as regras do jogo. B. Mude a regra e deixou uma porta dos fundos.
> 
> Pessoalmente, acho que é bom dar aos usuários mais açúcares e truques se a mudança não quebrar uma grande coisa.

dandrade-meli:
> > Let's identify the behavior we want first without worrying about implementation details. I think we also need to consider if/how Docsify's [`relativePath`](https://docsify.js.org/#/configuration?id=relativepath) configuration option weighs into this discussion.
> 
> Yea, If the `relativepath` config is the highest rule, what the users do on paths should follow this config, then things should work well. Unfortunately, users wanna out of the rule and they can manually set one or more specific paths to absolute. So, about this solution is based on : A. Force user abide by the rules of the game. B. Change the rule and left a back door.
> 
> Personally, I think it is fine to give users more sugars and tricks if the change would not break a big thing.

I believe the prefix `:absolute` idea adds the possibility without breaking the default behavior.

jhildenbiddle:
**TL;DR:**

- Review #1891
- Rename `:absolute` to `:basepath` and allow it to accept a value:

  ```markdown
  ![](image.png ':basepath=/assets/')
  ```

---

> I believe the prefix `:absolute` idea adds the possibility without breaking the default behavior.

It could, but...

1. Docsify's path handling is problematic and/or broken in multiple scenarios, not just this one. The issue this PR attempts to solve is just one symptom of larger problems that need to be addressed. 

   I don't mean that to sound critical of this PR as I appreciate the effort and recognize the need for a fix. My intention is only to call attention to the larger problems first so we can determine the best path forward.

   I've created a separate issue that explains the path-related issues in more detail and provides solutions here: #1891. It can be a lot to digest, but the summary is simply that if we make path handling 1) consistent and 2) behave as expected (based on how they work everywhere else) by default, additional configuration options and custom markdown syntax will not be needed by the overwhelming majority of users.

   For perspective, here is a table demonstrating the current path-related issues:

   CleanShot 2022-09-19 at 15 08 31@2x

2. Images are not the only resource that Docsify makes inaccessible via absolute path. The same issue affects embedded content (path is relative to the page route) and markdown pages in a nested site (path is relative to `index.html`). If we are going to add an `:absolute` attribute for image, we should do the same for other elements as well.

3. The preferred behavior for an `:absolute` attribute can change based on the scenario. For example, assuming `:absolute` means "relative to the top-level of the domain or directory hierarchy" works here:
   ```
   [Root]
   ├─ assets
   │  └─ image.png
   ├─ dir
   │  └─ page.md
   ├─ index.html
   └─ README.md
   ```
   ```markdown
   <!-- Works on README.md or page.md -->
   
   ![](/assets/image.png ':absolute')
   ```
   ```html
   <img src="/assets/image.png">
   ```
   But the same behavior does *not* work here:
   ```
   [Root]
   └─ docs
      ├─ assets
      │  └─ image.png
      ├─ dir
      │  └─ page.md
      ├─ index.html
      └─ README.md
   ```
   Users could add `/docs/` to their absolute path:
   ```markdown
   <!-- Works on README.md or page.md -->
   
   ![](/docs/assets/image.png ':absolute')
   ```
   ```html
   <img src="/docs/assets/image.png">
   ```
   But it is not unreasonable to assume or want "absolute" to mean "relative to the top-level of my site". In order to accomodate both scenarios, the attribute should accept a value. Renaming the attribute to `:basepath` makes sense to me since this is the purpose served by the attribute:
   ```markdown
   ![](image.png ':basepath=/assets/')
   ![](image.png ':basepath=/assets/docs')
   ```
   ```html
   <img src="/assets/image.png">
   <img src="/docs/assets/image.png">
   ```
   
> @Koooooo-7: Personally, I think it is fine to give users more sugars and tricks if the change would not break a big thing.

I believe the goal should be for Docsify to be as intuitive as possible. Docsify's path handling is not intuitive. This is the result of both bugs and questionable design decisions made long ago. These issues aren't obvious or common because they only present themselves is less-common scenarios (which is why they've survived this long), but when they do present themselves it's a mess.

FWIW, I have no issues adding configuration options or custom markdown syntax to to make Docsify more capable or enjoyable to use. I'm okay adding an `:absolute`-like feature so long as long as it makes sense in the big picture (i.e., after we fix the larger issues).

Koooooo-7:
close via #1891.
If we finish the path issue, we could reopen it if needs.

PR #1914, opened 2022-10-27
fix: Upgrade docsify from 4.12.4 to 4.13.0

sy-records:
Fix https://github.com/sy-records/docsify-nightly/actions/runs/3334514166/jobs/5518154822#step:4:155


Koooooo-7 (commented):
Hi @sy-records , This pr same to this without remove package dependencies.
PTAL #1915 .

PR #1933, opened 2022-12-06
fix: genIndex error for search

wangliang181230:
- [x] 1. url is wrong if multiple contents are on the same site, example: `https://xxxxx/docs/` and `https://xxxxx/blog/`
- [x] 2. when the `token.text` is undefined, `handlePostContent` is undefined too at line 217
https://github.com/docsifyjs/docsify/blob/develop/src/plugins/search/search.js#L217

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


src/plugins/search/search.js
```diff
@@ -277,9 +289,9 @@ export function init(config, vm) {
 
   const isExpired = localStorage.getItem(expireKey) < Date.now();
 
-  INDEXS = JSON.parse(localStorage.getItem(indexKey));
+  INDEXS = isExpired ? null : JSON.parse(localStorage.getItem(indexKey));
 
-  if (isExpired) {
+  if (!INDEXS) {
```
wangliang181230, line 294: When manually delete `docsify.search.index` and keep `docsify.search.expires`, this plugin can still operate normally

Koooooo-7: The user should know what he does when he gonna manually delete index.

src/plugins/search/search.js
```diff
@@ -98,6 +98,20 @@ export function genIndex(path, content = '', router, depth) {
...
+        flag = true;
+        if (pathname.endsWith('/')) {
+          pathname = pathname.substring(0, pathname.length - 1);
+        }
+        pathname = pathname.substring(0, pathname.lastIndexOf('/'));
+
+        slug = '#/' + slug.substring(5);
+      }
+      slug = pathname + (flag ? '/' : '') + slug;
+
```
Koooooo-7, line 114: What does those changes for ?  could you plz give a sample and add verify test cases.

src/plugins/search/search.js
```diff
@@ -128,9 +142,7 @@ export function genIndex(path, content = '', router, depth) {
         token.text = getTableData(token);
         token.text = getListData(token);
 
-        index[slug].body = index[slug].body
-          ? index[slug].body + token.text
-          : token.text;
```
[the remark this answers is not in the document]
Koooooo-7: WDYM?

src/plugins/search/search.js
wangliang181230: When my website has multiple paths, it can prevent 404 error.
For example:
`http://xxxxxx/A/`
`http://xxxxxx/B/`
When searching on page A and accessing page B, 404 will appear.

src/plugins/search/search.js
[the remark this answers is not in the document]
wangliang181230: When the `token.text` is `undefined`, it will make the `index[slug].body` also `undefined`.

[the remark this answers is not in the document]
wangliang181230: So i change the `token.text` to `token.text || ''`.

This code ``index[slug].body ? index[slug].body + token.text : ` ` is redundant, and it is not the cause of the bug.

src/plugins/search/search.js
Koooooo-7: > When my website has multiple paths, it can prevent 404 error. For example: `http://xxxxxx/A/` `http://xxxxxx/B/` When searching on page A and accessing page B, 404 will appear.

I see, maybe u can take a look on this `namespace` configs [here](https://docsify.js.org/#/plugins?id=full-text-search).

src/plugins/search/search.js
Koooooo-7: > Just to reduce a manual operation step during the test, haha.

If user wanna break it, he should break it.

src/plugins/search/search.js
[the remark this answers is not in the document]
Koooooo-7: nice point.

src/plugins/search/search.js
wangliang181230: Should I configure it like this?

```json
{
    search: [
        '/A',
        '/B'
    ]
}
```

src/plugins/search/search.js
wangliang181230: Then this change is not necessary, is it?
 Let me revert this code.

Koooooo-7: > Then this change is not necessary, is it? If not, I'll revert this code.

yep. personally, I think we do not need handle this case.

src/plugins/search/search.js
Koooooo-7: I guess you need distinguish those namespace for different paths (sites).
```js


      // Use different indexes for path prefixes (namespaces).
      // NOTE: Only works in 'auto' mode.
      //
      // When initialiazing an index, we look for the first path from the sidebar.
      // If it matches the prefix from the list, we switch to the corresponding index.
      pathNamespaces: ['/zh-cn', '/ru-ru', '/ru-ru/v1'],
```

src/plugins/search/search.js
wangliang181230: done

src/plugins/search/search.js
wangliang181230: Ok, I will try it.

wangliang181230:
PTAL

Koooooo-7 (approved):
LGTM.

sy-records (approved):

PR #1959, opened 2023-01-06
fix-themes-hyl

huangyulie:
## **Summary**
Fixed the dolphin .css no styling bug in the Chinese version

image
image

## **For any code change,**

- [X] Related documentation has been updated if needed

- [ ] Yes
- [X] No

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


sy-records:
Thank you for the PR.
The problem here is that the wrong option was added in the Chinese document, which was not in the English, and I re-did it

Issue #1803, opened 2022-05-19
external links don't work in history router mode


6ecuw:
#### Steps to reproduce

1. Set `routerMode` to `history`.
2. Create mailto link with `<my@mail.com>`, or any external link.
3. Click on rendered link.

#### What is current behaviour

Crashes with error:

```
SecurityError
Failed to execute 'replaceState' on 'History': A history state object with URL 'mailto:my@email.com' cannot be created in a document with origin 'https://kc32ry.csb.app' and URL 'https://kc32ry.csb.app/'.
```

#### What is the expected behaviour

Native behaviour

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: macOS Catalina, version 10.15.7
- Node.js version: 16.13.2
- npm/yarn version: npm v8.1.2
- Browser version: Google Chrome 100.0.4896.127
- Docsify version: 4.12.2
- Docsify plugins: -

#### Please create a reproducible sandbox 

[![Edit immutable-sunset-kc32ry](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/immutable-sunset-kc32ry)

Live example: https://docs.lume.io/

trusktr:
Indeed, non-docsify links (external links) are currently not working in history router mode. We need to fix this and add tests.

trusktr:
I found the issue. It happens when `externalLinkTarget: '_self'` is specified. Fix WIP

PR #1967, opened 2023-01-22
ref #1803: fix cross-origin links in history router mode

trusktr:
the crossOriginLinks option is no longer needed, remove misspelled "crossorgin" link helper, and instead detect cross origin links and don't handle them as internal

## **Summary**

External links are now navigable by default when `routerMode` is `history`, regardless of the `externalLinksTarget` value. 

The `crossOriginLinks` option was added as a way to force external links to not be handled as internal.

Instead, that feature has been removed, and now we use `isExternal` to check for external links, and avoid handling them as internal.

The value of an `<a>` tag's `target` is no longer used as our external link check.

  Bugfix
  Docs

## **For any code change,**

- [ ] Yes
- [x] No

The `crossOriginLinks` option is now ignored. I don't think this is breaking. I think people want external links to work properly, so at worst some people will have listed all their links in that array, but now they don't have to.

Thinking of tests to add...

## **Related issue, if any:**

fixes https://github.com/docsifyjs/docsify/issues/1803

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] IE (of course not!)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -140,19 +140,6 @@ window.$docsify = {
 };
 ```
 
-## crossOriginLinks
-
-- Type: `Array`
-
-When `routerMode: 'history'`, you may face cross-origin issues. See [#1379](https://github.com/docsifyjs/docsify/issues/1379).
-In Markdown content, there is a simple way to solve it: see extends Markdown syntax `Cross-Origin link` in [helpers](helpers.md).
-
-```js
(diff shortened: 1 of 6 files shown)


Koooooo-7 (approved):
LGTM !

sy-records (approved):

PR #1979, opened 2023-01-29
feat: collapse root sidsebar items

proke03:
## **For any code change,**

- [ ] Yes
- [ ] No

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -271,7 +271,9 @@ export class Compiler {
     let html = '';
 
     if (text) {
-      html = this.compile(text);
+      html = this.compile(text)
+        .replaceAll('<ul >', '<ul class="app-sub-sidebar" >')
+        .replace('<ul class="app-sub-sidebar" >', '<ul >');
     } else {
       for (let i = 0; i < toc.length; i++) {
         if (toc[i].ignoreSubHeading) {
--- a/src/core/event/sidebar.js
+++ b/src/core/event/sidebar.js
@@ -37,7 +37,7 @@ export function collapse(el) {
 
   dom.on(el, 'click', ({ target }) => {
     if (
-      target.nodeName === 'A' &&
+      (target.nodeName === 'A' || target.nodeName === 'P') &&
       target.nextSibling &&
       target.nextSibling.classList &&
       target.nextSibling.classList.contains('app-sub-sidebar')


src/core/render/compiler.js
```diff
@@ -271,7 +271,9 @@ export class Compiler {
     let html = '';
 
     if (text) {
-      html = this.compile(text);
+      html = this.compile(text)
+        .replaceAll('<ul >', '<ul class="app-sub-sidebar" >')
+        .replace('<ul class="app-sub-sidebar" >', '<ul >');
     } else {
```
Koooooo-7, line 277: what is this change for ?

proke03: text is the entire compiled html of the sidebar.
The first "ul" is an element that contains the entire sidebar menu. 
So I tried to give a class to the rest of the "ul" except the first "ul".
캡처_2023_02_05_19_16_53_1

Koooooo-7: > text is the entire compiled html of the sidebar. The first "ul" is an element that contains the entire sidebar menu. So I tried to give a class to the rest of the "ul" except the first "ul". 캡처_2023_02_05_19_16_53_1

I see, you gonna treat the parent sidebar node as the sub-sidebar of the root `ul`.
But the class `.app-sub-sidebar` you marked get the style change as well.

Before:
image

After:
image

Koooooo-7: Maybe you could marked those parent sidebar nodes with new style `app-sidebar` and refactor the `click` event function.

proke03: Implementing it with onClick seems to be a dual implementation method for one feature.

Koooooo-7 (commented):
LGTM.

trusktr (changes requested):
We should make the items seem clickable, with `cursor: pointer`, instead of the text selection  cursor.

We also need to test this with [docsify-themeable](https://jhildenbiddle.github.io/docsify-themeable/).

Can you please:

- ensure the top-level items seem like they are clickable (instead of seeming like selectable text)
- write a test
- add a new `docsify-themeable.html` file that uses docsify-themeable to ensure we didn't break docsify-themeable.

proke03:
@trusktr
The themeable file is an extension of docsify.
In themeable, style is specified.
So, in order for changes in docsify to be applied without changing themeable, the id and class values previously specified in themeable must not be applied elsewhere.
Is my understanding correct?

trusktr:
@proke03 basically the docsify-themeable CSS depends on the DOM of Docsify.

I think what we should do is in a totally separate pull request lets get a docsify-themeable deployment for n Vercel going, and then it will be available to all PR similar to our current Vercel deployment with vanilla Docsify.

PR #1994, opened 2023-02-21
chore: docsify-ignore config polish.

Koooooo-7:
## **Summary**

Update the option 2 to use `docsify-ignore`.

Docs

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

### DIFF
--- a/docs/more-pages.md
+++ b/docs/more-pages.md
@@ -134,4 +134,6 @@ To ignore all headers on a specific page, you can use `<!-- {docsify-ignore-all}
 This header won't appear in the sidebar table of contents.
 ```
 
-Both `{docsify-ignore}` and `{docsify-ignore-all}` will not be rendered on the page when used.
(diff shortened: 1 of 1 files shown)


sy-records (commented):
I don't think it needs to be modified.

Why do you want to modify here?

Koooooo-7:
> I don't think it needs to be modified.
> 
> Why do you want to modify here?

We have two ways to config the ignore setting.
both `<!-- {docsify-ignore-all} -->` and `{docsify-ignore-all}` works. we should mention them all here. and the `{docsify-ignore-all}` is useful when u do not wanna meet someone html escape issue.

sy-records:
That should add a more detailed description, now this change can be easily misinterpreted as a contextual error.

Koooooo-7:
> That should add a more detailed description, now this change can be easily misinterpreted as a contextual error.

yep, I update it.

sy-records (approved):

PR #2067, opened 2023-05-15
Deferred run

DarthSim:
## **Summary**

On the [imgproxy docs](https://docs.imgproxy.net/) site, we need to fetch imgproxy versions to configure Docsify. Since we need to send an HTTP request, we can't do it synchronously, so we need to postpone Docsify initialization.

This PR allows deferring Docsify initialization by setting the `window.DOCSIFY_DEFER` variable. It also adds the `window.runDocsify` function that can be used to initialize Docsify when its config is ready.

Feature

## **For any code change,**

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] IE


trusktr:
I think being able to load Docsify on demand is good. But I think this isn't the direction we should go.

We are recently moving all code to ESM format (PR is ready to go). Once we have that in place, we can make this easy to do:

```js
<script type=module>
  import {Docsify} from 'https://unpkg.com/docsify@5/lib/docsify.esm.js'

  new Docsify(container) // run this any time you want.
</script>
```

We're aiming to be more standards friendly and make it an element.

```html
<script type=module>
  import {Docsify, define} from 'https://unpkg.com/docsify@5/lib/docsify-element.js'
  define()

  const el = new Docsify()
  container.append(el)
</script>
```
or
```html
<script type=module>
  import {define} from 'https://unpkg.com/docsify@5/lib/docsify-element.js'
  define()

  const el = document.createElement('docsify-element') // name needs to be picked
  container.append(el)
</script>
```
or
```html
<docsify-element></docsify-element>

<script type=module>
  import {define} from 'https://unpkg.com/docsify@5/lib/docsify-element.js'
  define() // run this line when ready. Before then, the element will not be instantiated.
</script>
```

trusktr:
Thanks again @DarthSim for the idea! I will close this one, then once we merge ESM very soon let's revisit this. Totally agree it should be easy to control when Docsify loads.

PR #2130, opened 2023-07-13
chore: remove codesandbox

sy-records:
We have GitHub Action and Vercel, so we can eliminate the need for codesandbox.

And codesandbox limits the build memory, so it often fails.


Koooooo-7 (approved):

trusktr (approved):

trusktr:
This is good. The issue template can have a Codesandbox link that doesn't include a build, so we don't need this.

PR #2131, opened 2023-07-14
activate sidebar on resetEvents

andy-held:
## **Summary**

When no `fetch`is done, the active sidebar element is not updated when a link is clicked. This is changed in this commit by calling getAndActive in `$resetEvents`.

This is shown in the videos below. In both videos, the 'QuickStart' sidebar element was selected, then the 'Initialize' sub-sidebar child of 'Quickstart' and then the parent sidebar element, 'Quickstart' again.

### Current behavior

Clicking on a sub-sidebar, then clicking on the parent sidebar again results in no element being 'active' .

[Screencast from 14.07.2023 17:51:07.webm](https://github.com/docsifyjs/docsify/assets/6648759/b6fcf8ca-6212-4b56-9208-1edaab2f7bd3)

### Behavior after fix

Clicking on a sub-sidebar, then clicking on the parent sidebar again results in the sidebar element being 'active' .

[Screencast from 14.07.2023 17:50:45.webm](https://github.com/docsifyjs/docsify/assets/6648759/a0287877-c8ce-49d2-8b8f-ab260d9a13d8)

Bugfix

- [ ] Yes
- [x] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE


trusktr (changes requested):
Hmm, it is still funky that it unhighlights the top level item while scrolling (highlights the hidden child item).

Can you add a test for this?

Perhaps what we should do is highlight the parent items in lighter green, and the sub item in darker green, so that the whole *path* is highlighted.

sy-records:
Thanks for your contribution.

It should be fixed in V5.

Issue #559, opened 2018-06-29
Inconsistent ID Anchor Jumps (revisited)


mlschuh:
Related to #332 but with a reproduced project file.

Attached is a slim version of our project where we were able to reproduce it. [docsify-anchor-jump-incorrect.zip](https://github.com/docsifyjs/docsify/files/2150037/docsify-anchor-jump-incorrect.zip)

To Reproduce:

```
npm i docsify-cli -g
docsify serve docs
```

Go to http://localhost:3000/#/isaac-gui-noimages?id=user-pane and see that it appears close to the anchor. 
Go to http://localhost:3000/#/isaac-gui?id=user-pane and see that it appears far from the anchor. 

The only difference between the two pages is the images being included or not.

-Mitch

DixCar:
I have noticed similar problem. Some anchors are work correct with sidebar but some aren't.  


![ezgif com-video-to-gif](https://user-images.githubusercontent.com/32497506/42166374-b8c493c0-7e02-11e8-9bda-841d910c43f0.gif)

I'm using 'theme-simple-dark'

-Paw

silentHoo:
@mlschuh This is the same I'm struggeling with. Problem is that images are async loaded, but docsify doesn't wait until images are present. So if the image has not a given height, the images height is 0 when first visiting the page. If the image is loaded later, the offset between your anchor jump and the current scroll position is exactly the height of the image(s).

mlschuh:
@silentHoo That makes sense in my mind. I'll ask around to my web dev friends to see if there's a way to at least set some kind of (short) timeout on the anchor jump to at least give the browser a fighting chance of being closer. Tricky part in my mind is to make sure the user still has an impression that the page is loading during that timeout.

rizdaprasetya:
This is what I did: https://github.com/docsifyjs/docsify/issues/351#issuecomment-528318227

PR #2238, opened 2023-09-24
Fix auto-scroll undershooting correction location on initial page load

nrankin18:
## Summary

**Bug:** Navigating to an anchor link on docs with images causes the initial auto-scroll to "undershoot" the correct location. This is because the scroll location is calculated prior to the images loading, so the image heights aren't accounted for in the location calculation.

Fix: Wait for the document to finish loading and then scroll to the location. After scrolling to the location, call `el.scrollIntoView()` as a backup.

Additional Bug: Auto-scroll on an initial page load does not scroll the appropriate sidebar item into view. This is because `#highlight()` is blocked while the Tweezer animation is running and will not scroll the sidebar if the item is already active.

Fix: Call `#highlight()` after the Tweezer scroll animation is complete with a parameter to force the scroll.

## Related issue, if any:

https://github.com/docsifyjs/docsify/issues/351
https://github.com/docsifyjs/docsify/issues/559

  Bugfix

## For any code change,

No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [x] Edge

### DIFF
--- a/src/core/event/index.js
+++ b/src/core/event/index.js
@@ -53,7 +53,20 @@ export function Events(Base) {
     #enableScrollEvent = true;
     #coverHeight = 0;
 
-    #scrollTo(el, offset = 0) {
+    #delayScrollInterval;
+    async #scrollTo(el, offset = 0) {
+      if (document.readyState !== 'complete') {
+        clearInterval(this.#delayScrollInterval);
+        await new Promise(resolve => {
+          this.#delayScrollInterval = setInterval(() => {
+            if (document.readyState === 'complete') {
+              clearInterval(this.#delayScrollInterval);
+              resolve();
+            }
+          }, 100);
+        });
+      }
+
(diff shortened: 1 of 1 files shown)


src/core/event/index.js
```diff
@@ -53,7 +53,20 @@ export function Events(Base) {
...
+        clearInterval(this.#delayScrollInterval);
+        await new Promise(resolve => {
+          this.#delayScrollInterval = setInterval(() => {
+            if (document.readyState === 'complete') {
+              clearInterval(this.#delayScrollInterval);
+              resolve();
+            }
+          }, 100);
+        });
+      }
```
trusktr, line 68: Is this reliable? Does the page go into non-complete ready state on each markdown fetch/render? If not, perhaps we should instead track images (etc) ourself, and wait for `load` event on all of them. Need to check this out.

PR #2253, opened 2023-10-07
feat: Add "Skip to main content" link and update nav behavior

jhildenbiddle:
## Summary

These updates align with accessibility best practices. The [skip link](https://webaim.org/techniques/skipnav/) and the updated navigation behavior allow for more efficient navigation via the keyboard and screen readers.

1. Add a "Skip to main content" button as the first element in the DOM with options. 
   - Includes configuration options to disable, modify the label, and specify localized labels based on the route path.
   - Dynamically updates label text on route path change (use "Translations" menu in preview for demo) 
2. Updated navigation behavior:
   - When loading a Docsify site:
     - If the URL does not contains a heading `id`, focus will remain on the `<body>` element. This is the browser's default behavior, which allows the skip link to appear when visitors press <kbd>Tab</kbd> one time.
     - If the URL contains a heading `id`, focus will be moved to the linked heading.
   - When navigating to a new page, focus will move to the first heading in the content area.
   - When navigating to a heading, focus will move to the linked heading.
3. Updated documentation to include `skipLink` configuration options

This bug also addresses a separate issue where caused Docsify to treat successfully loaded but empty markdown files as 404 errors.

#### Related research

- [Accessibility in Single Page Apps](https://johnsweetaccessibility.com/2020/05/accessibility-in-spas-part-2/)
- [Scroll to top: Where should the focus land?](https://www.digitala11y.com/scroll-to-top-where-should-the-focus-land/)
- [Recommendations for Single Page Applications](https://a11y-guidelines.orange.com/en/articles/single-page-app/)
- [Managing focus for accessibility](https://dev.to/robdodson/managing-focus-64l)
- [Accessible page titles in a Single Page App](https://hidde.blog/accessible-page-titles-in-a-single-page-app/)

## Screenshots

Safari 17 on macOS 14.0

CleanShot 2023-10-07 at 17 28 38@2x

CleanShot 2023-10-07 at 17 30 03@2x

## Related issue, if any:

#494
#495
#2294 (Fixes loading of empty markdown files)

- Feature
- Docs
- Tests

## For any code change:

No

- [x] Chrome
- [x] Firefox
- [x] Safari


docs/configuration.md
```diff
@@ -811,6 +811,39 @@ window.$docsify = {
...
 
+## skipLink
+
+- Type: `Boolean|String|Object`
+- Default: `'Skip to main content'`
+
+Determines if/how the site's [skip navigation link](https://webaim.org/techniques/skipnav/) will be rendered.
+
+```js
+// Render skip link or all routes (default)
```
trusktr, line 822:
```suggestion
// Render skip link for all routes (default)
```

src/core/event/index.js
```diff
@@ -18,14 +18,20 @@ export function Events(Base) {
 
       // If 'history', rely on the browser's scroll auto-restoration when going back or forward
       if (source !== 'history') {
```
trusktr, line 20: Does it work in history mode?

src/core/render/compiler.js
```diff
@@ -225,7 +225,7 @@ export class Compiler {
       nextToc.slug = url;
       _self.toc.push(nextToc);
 
-      return `<h${level} id="${slug}"><a href="${url}" data-id="${slug}" class="anchor"><span>${str}</span></a></h${level}>`;
+      return `<h${level} id="${slug}" tabindex="-1"><a href="${url}" data-id="${slug}" class="anchor"><span>${str}</span></a></h${level}>`;
```
trusktr, line 228: A comment to explain the tabindex would be nice.

src/core/render/index.js
```diff
@@ -242,6 +242,26 @@ export function Render(Base) {
...
+
+    #renderSkipLink(vm) {
+      const { skipLink } = vm.config;
+
+      if (skipLink !== false) {
+        const el = dom.getNode('#skip-to-content');
+        const text =
+          typeof skipLink === 'string'
+            ? skipLink
+            : skipLink?.[vm.route.path] || 'Skip to main content';
```
trusktr, line 254: This is a property access, but should it be a regex? Can this currently fail in some cases?

src/core/event/index.js
```diff
@@ -18,14 +18,20 @@ export function Events(Base) {
...
+          this.#scrollIntoView(this.route.path, this.route.query.id, true);
         }
         // Scroll to top if a link was clicked and auto2top is enabled
-        if (source === 'navigate') {
+        else if (source === 'navigate') {
+          focusEl = dom.find('.content');
           auto2top && this.#scroll2Top(auto2top);
         }
+
+        focusEl && focusEl.focus();
```
trusktr, line 34: This make every page navigation focus on content. Is it standard for `shift+tab` to be the way to navigate to navigation (sidebar)? What are some real-world examples?

test/integration/render.test.js
```diff
@@ -264,4 +278,61 @@ describe('render', function () {
...
+      expect(elm.textContent).toBe(expectText);
+    });
+
+    test('renders custom label from config object', async () => {
+      const expectText = 'test';
+
+      await docsifyInit({
+        config: {
+          skipLink: {
+            '/': expectText,
```
trusktr, line 317: Would it make be good to also test a path like `/es/`?

Also what if the site has no README, and the default landing is something like `/foo` and `/es/foo`?

trusktr (approved):

src/core/event/index.js
jhildenbiddle: It will work the same way it does on initial page load of a site in `hash` mode.

src/core/render/index.js
jhildenbiddle: Documentation here available in the preview: 

https://docsify-preview-git-494-skip-link-docsifyjs.vercel.app/#/configuration?id=skiplink

src/core/event/index.js
jhildenbiddle: > This makes every page navigation focus on content after the sidebar.

Having the focus move to the `<body>` element on each new page load would more closely mimic the behavior of a "standard" where clicking a link reload the entire page. This will allow users to quickly access the content area using the "Skip to main content" link or continue on to the navigation after each page load.

I will make this change.

> Is it standard for shift+tab to be the way to navigate to navigation (i.e. move backwards through the sidebar)?

Not to my knowledge. Regardless, the change described above will remove this concern.

test/integration/render.test.js
jhildenbiddle: Yep. Wil do.

src/core/event/index.js
jhildenbiddle: Confirmed working as expected.

FYI: In the highlighted code, `history` is referring to the type of navigation that is causing the URL to change (e.g., clicked link vs browser back/forward functionality) and not Docsify's [routerMode](https://docsify.js.org/#/configuration?id=routermode) configuration option.

src/core/render/compiler.js
jhildenbiddle: Done.

> // Note: tabindex="-1" allows programmatically focusing on heading
> // elements after navigation. This is preferred over focusing on the link
> // within the heading because it matches the focus behavior of screen
> // readers when navigating content.

Specifically, focusing on the heading instead of the link will cause screen readers to behave the same way they would when navigating page content. This is because screen readers will "focus" on heading elements even though they are not typically elements that can receive focus. Note that this only occurs when navigating content using screen reader navigation keys, not the Tab key (e.g., CapsLock + Up/Down). The result is that when focused on a heading, screen readers will describe the heading before describing the heading contents. When focused on a link within a heading, screen readers will describe the link and its content before the heading level.

- Focus on heading: "Heading level [level]. Visited. Link. [heading text]."
- Focus on link within heading: "Visited. Link. [heading text]. You are currently on a heading level [level]."

From a user perspective, it also makes more sense to click a link and land on a heading since this is standard behavior for anchor links. It is much less common and potentially confusing to click on a link and land on another link.

test/integration/render.test.js
jhildenbiddle: Done.

Koooooo-7 (commented):
Can not show the btn locally or on the `vercel Preview`, although the dom is insert success.
Is there any style issue?

jhildenbiddle:
@Koooooo-7 --

> Can not show the btn locally or on the `vercel Preview`, although the dom is insert success. Is there any style issue?

I have reviewed the Vercel preview and confirmed that it is working as intended.

The "Skip to main content" link is intended to be the first "focusable" element in the DOM. It is hidden by default and only visible when it receives focus. The intention is for it to be hidden for visitors navigating with a pointing device, but easily accessible to keyboard navigators and assistive device users by pressing <kbd>Tab</kbd> after the initial site load.

If you are wondering why pressing <kbd>Tab</kbd> does not result in the skip link being displayed after subsequent page loads, the following explanation taken from [this comment](https://github.com/docsifyjs/docsify/pull/2253#discussion_r1367958825) explains:

> For SPAs, there is no definitively "correct" behavior regarding where to move the focus after new content has loaded, although the majority opinion based on my research is that moving focus to the first heading in the content area is the most desirable behavior. The alternative is to set the focus back on the <body> to mimic the page load behavior of "standard" web site (i.e. non-SPA sites), but this is a far less practical behavior that requires users to manually navigate to the content area after every new page is loaded. Less than a minute testing with the keyboard or a screen reader makes choosing between these two options easy.

sy-records (approved):
Good jobs!

Koooooo-7:
Hi @jhildenbiddle , sorry for the misunderstanding here.
Hit  the `Tab`, I can see it now !

Based on our site here, about the `multi langs` support, I found that it only works when fresh the page ( as you mentioned, trigger after the `initial Load`),  so every time fresh pages gonna trigger the `skipLink` by `tab` as well .
My concern is when we fresh the page, or in multi langs,  we may already on the `main content`, it seems we don't need this btn.
Should we only make it works on the coverage showing?

jhildenbiddle:
> Hi @jhildenbiddle , sorry for the misunderstanding here. Hit the `Tab`, I can see it now !

No worries, @Koooooo-7. Happy to hear it's working as expected for you!

> My concern is when we fresh the page, or in multi langs, we may already on the `main content`, it seems we don't need this btn. Should we only make it works on the coverage showing?

👍 I've made a small update to the PR. Focus will now move to the linked heading when loading a URL with a linked heading `id` (e.g., `?id=foo`). This behavior mimics the standard hash behavior in standard web pages (e.g, `#foo`).

A complete summary of the new navigation behavior is available in the summary and below:

> 2. Updated navigation behavior:
>    - When loading a Docsify site:
>      - If the URL does not contains a heading `id`, focus will remain on the `<body>` element. This is the browser's default behavior, which allows the skip link to appear when visitors press <kbd>Tab</kbd> one time.
>      - If the URL contains a heading `id`, focus will be moved to the linked heading.
>    - When navigating to a new page, focus will move to the first heading in the content area.
>    - When navigating to a heading, focus will move to the linked heading.

Given the above behavior, I believe your concerns are addressed:

- If a user loads a Docsify URL that contains a heading `id`, focus will move to the linked heading. 
- If a user navigates to a page using the "Translations" menu on [docsify.js.org](https://docsify.js.org), focus will move to the first heading in the content area.

In both scenarios above, pressing <kbd>Tab</kbd> will not display the skip link.

Koooooo-7 (dismissed):
LGTM. thx @jhildenbiddle for you ideas and contributions. 👍

jhildenbiddle:
@sy-records @Koooooo-7 --

Would you mind approving this PR again? I had to merge `develop` which cleared your previous approvals. Thx!

Koooooo-7 (approved):

sy-records (approved):

Issue #2226, opened 2023-09-12
Accessibility - current page in sidebar should have aria-current="page"


kburk1997:
#### Problem or desire

Currently, screen readers such as VoiceOver, JAWS, or NVDA cannot differentiate between selected and non-selected links.

#### Proposal

Use `aria-current="page"` on the selected link, and leave non-selected links as-is

#### Implementation

DOM tree:
![Screenshot of DOM tree with active link highlighted. Active link has aria-current="page"](https://github.com/docsifyjs/docsify/assets/11298269/de61ce74-4236-4b3d-9b69-22fd81e9a0f2)

VoiceOver screen recordings:

Without aria-current:

https://github.com/docsifyjs/docsify/assets/11298269/873fdfe3-6a95-43d5-a417-7d34799b10b0

With aria-current:

https://github.com/docsifyjs/docsify/assets/11298269/87b43a78-8c36-4c63-a71c-9b0f880a1d1a

Reference: [MDN - aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current)

jhildenbiddle:
Thanks for creating the issue and for your patience white waiting for it to be addressed, @kburk1997.

I've created a draft PR here: #2254. In that PR you will find a preview link which will allow you to review the changes:

![CleanShot 2023-10-17 at 08 59 27@2x](https://github.com/docsifyjs/docsify/assets/442527/7e4bf479-b02c-46c3-8e98-a99ffefa1ae5)

Before I mark this PR as ready for review by the maintainers, I would love to get your thoughts and/or feedback.

Thanks!

Issue #658, opened 2018-10-29
Sidebar button does not have a description useful for screen reader users


andrapetronela:
Accessibility test errors

![screenshot 143](https://user-images.githubusercontent.com/34648710/47645961-5d1d0780-db6a-11e8-999d-15cddd9dbc51.png)

It would be nice to have an aria-label attribute that will help screen reader users understand what action will perform the button.
 Without an `aria-label` attribute:

- the sidebar button is identified by the screen readers only as a button element
- impossible to the user to know what action will perform the button, once clicked / selected

`aria-label` has no effect on browsers. It only helps assistive technology attach a label to an HTML element.

cheng-kang:
This is definitely a thing to be added. Would be best if you could make some suggestions about the labels 😄

andrapetronela:
- <button `aria-label="Toggle sidebar"`>
- <input `aria-label="Search through site content"`>

What do you think, @cheng-kang? :)

chris-79:
_Bump_

We _need_ this

jhildenbiddle:
Thanks for creating the issue and for your patience white waiting for it to be addressed, @andrapetronela.

I've created a draft PR here: #2253. In that PR you will find a preview link which will allow you to review the changes:

![CleanShot 2023-10-17 at 08 59 27@2x](https://github.com/docsifyjs/docsify/assets/442527/7e4bf479-b02c-46c3-8e98-a99ffefa1ae5)

Before I mark this PR as ready for review by the maintainers, I would love to get your thoughts and/or feedback.

Thanks!

PR #2254, opened 2023-10-09
Feat: Add aria attributes to sidebar toggle button and current page link

jhildenbiddle:
## Summary

These updates align with accessibility best practices.

1. Add aria attributes to sidebar toggle to indicate expanded/collapsed state
2. Add aria attributes to indicate current page

### Before

- No indication of current page in sidebar navigation
- Sidebar toggle does not indicate purpose or current state

https://github.com/docsifyjs/docsify/assets/442527/afa88f09-8d87-4486-96c6-aea6fac08af4

### After

- Current page is indicated in sidebar navigation
- Sidebar toggle indicated purpose and current state

https://github.com/docsifyjs/docsify/assets/442527/222f784e-561e-475b-9935-4acd225f61ea

#### macOS VoiceOver

[VoiceOver User Guide](https://support.apple.com/guide/voiceover/welcome/mac)

1. Turn VoiceOver on. ⌘ + F5
2. Open the Rotor CTRL + ⌥ + U
4. Use the ← and → arrow keys until you reach the "Landmark" list.
5. Navigate the list using the ↑ and ↓ arrow keys.
6. Press ↵ or SPACE to select to jump to the current item (rotor closes afterwards).

#### Windows Narrator

[Complete guide to Narrator](https://support.microsoft.com/en-us/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1)

1. Press the Windows key  + Ctrl + Enter together to start Narrator. Press these keys again to stop Narrator. On many keyboards, the Windows logo key is located on the bottom row of keys, to the left or right of the Alt key.
2. Press the Windows key  + Ctrl + N to open Narrator settings, and then turn on the toggle under Use Narrator.

## Related issue, if any:

#658
#2226 

- Feature

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge


src/core/event/index.js
```diff
@@ -217,7 +219,15 @@ export function Events(Base) {
      * @void
      */
     #btn(el) {
-      const toggle = _ => dom.body.classList.toggle('close');
+      const toggle = _ => {
+        dom.body.classList.toggle('close');
+
+        const isClosed = isMobile
+          ? dom.body.classList.contains('close')
+          : !dom.body.classList.contains('close');
```
trusktr, line 227: This is so confusing without going and learning how the class works.

src/core/render/index.js
```diff
@@ -251,29 +251,26 @@ export function Render(Base) {
...
-        [
-          document.querySelector('aside.sidebar'),
-          document.querySelector('button.sidebar-toggle'),
-        ]
-          .filter(e => !!e)
-          .forEach(node => node.parentNode.removeChild(node));
-        document.querySelector('section.content').style.right = 'unset';
-        document.querySelector('section.content').style.left = 'unset';
-        document.querySelector('section.content').style.position = 'relative';
-        document.querySelector('section.content').style.width = '100%';
```
trusktr, line 266: Why remove this?

src/core/render/index.js
```diff
@@ -251,29 +251,26 @@ export function Render(Base) {
...
+        sidebarEl?.parentNode.removeChild(sidebarEl);
+        sidebarToggleEl?.parentNode.removeChild(sidebarToggleEl);
+
         return null;
       }
 
       this._renderTo('.sidebar-nav', this.compiler.sidebar(text, maxLevel));
+      sidebarToggleEl.setAttribute('aria-expanded', !isMobile);
```
trusktr, line 265: When no sidebar, it is expanded on desktop?

src/core/render/tpl.js
```diff
@@ -37,12 +37,12 @@ export function main(config) {
   const name = config.name ? config.name : '';
 
   const aside = /* html */ `
-    <button class="sidebar-toggle" aria-label="Menu">
-      <div class="sidebar-toggle-button">
+    <button id="__sidebar-toggle" class="sidebar-toggle" aria-label="Toggle menu" aria-controls="__sidebar" aria-haspopup="true" aria-expanded="">
```
trusktr, line 40:
```suggestion
    <button id="__sidebar-toggle" class="sidebar-toggle" aria-label="Toggle menu" aria-controls="__sidebar" aria-haspopup="true" aria-expanded>
```

src/core/event/index.js
jhildenbiddle: @trusktr --

This application of the `close` class is confusing because the existing implementation (which predates all of the current maintainers AFAIK) is confusing. Refactoring the sidebar visibility logic is outside of the scope of this PR which is focused on adding `aria` labels for accessibility purposes.

FYI:

The `close` class means two opposite things based on the viewport width *when the page was loaded*. It is only applied to the `<body>` tag when the user clicks the sidebar button. Since the sidebar menu is hidden on mobile by default but displayed on desktop by default:

When a Docsify site loads, the lack of a `close` class on `<body>` indicates...
- On mobile the sidebar is hidden
- On desktop the sidebar is visible

When the sidebar toggle button is clicked and a `close` class is added to `<body>`, it indicates...
- On mobile the sidebar is visible
- On desktop the sidebar is hidden

Besides being confusing for maintainers, the pre-existing implementation also has the downside of inverting the user-specified sidebar visibility menu state when resizing the viewport between desktop and mobile widths. This is because `isMobile` is calculated only one time (when the page initially loads) instead of when the viewport size is adjusted. For example, if you load a page at a desktop resolution and hide the sidebar, resizing the viewport to a mobile resolution will cause the sidebar to appear.

We should clean this up, but that is a discussion to be had outside of this PR.

src/core/render/index.js
jhildenbiddle: @trusktr --

From what I can tell these lines are unnecessary. Sidebar menu behavior is the same on desktop and mobile for the live docsify site and the preview for this PR. My guess is that setting style properties directly on the `section.content` element was a holdover from an older implementation that never got cleaned up.

Is there something I'm missing?

src/core/render/index.js
jhildenbiddle: Correct. See my comment above: 

> The close class means two opposite things based on the viewport width when the page was loaded.

https://github.com/docsifyjs/docsify/pull/2254#discussion_r1367938848

src/core/render/tpl.js
jhildenbiddle: Better yet, I will remove the `aria-attribute` altogether and add comment explaining that this value will be set dynamically.

Per [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded):

> When a [menu](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role) is displayed, the button object that toggles the visibility of that menu has aria-expanded="true" set. When the menu is hidden, aria-expanded can be omitted. If specified when the menu is hidden, it should be set as aria-expanded="false". When a child menu is not visible, its parent [menuitem](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role) has aria-expanded. It should be set to true when the child menu is visible.

FWIW, I opted for `<button aria-expanded="">` in the source to indicate "no value" by default. The issue I have with `<button aria-expanded>` without a quotes (even if empty) is that those unfamiliar with this attribute will likely assume its existence without a value has meaning or indicates `true`, but this is incorrect.

Regardless, we can follow MDN's recommendation and still indicate that the `aria-expanded` attribute is applied dynamically to the `<button>` element via a comment.

jhildenbiddle: Placeholder `aria-expanded` placeholder has been removed.

Koooooo-7 (approved):

jhildenbiddle:
@Koooooo-7 --

Would you mind approving this PR again? I had to merge `develop` which cleared your previous approvals. Thx!

Koooooo-7 (approved):

sy-records (approved):

Issue #2225, opened 2023-09-12
Accessibility - hidden anchors are still tabbable


kburk1997:
#### Steps to reproduce

1. Go to [develop preview](https://docsify-preview.vercel.app/#/)
2. Press Tab key to move to next focusable element on the page
3. Repeat step 2 until "Edit document" link is focused
#### Current behaviour
Sidebar items remain in the tab order regardless of whether the menu is open or closed.
Screen recording with keystrokes:

https://github.com/docsifyjs/docsify/assets/11298269/1431c518-df88-4776-8d2e-a56b553f583f

#### Expected behaviour
Hidden sidebar items are not tabbable when closed. 
Sidebar items are tabbable when the sidebar is open.

#### Other relevant information

- Docsify version: latest

- [x] Bug still occurs when all/other plugins are disabled?

- Docsify plugins (if the bug happens when plugins enabled, please try to isolate the issue): n/s

jhildenbiddle:
Thanks for creating the issue, @kburk1997!

I've created a draft PR here: #2265. In that PR you will find a preview link which will allow you to review the changes:

![CleanShot 2023-10-17 at 08 59 27@2x](https://github.com/docsifyjs/docsify/assets/442527/7e4bf479-b02c-46c3-8e98-a99ffefa1ae5)

Before I mark this PR as ready for review by the maintainers, I would love to get your thoughts and/or feedback.

Thanks!

PR #2265, opened 2023-10-14
Feat: Prevent focus on hidden sidebar

jhildenbiddle:
## Summary

See title and original issue (below).

## Related issue, if any:

#2225 

Feature

## For any code change,

- No documentation changes required
- No automated tests required

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/src/themes/basic/_layout.styl
+++ b/src/themes/basic/_layout.styl
@@ -219,7 +219,7 @@ main.hidden
   top 0
   bottom 0
   left 0
-  transition transform 250ms ease-out
+  transition transform 250ms ease-out, visibility 250ms
   width $sidebar-width
   z-index 20
 
@@ -406,6 +406,7 @@ body.sticky
 
 body.close
   .sidebar
+    visibility hidden
     transform translateX(- $sidebar-width)
 
   .sidebar-toggle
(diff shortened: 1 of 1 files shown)


trusktr (approved):
Nice! Can it be tested?

jhildenbiddle:
> Nice! Can it be tested?

It can, but if we are going to test this it should probably be part of more thorough sidebar testing which (at minimum) verifies that the `close` class is being applied properly. I was hoping to avoid tackling that work in this PR given how simple this PR is as well as the following reasons:

1. Elements that are hidden not being unable to receive focus has been in the spec and reliable across browsers for ages.
2. The current sidebar implementation is kinda wonky and will require mores tests than if it was implemented properly. See https://github.com/docsifyjs/docsify/pull/2254#discussion_r1367938848 for details.
3. Since there is talk of refactoring/updating our method of rending HTML elements, I was hoping to postpone writing additional tests until this work begins.

That said, I don't mind implementing some additional sidebar tests, but I'd like to avoid this fix turning into a "test all the sidebar things" rabbit hole.

Koooooo-7 (approved):

sy-records (approved):

Issue #2278, opened 2023-10-16
Feature: navigation keyboard bindings/shortcuts


jhildenbiddle:
## Feature request

Improve accessibility by implementing keyboard bindings/shortcuts for common navigation actions.

Linked PR: #2279

#### Proposal

Provide the following navigation keyboard bindings/shortcuts for all docsify sites by default:

1. Toggle the sidebar menu
2. Move focus on the main content area
3. Move focus to the search field (and toggle sidebar if necessary)

Because all three of the above shortcuts are associated with DOM elements, keyboard binding hints can be provided to the user by hovering (via `title` attribute) or focusing (via `aria-keyshortcuts` attribute) on each associated element (sidebar toggle button, the "Skip to main content" element from #2253, and the search input field).

#### Optional

Allow custom keyboard binding functions to be defined as part of the docsify configuration. Docsify will handle capturing these keyboard shortcuts and invoking the associated function.

#### Example Implementation

```js
window.$docsify = {
  keyBindings: {
    // Example: Docsify default shortcut
    'toggleSidebar': {
      binding: 'alt+t',
      label: 'Toggle sidebar menu',
      callback(e) {
          const toggleElm = document.querySelector('.sidebar-toggle');

          if (toggleElm) {
            toggleElm.click();
            toggleElm.focus();
          }
      },
    },
    // Example: Plugin custom shortcut
    'myPlugin': {
      binding: 'alt+ctrl+shift+meta+a',
      label: 'Do a plugin thing',
      callback(e) {
        // ...
      },
    },
  }
}
```

Disable key bindings:

```js
window.$docsify = {
  keyBindings: false,
}
```

#### Links

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts

trusktr:
I've implemented this recently in an app. Custom key bindings don't need to be Docsify specific, and it can be a standalone library.

Docsify could have its own bindings made with it, and we can turn it on/off, but if someone want their own bindings they just use the lib:

```js
import {Actions} from 'foo-actions-lib'

new Actions({
  target: document, // default
  actions: {
    // ... similar type of config object ...
  }
})
```

And *then* if users want to map actions to Docsify functions, they can call `docsifyInstance.openMenu()`, `closeMenu()`, etc

If we implement it that way, we could the decide if it is really worth adding a $docsify config for actions, which would itself just be a tiny plugin to pass it along to a `new Actions`.


---

[...]

PR #2279, opened 2023-10-17
feat: Keyboard bindings

jhildenbiddle:
## Summary

- Add new `keyBindings` configuration option
- Add default docsify bindings
  - <kbd>\\</kbd> Toggle sidebar menu. Includes hover/focus hints via `title`/`aria-keyshortcuts` attributes.
  - <kbd>/</kbd> Focus on search field (also supports <kbd>Cmd</kbd> / <kbd>Ctrl</kbd> + <kbd>k</kbd>). Includes visual and hover/focus hints via `title`/`aria-keyshortcuts` attributes. This is a standard shortcut as seen on competitors like [Docusaurus](https://docusaurus.io), [GitBook](https://meshnet.nordvpn.com/), [Starlight](https://starlight.astro.build), and [VitePress](https://vitepress.dev).
- Add `keyBindings` documentation (see preview link below)
- Add e2e tests

## Related issue, if any:

#2278

Feature

## For any code change

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge


src/core/config.js
```diff
@@ -66,6 +66,39 @@ export default function (vm) {
...
+    config.keyBindings = Object.assign(
+      // Default
+      {
+        focusContent: {
+          bindings: "'",
+          callback(e) {
+            const containerElm = document.querySelector('.markdown-section');
+            const focusElm = containerElm.querySelector(
+              'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
+            );
```
Koooooo-7, line 80: Personally, I have a little confused about the `'` (Focus on content area) behavior.
It more like ` toTop` for me since it will scroll to the content top.
And the `\\` , which hide the sidebar more like the `Focus on content area` to me.

src/core/event/index.js
```diff
@@ -34,15 +34,82 @@ export function Events(Base) {
...
+
+          // Convert key sequences to arrays
+          // Ex: ['alt+t', 'ctrl+t'] => [['alt', 't'], ['ctrl', 't']]
+          bindingConfig.bindings = bindingConfig.bindings.map(keys => {
+            if (typeof keys === 'string') {
+              keys = keys.split('+');
+            }
+
+            return Array.isArray(keys)
+              ? keys.map(k => k.toLowerCase().trim())
```
Koooooo-7, line 73: Shall we sorted the keys here? otherwise, such as the `meta+i` and `i+meta` gonna be the different keyBind , but I think they should be the same.

src/core/config.js
jhildenbiddle: Fair point. I have removed the "Focus on content area" keybinding (<kbd>'</kbd>).

The original intention behind this keybinding was to allow visitors navigating with the keyboard to quickly move from different areas of the screen back to the content area. However, I've learned that screen readers offer their own methods of navigating between "landmark" elements such as the `<main>` (or in our case, `role="main"`) container, which arguably make a "focus on content area" shortcut unnecessary.

src/core/event/index.js
jhildenbiddle: > Shall we need pre check and sort the keys? Such as the meta+i and i+meta gonna be the different keyBind currently, although it will both be called , I think they should be the same.

What kind of pre-checking are you thinking of?

Sorting isn't necessary (key order doesn't matter), but it's not a bad idea to normalize the key bindings. I've updated the PR so that key combinations are now normalized by forcing all to lowercase and sorting alphabetically while ensuring modifiers come before non-modifier keys. For example:

- `alt+a` => `['alt', 'a']` (no change)
- `a+alt` => `['alt', 'a']` (modifier first)
- `a+shift+meta+ctrl+alt` => `['alt', 'ctrl', 'meta', 'shift', 'a']` (modifiers first & alpha sort)

> About the case sensitive conditions, such as I and i, we treat them are the same key now.
Should they be the same key or not?

The are case insensitive. The key bindings for lowercase and uppercase letters are different:

- Lowercase "i" = `"i"` or `"I"` 
- Uppercase "I" = `"shift+i"`, `"shift+I"`, `"Shift+I"`, `"SHIFT+I"`, etc.

Koooooo-7 (approved):
LGTM and thx a lot for your meticulous explanation and works !

jhildenbiddle:
@sy-records? @trusktr?

sy-records (approved):

PR #2288, opened 2023-10-21
chore: Create .gitattributes to prevent errors with Prettier in Windows

trusktr:
## Summary

It is recommended by Prettier to use LF endings for all files including in Windows, and they officially recommend to set this option in .gitattributes file for consistency, and LF is the default in Prettier 2 or higher. The lack of this file bit me in another project whose build failed in Windows in GitHub Actions due to `prettier --check` failing in Windows with default LF mode.

More info: https://github.com/prettier/prettier/issues/7825#issuecomment-602171740

  Build-related changes
  Repo settings

## For any code change,

No

N/A

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


trusktr:
Note, our build was not failing before this change, but without this change, some user's git configuration in Windows (apparently mine) could be different, and CRLF vs LF issues would pop up. Basically we've gotten lucky in our testing so far, and I don't normally dev in Windows so I happened to stumble on this when I tried.

Koooooo-7 (approved):

sy-records (approved):

Issue #2290, opened 2023-10-22
build/emoji.js fails in Windows


trusktr:
#### Steps to reproduce

```
D:\src\docsifyjs+docsify> node build/emoji.js 
Build emoji
- Fetching emoji data from https://api.github.com/emojis
- Retrieved 1877 emoji entries
- Error: Cannot read properties of null (reading '1')
- ```

#### Current behaviour

fail
#### Expected behaviour

pass

#### Other relevant information

- Docsify version: develop
- Your OS: Windows
- Node.js version: 20.5.0
- npm/yarn version: npm 10.2.0
- Browser version: N/A



#### Please create a reproducible sandbox

#### Mention the docsify version in which this bug was not present (if any)

**Comment by trusktr** [comment]
After making the change in

- https://github.com/docsifyjs/docsify/pull/2291

I can see the full error in the output, and the script no longer exits with 0 after the error (it should exit non-zero if there's an error):

```
PS D:\src\lume+lume\packages\docsifyjs+docsify> node .\build\emoji.js
Build emoji
- Fetching emoji data from https://api.github.com/emojis
- Retrieved 1877 emoji entries
file:///D:/src/lume+lume/packages/docsifyjs+docsify/build/emoji.js:47
  const emojiMarkdownStart = emojiMatch[1].trim();
                                       ^

TypeError: Cannot read properties of null (reading '1')
    at writeEmojiPage (file:///D:/src/lume+lume/packages/docsifyjs+docsify/build/emoji.js:47:40)
    at file:///D:/src/lume+lume/packages/docsifyjs+docsify/build/emoji.js:101:3
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.5.0
```

trusktr:
I updated the PR so it also fixes the Windows issue. I'm not sure exactly how the issue happens, but I think it might be due to git settings.

I think that the other PR,

- #2288 

may prevent the CRLFs from popping in unexpectedly, but no harm in making the regex more robust just in case the CRLFs get in there somehow.

jhildenbiddle:
@trusktr --

Can we close this now that #2288 and #2436 have been merged?

PR #2291, opened 2023-10-22
prevent a try catch from hiding a build error

trusktr:
## Summary

I removed the try-catch because it prevented the output from being helpful when there was an error, and made the script exit with zero instead of non-zero upon error.

## Related issue, if any:

https://github.com/docsifyjs/docsify/issues/2290

  Build-related changes
  Repo settings

## For any code change,

No

N/A

### DIFF
--- a/build/emoji.js
+++ b/build/emoji.js
@@ -95,13 +95,9 @@ function writeEmojiJS(emojiData) {
 
 console.info('Build emoji');
 
-try {
-  const emojiData = await getEmojiData();
+const emojiData = await getEmojiData();
 
-  if (emojiData) {
-    writeEmojiPage(emojiData);
-    writeEmojiJS(emojiData);
-  }
-} catch (err) {
-  console.warn(`- Error: ${err.message}`);
+if (emojiData) {
+  writeEmojiPage(emojiData);
+  writeEmojiJS(emojiData);
 }


Koooooo-7 (approved):

build/emoji.js
```diff
@@ -95,13 +96,9 @@ function writeEmojiJS(emojiData) {
 
 console.info('Build emoji');
 
-try {
```
jhildenbiddle, line 98: @trusktr --

`getEmojiData()` makes an API call to check if new emoji data is available and update Docsify *if necessary*. A failed API call does not warrant a broken build (it just means a check for new Emoji data won't occur). This is what the `try/catch` was in place for.

We can replace the `try/catch` with something else, but the failed API call needs to be handled gracefully.

jhildenbiddle (changes requested):

build/emoji.js
trusktr: Why should this succeedd if the API call fails?

If this exits zero despite it failing, then other scripts in a pipeline will happily continue along and an issue may go unnoticed.

jhildenbiddle: How about this as a simple solution:

1. Remove `build:emoji` from the `build` script in `package.json`
2. Add `build:emoji` to the `prepare` script in `package.json`
3. Force failed emoji API call to exit zero (as you suggest) in `build/emoji.js`

Answers to your questions below...

---

> Why should this succeedd if the API call fails? What's the reasoning?

The original goals were:

1. To prevent GitHub API availability issues or changes from breaking *all* local and automated builds.
2. To allow developers to build and test while offline--specifically, to run the `build` script.

Remember, the API call is not _required_ to complete a build. This is by design. We do not want our build to be reliant on a remote API that we do not control, which is why emoji data is stored in `src/core/render/emoji-data.js`. The API call simply checks to see if new data is available and, if so, updates the `emoji-data.js` file which is used to generate distributable files in `/lib/`. If the API call fails, the build will complete successfully using the data from `emoji-data.js`, which is the data stored in our repo and obtained from the last successful API call.

> It seems better to not have unintentionally incorrect builds.

Builds won't be "incorrect". Builds will contain "unchanged" emoji data because they will use local emoji data from `src/core/render/emoji-data.js`.

> What if, for example, we are publishing, and the emoji build fails, and the published package is not up to date because we thought it worked?

The result would be that the emoji data used in the newly published package would be identical to the data used in the previously published package.

You make a good point though. We should not rely on developers noticing an error message in the console to ensure our emoji updates are working as expected. See proposed solution at the top of this message.

jhildenbiddle (changes requested):
See reply here: https://github.com/docsifyjs/docsify/pull/2291#discussion_r1409665430

build/emoji.js
trusktr: > 3\. Force failed emoji API call to exit zero (as you suggest) in `build/emoji.js`

It was exiting zero, I made it exit *non-zero*. I don't believe any failure should be silenced.

Moving it to publish script (and publish script failing if the API fails) might be better.

> We do not want our build to be reliant on a remote API that we do not control

We're doing that with `npm install`, relying on some remote server. Same difference. If the build isn't working, *I want to know*. Up until now (thanks to Windows), I never knew I was possibly missing failed emoji-data build.

The API failure is intermittent, so it is also easy to click the button to re-run failed jobs. The main difference between that and publish script is someone will encounter the issue at a different point less often (PRs with builds happening more often than when we publish), but we still want to know of issues (just like we want to know if `npm install` failed).

trusktr: > Remember, the API call is not _required_ to complete a build.

Why is that? Will someone try to use an emoji from GitHub that they thought would also work in Docsify? If so, I'd say we don't want that, under the premise that the Docsify feature fully works.

---

Another idea is, if we really want to silence the failure of the API (and sometimes fail to support some emojis?) then we should catch that specific failure but not silence the whole script (we want other non-API failures to still fail).

jhildenbiddle: TL;DR:

I think we're getting our wires crossed but both intentions are good. I propose the following:

1. Update the console messaging to make it clear that an emoji "update" is unavailable instead of an emoji "build" failing. It's a small change, but I can see how the current wording may have contributed to misunderstanding how and why the emoji build works the way it does.
2. Call the `build:emoji` script before updating the project version (currently handled in `build/release.sh`, called using the `pub` and `pub:next` npm scripts).
3. Modify `build:emoji` and `build/release.sh` to prevent a version bump from happening on `build:emoji` error or when an emoji update has occurred resulting in uncommitted changes to `/src/core/render/emoji-data.js`. The remedy is simply to commit these changes separately so we see the update in the changelog, then continue with the version bump as planned.

This moves us away from checking for emoji updates on every build (which should not break the build because people and APIs go offline) to checking only when Docsify is version bumping in preparation of a new release (which is probably okay since the dev initiating the version bump need to be online to push the new commit(s) and version tag).

FWIW...

Details on how and why the `build:emoji` script works the way it does are provided below. I'm sharing these details only because there appears to still be a misunderstanding about how and why our emoji data build/update system works the way it does. Feel free to skip if uninterested.

---

> > Remember, the API call is not _required_ to complete a build.
> 
> Why is that? Will someone try to use an emoji from GitHub that they thought would also work in Docsify but it is missing? If so, I'd say we don't want that, under the premise that the Docsify feature fully works.

No. This was explained in https://github.com/docsifyjs/docsify/pull/2291#discussion_r1409665430:

"The API call simply checks to see if new data is available and, if so, updates the emoji-data.js file which is used to generate distributable files in /lib/. **If the API call fails, the build will complete successfully using the data from emoji-data.js, which is the data stored in our repo and obtained from the last successful API call.**"

In other words, the worst case scenario of the GitHub emoji API call failing is that a check for and possible addition of new emoji data does not happen. Docsify's emoji support will be unchanged. No new functionality is gained, but no existing functionality is lost. You can review the existing emoji data here: [`/src/core/render/emoji-data.js`](https://github.com/docsifyjs/docsify/blob/develop/src/core/render/emoji-data.js)

> If we really want to silence the failure of the API (and sometimes fail to support some emojis?) let's catch that specific failure but not silence the whole script (we want other non-API failures to still fail).

The failure isn't silent. The following message is displayed when the request fails:

```shell
Build emoji
- Fetching emoji data from https://api.github.com/emojis
- Error: getaddrinfo ENOTFOUND api.github.com
```

If you think it would help clarify, we can change the wording to make it clear that an emoji check/upate failed rather than an emoji "build" failing.

> We're doing that with npm install, relying on some remote server.

Using your `npm install` example to demonstrate...

If you attempt to build Docsify without having run `npm install` previously, the build will fail because none of the project's dependencies are available locally. After you run `npm install` and have the project's dependencies available locally, you only need `npm install` to update the dependencies you previously installed. If you or the remote API/server go offline, you can run `npm install package@latest` and see a failure message, then continue building successfully. Sure, you didn't get the package update you wanted, but you can use the package version you have until you or the API are back online.

This is exactly what's happening with the `build:emoji` script.

Emoji data has already been fetched, transformed, stored in the repo, and installed on your dev machine when you checked out the project. If you or the remote GitHub emoji API/server go offline, you can run `npm run build:emoji` and see a failure message, then continue to build successfully. Sure, you didn't the updated emoji data you were hoping for, but you can use the emoji data version you have until you or the API are back online.

Koooooo-7: As per the discussion there , how about that I move the `build emoji` stuff to workflow? The changes:

1. Make the `build:emoji` throw Exceptions when it failed.
2. Remove the `build:emoji` from the `npm build`, that we don't fetch the emoji in every build, use the `docsify` maintained emojis by default / always.
3. Force run the `build:emoji` when we do release  (`merge to main`) and we need fix it when it is actually failed. Because we should ensure it should be the latest version sync of github at the release date.
4. An `rotate emojiData` workflow, running every weekend （or one day?）it looks like the `dependabot`, if find the changes, raise PR to `develop`. i.e.
  ```
  update(emoji): sync emojiData and find update emojis in [2024-05-20]
  ```

After the changes:
1. We can make sure the release won't missing the emoji update, it should sync the latest version in release date.
2. We don't need the emoji update in every `build`. The largest out of sync date of `develop` emoji is `one week` (or one day, depends on the schedule), IMO, it is acceptable.
3. The `build: emoji` allows users to manually update the emojis if they really want some new released emojis within a weekend (or one day). instead, they need fix the exceptions by themselves (timeout, or something).

@trusktr , @jhildenbiddle  Thoughts?

jhildenbiddle: @Koooooo-7 --

Why not just do this part daily:

> 4. A rotate emojiData workflow, running every weekend (or one day?) it looks like the dependabot, if find the changes, raise PR to develop.

The PR will kick off a test run to verify that all [emoji tests](https://github.com/docsifyjs/docsify/blob/develop/test/integration/emoji.test.js) pass and we'll have an opportunity to review the changes before they are merged.

A daily run means we don't have to tie emoji checks to a version bump, publishing a new release, merging `develop` to `main`, etc. Yes, this approach means it's possible that a new release be published without the latest updates, but it's highly unlikely. The only two scenarios that would cause that to happen are if we publish a new release when 1) new emoji data is available but our daily emoji update workflow has not run or 2) we ignore an existing emoji update PR.

That works for me. @trusktr? @sy-records?

trusktr: > This is exactly what's happening with the `build:emoji` script.

The main reason that raised this issue is that the script failed for a reason _unrelated to the emoji API_, because _all errors_ in the file were supressed. Just want to make this part obvious because it wasn't covered in your reply.

**(Side note, the API sometimes fails even if we have internet, due to an API error from GitHub)**

> Builds won't be "incorrect". Builds will contain "unchanged" emoji data because they will use local emoji data from `src/core/render/emoji-data.js`.


I understand that the emoji-data will be unchanged if the script fails. But if GitHub introduces a new emoji `:whatever:`, and if a user tries to use that new `:whatever:` emoji but Docsify doesn't render the emoji because the URL is not in that data file, and we release a Docsify update, then the feature could be considered to be partially not working as expected (that's *ok*, but not *the best*).

Here in this file:

https://github.com/docsifyjs/docsify/blob/8ef9a3cd036bc86278709199e1da19da36f04ed1/src/core/render/emojify.js

We can see that it will only process emojis that exist in the emoji-data list. We can also see that Docsify is relying on GitHub as source of truth for names for native emojis based on unicode characters that the API sends back.

(On a related note, if GitHub removes an emoji (unlikely I imagine?) we might also want to show a warning if someone is still using it.)

Based on this, and on

- #2431 (👍 )

perhaps we should

- move emoji stuff back into the deprecated 
[emoji plugin](https://github.com/docsifyjs/docsify/blob/8ef9a3cd036bc86278709199e1da19da36f04ed1/src/plugins/emoji.js) lol.
- run the "build" on each publish to ensure that GitHub (and named native) emojis align with the release
  - without blocking script failure

The only people running the release is one or two of us, so I think it is totally fine if it fails and we try again. We can automate it later if really needed but I don't think it is necessary.

Koooooo-7: > > This is exactly what's happening with the `build:emoji` script.

> **(Side note, the API sometimes fails even if we have internet, due to an API error from GitHub)**
> 
> > Builds won't be "incorrect". Builds will contain "unchanged" emoji data because they will use local emoji data from `src/core/render/emoji-data.js`.
> 
> I understand that the emoji-data will be unchanged if the script fails. But if GitHub introduces a new emoji `:whatever:`, and if a user tries to use that new `:whatever:` emoji but Docsify doesn't render the emoji because the URL is not in that data file, and we release a Docsify update, then the feature could be considered to be partially not working as expected (that's _ok_, but not _the best_).


If we choose to maintain the emojis in our side and sync it from github in every release, it always has the gap that we may miss some latest updated emojis until we do the next release. If we need keep the emoji in real time, we need go back to the beginning solution that concat the emoji directly... 😅 


>move emoji stuff back into the deprecated [emoji plugin](https://github.com/docsifyjs/docsify/blob/8ef9a3cd036bc86278709199e1da19da36f04ed1/src/plugins/emoji.js) lol.
>run the "build" on each publish to ensure that GitHub (and named native) emojis align with the release
without blocking script failure
>The only people running the release is one or two of us, so I think it is totally fine if it fails and we try again. We can automate it later if really needed but I don't think it is necessary.


I think the `emoji plugin` solution is good.
Base on the  [emoji plugin](https://github.com/docsifyjs/docsify/blob/8ef9a3cd036bc86278709199e1da19da36f04ed1/src/plugins/emoji.js) , we could support `mount` the emojiData instead of keeping it inside (bind the release of docsify itself). Then, if we rotate the `emojiData` in the `emoji plugin` on time and do the plugin release asap, it should be more likely up-to-date.
But if users use something like `emoji-plugin@v5`, it goes back to the same issue tho...

jhildenbiddle: We need to wrap this up. Way too much time has been spent on this seven month old , arguably low-priority ~10 line change that could be spent on far more impactful PRs. 

Moving to an emoji plugin as described in #2431 is not an option for v5. This is why I added "This is a post-v5 feature request (see Milestone)" to the top of the issue when I created it earlier today. v5 needs to be v4 compatible (minus legacy browser support) for all of the reasons we have discussed previously. We can explore #2431 and other breaking options after v5 is released.

The solution proposed in https://github.com/docsifyjs/docsify/pull/2291#discussion_r1606893877 addresses the issue described in this PR and all of the concerns shared above. It sounds like @Koooooo-7 is willing to work on it (correct?). If there is a better solution, let's hear it. If not, let's proceed with automating emoji updates via a daily workflow so we all can move on to other tasks.

Koooooo-7: I think I may raise a PR for the daily upgrade auto check workflow first, and if we all agree to do it so. we can merge it and set it up to work.

jhildenbiddle:
@trusktr --

Can we close this now that #2288a and #2436 have been merged?

Issue #1929, opened 2022-11-20
Relative path link with ':ignore' dosen't work


aKuad:
## Bug Report

As far as I know, there are no same issues on here, no same questions on StackOverflow.
Sorry if already exist same one.

#### Steps to reproduce

0. File tree will be like this:

```
docs/
  |- index.html
  |- page/
        |- any_doc.md
        |- any_file.pdf
```

1. Add a docsify config to `index.html`:

```
relativePath: true
```

2. Write `any_doc.md`:

```md
[Click to view PDF](./any_file.pdf ':ignore')
```

3. Execute docsify:

```sh
docsify serve docs/
```

4. Make any PDF file as `docs/page/any_file.pdf`

5. Access to `http://localhost:3000/#/page/any_doc`

#### What is current behaviour

The link follows `http://localhost:3000/#/page/./any_file.pdf`

It looks `':ignore'` (no compile option) and relative path resolving are not works.

#### What is the expected behaviour

The link follows `http://localhost:3000/page/any_file.pdf`

#### Other relevant information

- [x] Bug does still occur when all/other plugins are disabled?

- Your OS: Linux Mint 20.3 Uma
- Node.js version: 16.16.0
- npm/yarn version: 9.1.2
- Browser version: Brave 1.45.127 Chromium: 107.0.5304.110（Official Build）
- Docsify version: 4.13.0
- Docsify plugins: *none (running in default)*

#### Please create a reproducible sandbox

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/holy-frog-ol1642)

#### Mention the docsify version in which this bug was not present (if any)

---

## Supplement

I tried specifying absolute path like this:

```
[Click to view PDF](/page/any_file.pdf ':ignore')
```

It works on local environment. But on GitHub pages, not works.

On local environment:

follows: `http://localhost:3000/page/any_file.pdf`

On GitHub pages (github.io):

follows: `https://{user_name}.github.io/page/any_file.pdf`

should be: `https://{user_name}.github.io/{repo_name}/page/any_file.pdf`

aKuad:
I think, this part has some probrem...

In `/src/core/render/compiler/link.js`, Line 21:
(+ my guessing note)

```js
if (
  !isAbsolutePath(href) &&  // -> true
  !compilerClass._matchNotCompileLink(href) && // -> true
  !config.ignore            // -> false
) {
  /* in this case: this part will not run */
[...]

PR #2300, opened 2023-11-06
Better support for relative links (e.g. download links) for non-markdown files

MartenBE:
## Summary

Currently, using relative links to non-markdown files are broken and give a 404. This PR addresses this and should fix cases such as described in https://github.com/docsifyjs/docsify/issues/1929 and others.

## Related issue, if any:

https://github.com/docsifyjs/docsify/issues/1929

- Bugfix
- Refactor

## For any code change,

- [ ] Related documentation has been updated, if needed -> Not needed

No, although there has been some refactoring.

- [ ] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge

## Notes

Relative linked files without extension are still seen as markdown files: I don't know if this is intentional or not, but seems counterintuitive. Better would be to require to link to .md-files with an explicit mention of the .md-extension (e.g. `[test](test.md)` of `[test](./test.md)`.

### DIFF
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -231,6 +231,7 @@ export class Compiler {
     origin.code = highlightCodeCompiler({ renderer });
     origin.link = linkCompiler({
       renderer,
+      contentBase,
       router,
       linkTarget,
       linkRel,
--- a/src/core/render/compiler/link.js
+++ b/src/core/render/compiler/link.js
@@ -1,8 +1,12 @@
 import { getAndRemoveConfig } from '../utils.js';
-import { isAbsolutePath } from '../../router/util.js';
+import { isAbsolutePath, getPath, getParentPath } from '../../router/util.js';
+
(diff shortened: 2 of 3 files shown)


trusktr:
@MartenBE Hey can you add some non-.md files to the Docsify site so that we can manually test as well? (We can view the deployment for the PR for manual test). Once the PR is ready then we can delete them.

MartenBE:
I've added some test files, but the latest commits feel "off". It is rather unstable and sometimes seems to cause reload loops and weird behavior. Unfortunately, I am out of my league here as my understanding of docsify is rather superficial. I think the original commit (https://github.com/docsifyjs/docsify/pull/2300/commits/e594e39e8f6b18b892473846bbefb0cc9cabe522) was (although not as architectal "proper" as we would like) more robust, and also uses the same workings as image links.

MartenBE:
In the previous months, we have decided to go with MkDocs instead of Docsify due to this bug and other reasons. So I'll probably won't work on this any further.

Issue #2294, opened 2023-10-31
Feature: Add response status to `route` object and plugins


jhildenbiddle:
## Feature request

- Add response status to `route` object
- Make response status available to plugin hooks
- Identify and handle request errors properly (not just 404)

#### Problem or desire

1. Internally, Docsify does not know response status details about the current route beyond basic success/error. This makes it difficult for contributors to handle various request errors properly.
2. Externally, Docsify does not expose response status details to plugin authors. This makes it difficult fro plugin authors to handle various request errors properly.

As a result of the two issues above:

3. Docsify incorrectly treats successful requests for empty markdown files as 404 errors. Docsify should handle empty pages properly and display an empty page as expected.
4. Docsify assumes all failed requests are 404 errors and displays "404 - Not found" in the content area. This is misleading when requests fail for reasons other than a 404 error. Docsify should be able to identify and display all request error codes and status messages.

#### Proposal

Current `route` object:

```js
{
  file: "README.md",
  path: "/",
  query: {}
}
```

Proposed `route` object:

```js
// Success
{
  file: "README.md",
  path: "/",
  query: {},
  response: {
    ok: true,
    status: 200,
    statusText: "OK"
  }
}

// Failure
{
  file: "README.md",
  path: "/",
  query: {},
  response: {
    ok: false,
    status: 404,
    statusText: "Not Found"
  }
}
```

Plugin access:

```js
window.$docsfy = {
  // ...
  plugins: [
    function(hook, vm) {
      hook.beforeEach(html => {
        console.log(vm.route.response.ok); // true
        console.log(vm.route.response.status); // 200
        console.log(vm.route.response.statusText); // "OK"
      }
    }
  ],
};
```

#### Implementation

See above.

PR #2303, opened 2023-11-14
Feat: Improved error handling and response status availability

jhildenbiddle:
## Summary

- Add response status to internal route object
- Expose response status to plugin hooks to allow for custom error handling.
- Update default error page content with error status and status text (e.g. "401 - Not Authorized", "500 - Internal Server Error", etc.) instead of displaying "404 - Not Found" for *all* errors.
- Fix issue where initial site render was incomplete when content fetch failed (e.g., no sidebar, plugins halted, etc.) 
- Fix issue where empty markdown pages/routes were handled as 404 errors.
- Add tests to verify default error handling
- Add tests to verify `notFoundPage` error handling
- Add tests to verify response data availability via plugins

See #2294 for details.

### Custom error handling via plugins (example)

CleanShot 2023-11-16 at 21 57 43@2x

```js
window.$docsify = {
  // ...
  plugins: [
    function (hook, vm) {
      hook.beforeEach(html => {
        const { file, response } = vm.route;
        const { ok, status, statusText } = response;

        if (ok === false) {
          return [
            `# Oops!`,
            `A **${status} - ${statusText}** error occurred while requesting **${vm.route.file}**`,
          ].join('\n\n');
        }
      });
    },
  ],
}
```

---

### Default error content updates (500 error)

**Before:** Displays inaccurate error information (shows 404 but should be 500)

CleanShot 2023-11-16 at 19 03 25@2x

**After:** Displays accurate error information

CleanShot 2023-11-16 at 19 03 32@2x

---

### Initial render with content load error

**Before:** No sidebar, plugins halted

CleanShot 2023-11-16 at 18 22 49@2x

**After:** Sidebar rendered, plugins complete

CleanShot 2023-11-16 at 18 22 56@2x

---

### Rendering empty files/routes

**Before:** Empty files incorrectly treated as 404 errors

CleanShot 2023-11-16 at 19 10 49@2x

**After:** Empty files rendered properly

CleanShot 2023-11-16 at 19 11 01@2x

---

## Related issue, if any:

#2294

Feature

## For any code change,

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

### DIFF
--- a/src/core/fetch/index.js
+++ b/src/core/fetch/index.js
@@ -102,7 +102,8 @@ export function Fetch(Base) {
         this.isHTML = /\.html$/g.test(file);
 
         // create a handler that should be called if content was fetched successfully
-        const contentFetched = (text, opt) => {
+        const contentFetched = (text, opt, response) => {
+          this.route.response = response;
           this._renderMain(
             text,
             opt,
@@ -111,7 +112,8 @@ export function Fetch(Base) {
         };
 
         // and a handler that is called if content failed to fetch
-        const contentFailedToFetch = _error => {
+        const contentFailedToFetch = (_error, response) => {
+          this.route.response = response;
           this._fetchFallbackPage(path, qs, cb) || this._fetch404(file, qs, cb);
         };
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -534,7 +534,7 @@ To disable emoji parsing of individual shorthand codes, replace `:` characters w
 - Type: `Boolean` | `String` | `Object`
 - Default: `false`
 
-Display default "404 - Not found" message:
+Display default "404 - Not Found" message:
 
 ```js
 window.$docsify = {
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -30,7 +30,7 @@ export default function (vm) {
       nativeEmoji: false,
       noCompileLinks: [],
       noEmoji: false,
-      notFoundPage: true,
+      notFoundPage: false,
       plugins: [],
       relativePath: false,
       repo: '',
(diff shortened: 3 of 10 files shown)


sy-records (approved):
LGTM.

src/core/fetch/index.js
```diff
@@ -111,7 +112,8 @@ export function Fetch(Base) {
         };
 
         // and a handler that is called if content failed to fetch
-        const contentFailedToFetch = _error => {
+        const contentFailedToFetch = (_error, response) => {
+          this.route.response = response;
           this._fetchFallbackPage(path, qs, cb) || this._fetch404(file, qs, cb);
```
Koooooo-7, line 117: we have a ` this._fetch404` to load custom pages here, via config [notfoundpage](https://docsify.js.org/#/configuration?id=notfoundpage).

Shall we keep return 404 pages behaviors, no matter what response status is?
It seem inconsistent behavior when we wanna distinguish response status codes in changes.
Besides, somehow it seems broke the custom 404 page function either when I test this PR locally.

jhildenbiddle: @Koooooo-7 @sy-records --

Good catch. This has been fixed. 

The `notFoundPage` configuration option is now handled properly, however it now only applies to 404 errors instead of *all* errors (as it did previously). This is (IMHO) the right thing to do, as both the option name and our documentation indicates that this option *is intended* to only apply to 404 errors.

All non-404 errors will render the default error content which displays the error status and status text (see screenshot above). Users who want to render error-specific content can now do so by checking the `vm.route.response` data and returning custom content via a plugin (see screenshot above).

sy-records (approved):

src/core/fetch/index.js
sy-records: Good job! Thank you!

Koooooo-7 (approved):
LGTM, that really a customize handler hook enhancement.

Issue #2218, opened 2023-09-09
switch from live-server to five-server


trusktr:
I've been having issues with `live-server` lately (in other projects): it is sometimes flaky, and sometimes the network connection is really slow, especially in **Windows PowerShell**. `live-server` hasn't been updated in a while, so it seems that something has changed in my system that somehow affects `live-server`.

There's a newer package called `five-server` (a **f**ork of l**ive-server**) that is a drop-in replacement for it, plus new features, and it has been working great without any issues.

Koooooo-7:
Hi Joe,  you mean this one https://github.com/yandeu/five-server ?
It seems not very active for now. I suppose that we could hold migrate to it later when it is stable.

jhildenbiddle:
My vote would be to replace live-server with [BrowserSync](https://browsersync.io) since we are already using it for our test environment. It supports all of the niceties of `live-server` and then some. It is also actively maintained.

Issue #2104, opened 2023-06-25
Simplify and modernize Docsify


trusktr:
We have some baggage, 

- support for CommonJS build from ES Modules, and code that is CommonJS
- old code to support ancient browsers like IE 11, which is _dead_
- experimental and incomplete SSR that no one has had been interested in or had time for finishing, and some of us believe there is a better way forward
- old `.styl` syntax for our CSS. CSS has lots of new language features, including CSS Nesting which is out in all browser (Firefox with a flag), container queries, etc. 

I propose:

- [ ] delete SSR
- [ ] convert to ES Modules and delete all traces of CommonJS
- [ ] modernize: remove polyfills, use modern features 
- [ ] delete Stylus, write vanilla CSS (using postcss for now only until CSS Nesting lands in Firefox).
- [ ] Possibly convert to TypeScript. Have the best intellisense and self-documented code.

jhildenbiddle:
This is a good list. 

I think it's important to recognize that while this type of cleanup work will benefit maintainers, the end result of the items listed above do not provide a compelling reason for existing users to upgrade from v4. This isn't meant as a negative comment. I think Docsify *needs* this work to be done and that it will be in a much better place when this work is completed. My intention is only to point out that if we are going to introduce changes that will require a major version bump, we should also consider changes that directly benefit users so that they are incentivized to upgrade.

One suggestion is to rewrite the HTML and CSS generated by Docsify as part of the v5 release. If proposals for v5 include converting all of our CSS from Stylus to vanilla CSS and converting our HTML from string/templates to components then this would be an opportune time to do it. By incorporating lessons learned from docsify-themeable and popular component libraries we can offer an updated UI with flexible theming options. Since this change would be highly visible, I believe this change alone would motivate users to upgrade.

A few quick thoughts regarding other items on the list above:

1. I think we should consider adding [Babel](https://babeljs.io) to our Rollup configuration along with a [browserslist](https://browsersl.ist) configuration that specifies that last "X" version of evergreen browsers ("X" is TBD, perhaps 3?). This ensures our distributable work as expected on multiple versions of evergreen browsers, which is important considering that our test environment tests only against the *latest* versions of the major browser engines (Blink, Gecko, and WebKit). We can still write modern JS and the expectation is that Babel will transpile very little (if any) code during the build process, but it is there as a safeguard in case a dev uses a new JS feature that would otherwise break in an evergreen browser 1 or 2 versions behind the latest version.

[...]

PR #2316, opened 2023-12-02
Chore: Replace live-server, clean up server implementation, and update test docs

jhildenbiddle:
## Summary

This PR cleans up our serve configuration as discussed in #2104 and #2218:

- Standardize on a single project server by replacing [live-server](https://www.npmjs.com/package/live-server) with [BrowserSync](https://browsersync.io) for development server
- Consolidate server configurations into single file for easier management (`server.config.js`)
- Update server configurations to allow multiple  servers to run simultaneously:
  - Development: 3000
  - Test: 4000
  - Production: 8080
- Update `serve` script to serve `/docs` in production mode
- Add `serve:dev` script to serve `/docs` in development mode
- Fix development rendering issue caused by `DocsifyCarbon.create` error.
- Add server port checking to prevent potential test/CI issues from BrowserSync's port auto port selection behavior
- Update test documentation in `/test/README.md`
- Remove `index.html` file in root (was confusing, now no longer needed)
- Remove live-server dependency

## Related issue, if any:

#2218

Other

## For any code change,

No


trusktr (approved):
Nice! 

I linked this to the todo item

> We have both /index.html and docs/index.html. Do we need duplicate code? Delete one.

in https://github.com/docsifyjs/docsify/issues/2104. That's the only item from that list that this handles right?

I also linked #2218 so that'll be closed when this is merged.

trusktr (changes requested):
Looks like the deployments are not working now (I'm guessing the index.html change) and needs an update so that we can verify it works.

jhildenbiddle:
Thanks for the quick review and approval, @trusktr!

> in https://github.com/docsifyjs/docsify/issues/2104. That's the only item from that list that this handles right?

Correct.

sy-records:
Okay. I got it.

Let me try.

jhildenbiddle:
I did a little research and it doesn't look like Vercel will allow us to serve files via our own server / script. 

We can use serverless functions to serve files with our own server (see [here](https://vercel.com/guides/using-express-with-vercel)), but the implementation feels like a hack and will force our previews to be served under an `/api` directory on Vercel. Not great.

Unless there are other suggestions, I think the easiest solution is create a separate `index.html` file for Vercel to serve, similar to what we had before but implemented in a more elegant way. For example, we could have a `build:preview` script that generates the `index.html` file and copies all necessary files to `.vercel/output/static` to be served as static files at `/` (see [here](https://vercel.com/docs/build-output-api/v3/primitives#static-files) for details). Just an idea.

I'm open to other suggestions, too. 😄

sy-records (dismissed):

jhildenbiddle:
All set!

Vercel preview deployments are working again. I've update the summary above to include the new changes. The changes made specifically to address the preview deployments are:

> - Add `build:html` script to generate `/docs/preview.html` from `/docs/index.html`
> - Add `vercel.json` file to handle preview redirect to `/docs/preview.html`
> - Remove duplicate `/themes` directory in root (no longer needed)

sy-records (approved):
Goog job!

Koooooo-7:
Hi @jhildenbiddle 
I think we have different thoughts about the `remove root index.html` change here. 😅 

I thought we gonna remove the duplicated `index.html`, then, we use the `/docs/index.html` with `local path imports` (`/lib/docsify.js`...etc) which could benifit us to  dev and test locally. 

About the pages and preview issues, we could enhance CI stuff to do adoptions, thats why I try to config  vercel and custom github pages flow.

---

In current changes, we delete the root `index.html` but generated a `/docs/preview.html` to support the vercel-preview and local dev.
I'm puzzled now.
Because the old `/index.html` can do it already, why should we remove it and instead create `preview.html`?
It just change to dynamic generate the `preview.index` at `dev mode` on every build instead of putting some `for dev index.html` at root page directly (the `/index.html`).

--- 
If we decide use this `"preview.html way"`, I think we could do some refine as well.
- Generate the `preview.html` in root path and named `index.html`. (literally, generate the old `/index.html`)
- Change the `docsify.min.js` to `docsify.js`. (good for debug)

After those updates, we don't need the `vercel.config` either.

Or we change to enhance it in `CI` and leave the  `/docs/index.html` with local import. ( I could do this part, if needs.) 

WDYT?

Koooooo-7 (approved):
Update:

Separate this PR as 2 parts:
1. Clean up server implementation. (live-server, remove `/index.html`)
2. How to adapt the deploy stuff. (preview, github pages)

In order to keep things move on, I think we could approve the changes  and left my questions, it could be discussed and do the refinements later.

Hi, @jhildenbiddle , plz go ahead and we can merge it asap.

jhildenbiddle:
Apologies for the slow response, @Koooooo-7. I got pulled away on some other projects.

> In order to keep things move on, I think we could approve the changes and left my questions, it could be discussed and do the refinements later.

Perfect. This is exactly what I was going to suggest. :)

PR #2317, opened 2023-12-03
action: refactor config vercel and pages publish.

Koooooo-7:
## Summary

Redirect output root path to `docs/index.html` since we remove the root `index.html` under #2316 .
If it is acceptable, we could merge it into #2316 and move on.

Update:
Custom github-pages publish flow to adapt `index.html` changes.

 Repo settings
## For any code change,

Non core functions related changes.

No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


sy-records (dismissed):
But it is not without preview. use cdn resources instead of lib.

Koooooo-7:
Hi @jhildenbiddle , I see. 
I think the problem is the default github pages flow won't  pre-build project, hence there is no `/lib` resources folder.
If we accept the `redirect way` to make the `vercel` works, I think I could re-write the gitpage publish with custom github action ( pre-build the `/lib` folder out ) to make the `/doc/index.html` works as well. 
I could have a try on it.

jhildenbiddle:
> If we accept the redirect way to make the vercel works, I think I could re-write the gitpage publish with custom github action ( pre-build the /lib folder out ) to make the /doc/index.html works as well.
Or instead, I rewrite the path to CDN path in actions.

If we do that, we're just introducing a different version of the previous `index.html` issue by having duplicate URL rewrite logic locally (BrowserSync) and in CI (GitHub actions).

A simpler way to accomplish the same thing:

1. Copy the `/lib/` folder to `/docs/` during the build process
2. Use local paths instead of CDN URLs for docsify files in `/docs/index.html`. 
3. Use the redirect as you've implemented in this PR.

I've explored a few options that solve the Vercel preview issue already. I think the above approach is best for a few reasons which I will outline in #2316 shortly. I will push these changes to that PR in the next few minutes.

docs/index.html
```diff
@@ -17,32 +17,18 @@
...
     <link
       rel="stylesheet"
-      href="//cdn.jsdelivr.net/npm/docsify@4/lib/themes/pure.css"
-      title="pure"
-      disabled
-    />
-    <link
-      rel="stylesheet"
-      href="//cdn.jsdelivr.net/npm/docsify@4/lib/themes/dolphin.css"
+      href="../themes/dolphin.css"
```
sy-records, line 31: why not `/lib/themes`

jhildenbiddle:
@Koooooo-7 --

I've addressed the Vercel deployment issue in #2316.

PR #2335, opened 2023-12-14
fix: missing themes for root

sy-records:
ref #2316, Remove duplicate `/themes` directory in root (no longer needed)

We actually need it because it's referenced in the documentation. see https://docsify.js.org/#/quickstart?id=specifying-docsify-versions

And it's uncompressed css, `lib/themes` is compressed.

I think it should be keep it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /docs/preview.html
 /lib
 /node_modules
+/themes
 
 # exceptions
 !.gitkeep
--- a/build/mincss.js
+++ b/build/mincss.js
@@ -7,7 +7,7 @@ const files = fs.readdirSync(path.resolve('lib/themes'));
 files.forEach(file => {
   file = path.resolve('lib/themes', file);
   cssnano
-    .process(fs.readFileSync(file))
+    .process(fs.readFileSync(file), {from: undefined})
     .then(result => {
       fs.writeFileSync(file, result.css);
     })
--- a/build/mincss.js
+++ b/build/mincss.js
@@ -7,7 +7,7 @@ const files = fs.readdirSync(path.resolve('lib/themes'));
 files.forEach(file => {
   file = path.resolve('lib/themes', file);
   cssnano
-    .process(fs.readFileSync(file), {from: undefined})
+    .process(fs.readFileSync(file), { from: undefined })
     .then(result => {
       fs.writeFileSync(file, result.css);
     })
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   ],
   "scripts": {
     "build:cover": "node build/cover.js",
-    "build:css:min": "node build/mincss.js",
-    "build:css": "mkdirp lib/themes && node build/css -o lib/themes",
+    "build:css:min": "mkdirp lib/themes && node build/css -o lib/themes && node build/mincss.js",
+    "build:css": "mkdirp themes && node build/css -o themes",
     "build:emoji": "node ./build/emoji.js",
     "build:html": "node ./build/html.js",
     "build:js": "cross-env NODE_ENV=production node build/build.js",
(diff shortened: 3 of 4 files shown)


jhildenbiddle:
> We actually need it because it's referenced in the documentation. see https://docsify.js.org/#/quickstart?id=specifying-docsify-versions

Good catch, however I do not think we should restore the `/themes` directory. 

Having duplicate files in multiple locations in a single repo is a bad idea. Doing so invites confusion among those unfamiliar with the details of why duplicate files exist, which can lead to public-facing mistakes that require long-term support. This is an excellent example, and I believe it is a mistake we should correct for v5:

1. The next release is going to be a breaking change because our codebase no longer support legacy browsers. If we want to make a change, this is the time to do it.
2. Our documentation should not recommend loading unminified CSS or JS files on a production site. Similarly, our CLI output should not load unminified CSS or JS by default. We can offer unminified CSS/JS as options, but our recommendation should be to use minified CSS/JS.
3. All distributable files belong in the same directory. For v4, this is `/lib`. For v5, the proposed location is `/dist`.
4. If we want to offer minified and unminified CSS files, we should differentiate using the filename and not the location, just as we do with `docsify.js` and`docsify.min.js`. For v5, this would mean something like `/dist/themes/vue.css` and `/dist/themes/vue.min.css`.

I planned to address this issue in an upcoming PR as part of our v4 => v5 transition:

1. Modify build to output distributable files to `/dist` instead of `/lib`. This is part of the v4 => v5 transition.
2. Update URLs as needed in `docsify` and `docsify-cli` repos to reference `/dist` instead of `/lib`.
3. Create static redirects/imports from `/lib` and `/themes` to `@4/lib` and `@4/themes` CDN URLs. This allows v4 sites without a URL version lock to continue working when they unintentionally load v5 after its release.
4. Add a console deprecation / upgrade notice for sites loading resources from `/lib` or `/themes` redirects/imports.

The static redirects/imports in `/lib` and `/themes` are a temporary solution. They exist only to prevent sites that load docsify resources without URL version locks from breaking when v5 releases because v4 was released in 2017 but our docs did not recommend URL version locks until mid-2020. My hope is that they will can remove them by the time v6 is released.

jhildenbiddle (commented):
> We actually need it because it's referenced in the documentation. see https://docsify.js.org/#/quickstart?id=specifying-docsify-versions

Good catch, however I do not think we should restore the `/themes` directory. 

Having duplicate files in multiple locations in a single repo is a bad idea. Doing so invites confusion among those unfamiliar with the details of why duplicate files exist, which can lead to public-facing mistakes that require long-term support. This is an excellent example, and I believe it is a mistake we should correct for v5:

1. The next release is going to be a breaking change because our codebase no longer support legacy browsers. If we want to make a change, this is the time to do it.
2. Our documentation should not recommend loading unminified CSS or JS files on a production site. Similarly, our CLI output should not load unminified CSS or JS by default. We can offer unminified CSS/JS as options, but our recommendation should be to use minified CSS/JS.
3. All distributable files belong in the same directory. For v4, this is `/lib`. For v5, the proposed location is `/dist`.
4. If we want to offer minified and unminified CSS files, we should differentiate using the filename and not the location, just as we do with `docsify.js` and`docsify.min.js`. For v5, this would mean something like `/dist/themes/vue.css` and `/dist/themes/vue.min.css`.

I planned to address this issue in an upcoming PR as part of our v4 => v5 transition:

1. Modify build to output distributable files to `/dist` instead of `/lib`. This is part of the v4 => v5 transition.
2. Update URLs as needed in `docsify` and `docsify-cli` repos to reference `/dist` instead of `/lib`.
3. Create static redirects/imports from `/lib` and `/themes` to `@4/lib` and `@4/themes` CDN URLs. This allows v4 sites without a URL version lock to continue working when they unintentionally load v5 after its release.
4. Add a console deprecation / upgrade notice for sites loading resources from `/lib` or `/themes` redirects/imports.

The static redirects/imports in `/lib` and `/themes` are a temporary solution. They exist only to prevent sites that load docsify resources without URL version locks from breaking when v5 releases because v4 was released in 2017 but our docs did not recommend URL version locks until mid-2020. My hope is that they will can remove them by the time v6 is released.

sy-records:
okay, I created a new issue. goto #2336

PR #2344, opened 2023-12-19
Update logo doc

nonam4:
## Summary
If name prop is not set logo is not visible, this should be described to user

## Related issue, if any:
#1646 

## What kind of change does this PR introduce?
Docs

## Does this PR introduce a breaking change?
No

Issue #1646, opened 2021-10-06
Sidebar logo not working


blaasvaer:
## Bug Report
This does nothing:
```
window.$docsify = {
  logo: '/_media/icon.svg',
};
```
I've tried all variations of:
` '/_media/icon.svg',`
` '_media/icon.svg',`
` '/media/icon.svg',`
` 'media/icon.svg',`

None make any difference. I can easily place the logo on the coverpage ... but using that same url with 'logo:' doesn't work either.

#### Steps to reproduce

Erhm, well change the path ...

#### What is current behaviour

Noting happens.

#### What is the expected behaviour

I should see a logo in the sidebar (I guess), as the docs describe.

#### Other relevant information

- [ ] Bug does still occur when all/other plugins are disabled?

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 
- Docsify plugins: 

#### Please create a reproducible sandbox 

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/307qqv236)

sy-records:
You can create a reproducible sandbox, or provide a repo

blaasvaer:
Tried a fresh sandbox: https://codesandbox.io/s/flamboyant-surf-bvvj5?file=/index.html

blaasvaer:
Simple make a check for an empty name somewhere in the build process … and handle accordingly.

PR #2347, opened 2023-12-20
docs: Update logo doc

nonam4:
## Summary
If name prop is not set logo is not visible, this should be described to user

## Related issue, if any:
#1646
Also PR #2344 

Docs

No

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -376,6 +376,8 @@ window.$docsify = {
 
 Website logo as it appears in the sidebar. You can resize it using CSS.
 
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -376,6 +376,8 @@ window.$docsify = {
 
 Website logo as it appears in the sidebar. You can resize it using CSS.
 
+!> Logo will only bee visible if `name` prop is also set. See [name configuration](configuration?id=name).
```
sy-records, line 379:
```suggestion
!> Logo will only bee visible if `name` prop is also set. See [name](#name) configuration.
```

sy-records (approved):
Thanks!

PR #2348, opened 2023-12-21
chore: Update local and Vercel preview. Fix CSS watch functionality.

jhildenbiddle:
This PR continues the cleanup work done in #2316 by removing the need for additional files to be generated or copied to render either a preview or production docs site. The `/docs/` directory is a ready-to-be-served production site, and the local and Vercel server configurations are used to virtualize the necessary changes based on the files in `/docs/`.

FWIW, I would have done this in #2316, but I was unaware of Vercel's middleware capabilities.

If this PR is accepted, PRs #2317 and #2322 can be closed.

## Summary

- Update Vercel preview to be served from virtual `/preview` directory instead of `/preview.html` file.
- Update local dev server configuration to serve `/docs/index.html` with URL rewrites
- Update local prod server configuration to serve `/docs/` directory without modification
- Remove `build:html` script and resulting `preview.html` file
- Fix CSS watch functionality

## Related issue, if any:

#2218

close #2317 
close #2322

Bugfix
Build-related changes

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge


middleware.js
```diff
@@ -0,0 +1,53 @@
...
+    replace: '/CHANGELOG.md',
+  },
+  {
+    // CDN versioned default
+    // Ex1: //cdn.com/package-name
+    // Ex2: http://cdn.com/package-name@1.0.0
+    // Ex3: https://cdn.com/package-name@latest
+    match: /(?:https?:)*\/\/.*cdn.*docsify[@\d.latest]*(?=["'])/g,
+    replace: '/lib/docsify.min.js',
+  },
```
Koooooo-7, line 21: Shall we replace it to `/lib/docsify.js`? I believe it could benefit for debug in dev and preview side.

jhildenbiddle: Since all of our templates use `docsify.min.js`, I think that is the file we should test with as well. It's possible that something could go wrong in the minification process, and if we are previewing with the unminified `docsify.js` file we may not catch it.

But your point about debugging is a good one. I think our best option is to generate a source map for `docsify.min.js` so that we can load the minified bundle but review unminified source via the dev tools when an error occurs.

How does that sound?

Koooooo-7: I see, thats fine to me.

Koooooo-7 (approved):

sy-records (approved):

Issue #2135, opened 2023-07-16
Encapsulating Docsify


trusktr:
We need allow the ability to fully encapsulate Docsify features and state. With encapsulation, it should be possible to mount any number of Docsify "apps" anywhere in a web site, as well as become possible to wrap Docsify as a React/Vue/Svelte/etc component that can be used in any framework.

For example:

- [ ] progress bar currently appends to body. There can be an option to provide the location where it should be appended. It can default to body for backwards compatibility.
- [ ] Routing: it currently assumes control of the address bar. For backwards compatibility, this can be the default.
  - [ ] We need an option to make it route _virtually_, without touching the address bar, for cases when that is desirable
  - [ ] Perhaps add the ability to listen to a `Docsify` instance's route changes, so that they can be mapped to the address bar in a custom way. Also allow setting Docsify's route from outside, so that a website managing multiple instances can "paginate" Docsify manually.
- [ ] Some Docsify state is at the module level (singleton), including the aforementioned progress bar, but also including state variables. We should bring state into the `Docsify` class. Some of this this in #2104 in the items where we're bringing some functions into the classes as private methods.

jhildenbiddle:
What if Docsify uses the existence of a shadow root to determine behavior?

- If the Docsify mount element is not a shadow root or not contained within a shadow root, then Docsify behaves exactly as it does today.
- If the Docsify mount element is a shadow root or is contained within a shadow root, then...
  - Event listeners are added to the shadow root that capture navigation events and prevent default behavior
  - URL state is mocked using a `data-url` attribute on the shadow root

Optionally, the `$docsify` configuration object could be extended to accept either an object (as it does today) or an array of objects to allow initializing multiple encapsulated Docsify instances within a single page.

Thoughts?

jhildenbiddle:
Offering Docsify as a web component would provide encapsulation. Below is a high-level POC using of a `<docsify-root>` web component (name TBD).

**Note:** If we go this route, life will be easier if we standardize on this as the one and only way to create Docsify instances. If we try to support both an IIFE for standard sites (as we do today) and a web component, everything gets more complicated because plugins, themes, testing, etc. have to support both types of Docsify instances.

#### Old template

```html
<!DOCTYPE html>
<html>
  <head>
    ...
    <link rel="stylesheet" href="path/to/theme.css" />
  </head>
  <body>
    <div id="app"></div>
    <script>
      window.$docsify = {
        //...
      };
    </script>
[...]

PR #2392, opened 2024-03-28
output type definitions 

trusktr:
and adjust package.json so that types are picked up in consumer projects

## Summary

The code base is still plain JS as before (the build did not change) but this now allows outputting declaration files so that consumers that will start to use ES Modules will get type definitions.

## Related issue, if any:

- https://github.com/docsifyjs/docsify/discussions/2376

  Feature (dev experience)
  Build-related changes (adds a build step that outputs declaration files, but otherwise does not change the build in any other way)

## For any code change,

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge

## Additional info

This outputs types to a new `dist/` folder. I figured I'd output here because the intent we have is to output to move all outputs to `dist/` so I figured I'd get started. We *could* output types somewhere else like `types/`, but I figured why do that and then move it later, when we can put it where we know we'll want it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,22 +12,37 @@
   "type": "module",
   "// The 'main' and 'unpkg' fields will remain as legacy for backwards compatibility for now. We will add deprectaion warnings to these outputs to give people time to see the warnings in their apps in a non-breaking way, and eventually we can remove the legacy stuff.": "",
   "main": "lib/docsify.js",
+  "types": "dist/core/Docsify.d.ts",
   "unpkg": "lib/docsify.min.js",
   "// We're using the 'exports' field as an override of the 'main' field to provide the new ESM setup. Once we remove legacy 'main', we will remove the 'exports' field and have a simple ESM setup with only a 'main' field.": "",
+  "// These exports require moduleResolution:NodeNext to be enabled in the consumer.": "",
+  "// TODO native ESM (in browsers) does not work yet because prismjs is not ESM and Docsify source imports it as CommonJS": "",
   "exports": {
-    ".": "./src/Docsify.js",
-    "./*": "./*"
+    "./*": "./*",
+    ".": {
+      "types": "./dist/core/Docsify.d.ts",
+      "default": "./src/core/Docsify.js"
+    }
   },
   "files": [
     "lib",
     "themes"
   ],
+  "// Using 'typesVersions' here is the only way we could figure out how to get types working for imports of any subpath without any of the problems other approaches have when not using modeResolution:NodeNext (listed in https://stackoverflow.com/questions/77856692/how-to-publish-plain-jsjsdoc-library-for-typescript-consumers)": "",
+  "typesVersions": {
+    "*": {
+      "src/*": [
+        "dist/*"
+      ]
+    }
+  },
   "scripts": {
     "build:cover": "node build/cover.js",
     "build:css:min": "node build/mincss.js",
     "build:css": "mkdirp lib/themes && node build/css -o lib/themes",
     "build:emoji": "node ./build/emoji.js",
-    "build:js": "cross-env NODE_ENV=production node build/build.js",
+    "build:js": "npm run build:types && cross-env NODE_ENV=production node build/build.js",
+    "build:types": "tsc -p tsconfig.json || true",
     "build:test": "npm run build && npm test",
     "build": "rimraf lib themes && run-s build:js build:css build:css:min build:cover build:emoji",
     "dev": "run-p serve:dev watch:*",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,25 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "target": "ESNext",
+    "lib": ["DOM", "ESNext"],
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "resolveJsonModule": true,
+    "outDir": "dist/"
+  },
+  "include": ["src/**/*.js"],
+  "exclude": [
+    // "./**/*.test.js",
+    // "test/",
+    // "jest.config.js",
+    // "middleware.js",
+    // "playwright.config.js",
+    // "server.configs.js",
+    // "server.js"
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /_playwright-report
 /_playwright-results
 /lib
+/dist
 /node_modules
 /.husky/_
 
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "cSpell.words": ["coverpage"]
+  "cSpell.words": ["coverpage"],
+  "typescript.tsdk": "node_modules/typescript/lib"
 }


sy-records:
Do these need to be fixed?

image

trusktr:
Not necessary to fix, the code works as-is. We can fix type errors as we go along.

Later we can make type definitions for the end users for the options, and for plugins, with nice docs when they hover in their editor.

sy-records (dismissed):

jhildenbiddle:
> Do these need to be fixed?

Yes. We should not merge code that fails our tests and automated checks, and should not merge code assuming someone else will fix these issues later.

> Not necessary to fix, the code works as-is.

When we merge code because it "works as-is" despite the fact that it produces errors, fails tests, or fails our automated checks, ever other maintainer is forced to deal with these issues locally and while reviewing subsequent PRs.

---

Aside from that, this seems like part of a larger effort to allow end-users to utilize docsify as an ESM module, yes? Perhaps I'm missing something, but this could be done *much* more simply than the approach used in the linked [docsify-template-plain-js-type-checked](https://github.com/docsifyjs/docsify-template-plain-js-type-checked) which apparently won't work anyway because PrismJS can't be loaded as an ESM module, right?

So... why would we consider merging this without a better understanding of how consuming docsify as an ESM module is goin to work along with documentation and fully-functional code to make it happen?

jhildenbiddle (changes requested):
As I understand it, this PR is intended to support users who are consuming Docsify as an ESM module, correct?

If so, I don't think we should be distributing TypeScript definitions until end users are able to actually benefit from them. Since Docsify currently can not be consumed as an ESM module, there is no real-world benefit to generating a `d.ts` file. That may change in the future, but until then these changes only add complexity to the project and stand to confuse users who may wonder why we offer a `d.ts` file when there is no way to use it. 

My concern here is with merging code to support future work that we don't have commitments for or documentation to support. I've implemented similar configurations in my own projects (e.g. [mergician](https://github.com/jhildenbiddle/mergician)) so I am not opposed to the approach used in the PR. I simply prefer to see these changes incorporated as part of a larger "Docsify as ESM module" effort since these changes don't provide real-world benefit until the larger effort is complete.

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "target": "ESNext",
+    "lib": ["DOM", "ESNext"],
+    "declaration": true,
+    "emitDeclarationOnly": true,
```
jhildenbiddle, line 11: @trusktr --

This PR currently compiles every `.js` file in `/src/core` and `/src/plugins` to a corresponding `.ts` file `/dist`. It is no emitting a `d.ts` file on my dev machine. This is after a clean install.

CleanShot 2024-03-31 at 12 53 17@2x

trusktr: The screenshot shows `.d.ts` declaration files, which is the expected result. That's what makes the types available in the demo: 

https://github.com/docsifyjs/docsify-template-plain-js-type-checked/blob/main/index.js

however the demo has docsify locally symlinked, which includes the src code. The lib otherwise doesn't export anything from the index file.

I converted the PR into a draft. Perhaps first we should export `Docsify` and allow it to accept options with `new Docsify({...})` and deprecate global `$docsify` as a way to move towards a componentization, non-globals, and ESM, and then having the type defs will make a lot more sense.

trusktr: Ok here's what I'll do: I'll make the deprecation node for `$docsify` in the same PR and expose the new non-global API (with type support as above). Let me know if you prefer the change to be separate (it won't be a big additional change, you'll be able to take a look at the specific commit).

Koooooo-7:
IMO，for current propose to replace `$docsify` seems not necessary （module-lize).
WDYT?

trusktr:
Let's chat about the concept, let me know what you think of this path.

In the [last commit](https://github.com/docsifyjs/docsify/pull/2392/commits/1a352afa205342698aee91c3fc07a42c2f01a13a) I marked `$docsify` as "deprecated", and added `console.warn` messages. I also exposed `src/` files in `package.json``.files`, and adjusted `exports` so that imports of original source modules as well as single-file bundle modules in dist have type definitions visible when imported.



I updated the demo repository so that each of the following files show how the `import` looks like (with type checking and intellisense included) while the current version of Docsify from this PR is symlinked into its `node_modules` (to try it, symlink docsify using the same instructions as in the OP):

- https://github.com/docsifyjs/docsify-template-plain-js-type-checked/blob/f7e64899bcaef6a270e5c341e8c662be9aeda1c9/index.js
- https://github.com/docsifyjs/docsify-template-plain-js-type-checked/blob/f7e64899bcaef6a270e5c341e8c662be9aeda1c9/index2.js
- https://github.com/docsifyjs/docsify-template-plain-js-type-checked/blob/f7e64899bcaef6a270e5c341e8c662be9aeda1c9/index3.js
- https://github.com/docsifyjs/docsify-template-plain-js-type-checked/blob/f7e64899bcaef6a270e5c341e8c662be9aeda1c9/index4.js

The template repo is WIP, it can't run because there's no build and the `importmap` is not done (it requires handling of prismjs which is UMD format), but it shows that type definitions are working with 4 different `import` methods, and people who do have a build would be able to start importing `Docsify` in apps with webpack, rollup, etc. We can make another template that uses a build.

trusktr:
The `console.warn` message cause some playwright `expect(consoleMsg).toBeUndefined()` tests to fail, but that's easy to fix (should we even be testing that?). The code otherwise still works (back compat).

jhildenbiddle:
I think this is a good example of why these types of changes should be introduced as either a new [issue](https://github.com/docsifyjs/docsify/issues) (better for triage/tracking) or [discussion](https://github.com/docsifyjs/docsify/discussions) (better for higher-level discussions) instead of a PR. What started off as a PR about generating TypeScript types has evolved into a discussion involving ESM modules, deprecating globals, and componentization. These are all good things to discuss, but the comments section of this PR is not the right place IMHO.

@trusktr -- There's some good stuff in here, so please don't take the above comment negatively. I am onboard with the core concepts of what is being proposed here, but I'd like us to align and execute efficiently since we have limited resources available. For example, we have agreed to keep v5 compatible with v4 with the exception of SSR and legacy browser support. That means we cannot deprecate Docsify globals in v5, but we *could* offer Docsify as an ESM module in addition to our current IIFE version. If do that, I'm on board with generating types but I would approach the implementation differently which is worth discussing.

Can we create / move to separate issues so can align and triage?

1. ESM export => New issue
2. TypeScript support => New issue
3. Docsify encapsulation => #2135

Thanks!

jhildenbiddle (changes requested):
I think this is a good example of why these types of changes should be introduced as either a new [issue](https://github.com/docsifyjs/docsify/issues) (better for triage/tracking) or [discussion](https://github.com/docsifyjs/docsify/discussions) (better for higher-level discussions that produce issue(s) to track/triage) instead of a PR. What started off as a PR about generating TypeScript types has evolved into a discussion involving ESM modules, deprecating globals, and componentization. These are all good things to discuss, but the comments section of this PR is not the right place IMHO.

@trusktr -- There's some good stuff in here, so please don't take the above comment negatively. I am onboard with the core concepts of what is being proposed here, but I'd like us to align and execute efficiently since we have limited resources available. For example:

- We cannot deprecate Docsify globals because we have agreed to keep v5 compatible with v4 with the exception of SSR and legacy browser support. Deprecating globals would therefore have to be targeted for a future release.
- We *could* offer Docsify as an ESM export for v5 in addition to our current IIFE version, but if we do this we also have to consider if/how we export plugins and themes.
- I like the idea of providing better IntelliSense for docsify configurations, but the implementation I would use is different than the one in this PR.

Can we create / move to separate issues so can align and triage?

1. ESM export => New issue
   TypeScript / IntelliSense support can be part of this discussion.
2. Docsify encapsulation => #2135 
   I've updated this thread with high-level POC. Deprecation of globals can be part of this discussion. 

Thanks!

trusktr:
Hey guys, no offense taken at all! I usually propose ideas in the form of code changes. This could be a prototype to help discuss how type checking (without writing TS, only JS with JSDoc comments) can be implemented.

I'm currently experimenting with Astro, and will get a custom element setup with SSR/SSG working in Astro with Lume Element (not Docsify related, just Astro + Lume Element only, to learn about the approach).

One thing that would be nice for when we eventually get Docsify converted into a Custom Element with SSR/SSG working, is to have type checking.

In Astro, for example, all markup is type checked. So for example, it will be possible for a user to write the following in Astro, React JSX, Solid JSX, etc, and they will get type checking. For example, the current [`maxLevel` option](https://docsify.js.org/#/configuration?id=maxlevel) accepts numbers only. A user will be able to write this:

```js
function MyReactComponent() {
  return <docsify-app max-level={3} />
}
```

but if they were to write this:


```js
function MyReactComponent() {
  return <docsify-app max-level={"foo"} />
}
```

they will get a type error in their editor (f.e. VS Code) because `"foo"` is a `string`, not a `number`.

So basically with this PR I'm just starting to get things in place for this future, but it may not be in the ideal final format yet. Let's use this as a discussion point so we can ideate.

In the meantime I'm in Astro land (https://astro.build/chat). If you go there 🚀 let me know!

trusktr:
> Can we create / move to separate issues so can align and triage?

@jhildenbiddle yeah please feel free to make issues as needed. I have to admit I will not be too active over here while I'm in Astro land. But I'll be back with some info on what I find.

trusktr:
I've returned 🚀. I got some experience with SSR in Astro.

I've updated this PR with the latest from `develop`, and added an example of consuming ESM in a sample project in `test/consume-types/` (it runs as part of the tests in CI to ensure types check out).

The basic idea is:

```js
import {Docsify} from 'docsify'

// With type checking and intellisense, we can see the
// available Docsify options in our IDE.
new Docsify({
  el: '#app',
  // ...etc...
})
```

Now, this is not a custom element yet, but it's a middle ground: using globals is bad and not suitable for a good ESM TypeScript experience, so this PR makes it possible to consume the `Docsify` in an ESM project until we convert to a custom element (I'm imagining that'll may be v6) with type awareness and intellisense.

trusktr:
Aaaand, I broke something. But we're almost there. Must be a tiny change somewhere.

src/core/event/index.js
```diff
@@ -242,13 +246,13 @@ export function Events(Base) {
...
-      dom.on(sidebarElm, 'click', ({ target }) => {
-        const linkElm = target.closest('a');
+      dom.on(sidebarElm, 'click', (/** @type {MouseEvent} */ { target }) => {
+        const linkElm = /** @type {HTMLElement} */ (target).closest('a');
         const linkParent = linkElm?.closest('li');
         const hasSubSidebar = linkParent?.querySelector('.app-sub-sidebar');
 
         if (hasSubSidebar) {
-          dom.toggleClass(linkParent, 'collapse');
+          dom.body.classList.toggle('collapse');
```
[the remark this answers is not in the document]
trusktr: Tests did not catch this!

test/consume-types/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Commands
+
+### `npm run typecheck`
+
+Run type checking for the project.
+
+### `npm run serve`
+
+Serve the example at http://localhost:5500 to verifiy vanilla ESM usage works.
```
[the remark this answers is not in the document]
trusktr: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/docsifyjs/docsify/pull/2392#discussion_r2533418232)

src/core/util/dom.js
```diff
@@ -1,16 +1,18 @@
...
  * Get Node
  * @param  {String|Element} el A DOM element
  * @param  {Boolean} noCache Flag to use or not use the cache
- * @return {HTMLElement | SVGElement} The found node element
+ * @return {Element} The found node element
  */
 export function getNode(el, noCache = false) {
   if (typeof el === 'string') {
     if (typeof window.Vue !== 'undefined') {
+      window.Vue.version;
```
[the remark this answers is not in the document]
trusktr: @copilot delete the line

test/consume-types/prism.js
```diff
@@ -0,0 +1,27 @@
...
+// tricking its CommonJS format to see module.exports.
+
+const __prismCjs = await fetch('/node_modules/prismjs/prism.js').then(res =>
+  res.text(),
+);
+
+// Emulate CommonJS environment
+const module = { exports: {} };
+// eslint-disable-next-line no-unused-vars
+const exports = module.exports;
```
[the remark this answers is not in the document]
trusktr: This is used in scope of the `eval()` for CJS emulation.

package.json
```diff
@@ -24,18 +24,22 @@
...
   "main": "dist/docsify.js",
+  "types": "src/core/Docsify.d.ts",
   "exports": {
     ".": "./src/core/Docsify.js",
     "./*": "./*"
   },
   "files": [
-    "dist"
+    "dist",
+    "src"
```
sy-records: The src folder has been removed.

Will adding it back now affect users?

trusktr: This is how we can give people type definitions with good DX (f.e. `Go To Definition` works well) with the original ESM files. This also gives people with more flexibility in configuring how they import and/or build Docsify into their app (build example not shown in this PR).

trusktr: Try opening the `test/consume-types` folder in VS Code as the root folder (f.e. `code test/consume-types`), then run `npm clean-install --install-links` (the `--install-links` option will take the Docsify source from the root of the repo, build it, and install it as if it were downloaded from the npm registry). Finally you can open up the `example.js` file and try out type checking and intellisense (f.e. try looking at config option types, or running `Go To Definition` on `Docsify`).

trusktr: Here's a video showing the intellisense:


https://github.com/user-attachments/assets/b0214aa5-abb0-422a-a8bd-56bb2dcc791d

sy-records (approved):

PR #2401, opened 2024-04-11
fix: skip-to-content scroll behavior

jhildenbiddle:
## Summary

Fixes an issue the new (unreleased) "Skip to Content" link that caused an immediate jump to to the content (instead scrolling) and the main content to scroll down (instead of being static) into place. 

**Example of bug** (develop branch preview w/ Chrome 123)

The incorrect behavior is the result of three separate things occurring at the same time:

1. By default, browsers will automatically scroll any offscreen element that receives focus into view. This is the correct behavior for accessibility purposes because the focused item should be in view.
2. Clicking the "Skip to Content" link places the focus on the main content area. This is the correct behavior for accessibility purposes because the focused item should be in view. However, by placing the focus on an off-screen element, the browser's default behavior (see above) causes the scroll position to jump to the content instead of smoothly scrolling to it.
3. Docsify's "smooth scroll" behavior aligns the main content container element to the top edge of the viewport.

https://github.com/docsifyjs/docsify/assets/442527/7205bd0e-598e-4e9f-97d6-11d586cad434

**Example of fix** (PR changes w/ Chrome 123)

The fix if simple thanks to modern APIs: set the `preventScroll` option to `true` when calling the browser's native `focus()` method:

https://github.com/docsifyjs/docsify/assets/442527/9509967b-40e1-469f-9cb5-e6b64e6c3f1e

## Related issue, if any:

None

Bugfix

## For any code change,

N/A

No

- [x] Chrome
- [x] Firefox
- [x] Safari

### DIFF
--- a/src/core/event/index.js
+++ b/src/core/event/index.js
@@ -31,7 +31,7 @@ export function Events(Base) {
 
       // Move focus to content
       if (query.id || source === 'navigate') {
-        this.focusContent();
+        this.#focusContent();
       }
 
       if (loadNavbar) {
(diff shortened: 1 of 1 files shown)


Koooooo-7:
Hi @jhildenbiddle , could plz more details on it? (Much thx that if you could provide a screenshot)  
sorry, that I don't full understand the behaviors.

jhildenbiddle:
@Koooooo-7 --

Videos demonstrating bug and fix added to PR summary (above).

Koooooo-7 (approved):
LGTM, and nice experience improvement.

sy-records (approved):
LGTM. Good job!

PR #2402, opened 2024-04-11
fix: dev mode hot reload and add sourcemaps

jhildenbiddle:
## Summary

Fixes two development issues:

1. Hot reload was not working because the docs site was loading `lib/docsify.min.js` which was only updated on "production" builds. The build now generates minified and non-minified bundles for all builds.
2. Source maps are now generated for all `.min.js` files in `/lib/`. This solution is best for troubleshooting during development because source maps provide unminified source code along with accurate file and line locations on errors. Using non-minified bundles for development was previously discussed, however this approach does not provide accurate source file and line locations.

## Related issue, if any:

None

Bugfix
Build-related changes

## For any code change,

N/A

No

- [x] Chrome
- [x] Firefox
- [x] Safari

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -3,8 +3,8 @@ import { promises as fs } from 'fs';
 import * as rollup from 'rollup';
 import commonjs from '@rollup/plugin-commonjs';
 import nodeResolve from '@rollup/plugin-node-resolve';
-import uglify from '@rollup/plugin-terser';
 import replace from '@rollup/plugin-replace';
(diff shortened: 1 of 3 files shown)


Koooooo-7 (approved):
LGTM.

sy-records (approved):

jhildenbiddle:
@Koooooo-7 --

Awaiting approval to merge PR. Thx!

Koooooo-7 (approved):

PR #2403, opened 2024-04-11
fix: prevent unnecessary themeColor deprecation notice

jhildenbiddle:
## Summary

The following deprecation warning was spamming the console even when `$docsify.themeColor` was not being used: 

```
$docsify.themeColor is deprecated. Use a --theme-color property in your style sheet. Example:
<style>
  :root {
    --theme-color: deeppink;
  }
</style>
```

This console message will now only appear when $docsify.themeColor is used.

## Related issue, if any:

None

Bugfix

## For any code change,

N/A

No

- [x] Chrome
- [x] Firefox
- [x] Safari

### DIFF
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -47,17 +47,19 @@ export default function (vm) {
         return this.__themeColor;
       },
       set themeColor(value) {
-        this.__themeColor = value;
-        console.warn(
-          stripIndent(/* html */ `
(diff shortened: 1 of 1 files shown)


Koooooo-7 (approved):
LGTM.

sy-records (approved):

PR #2404, opened 2024-04-11
refactor: event modernization and optimization

jhildenbiddle:
## Summary

The motivation for this PR is to remove/replace event handling code no longer needed for legacy browsers and to optimize event handling so that we can more easily address the [many scroll-related issues](https://github.com/docsifyjs/docsify/issues?q=is%3Aissue+is%3Aopen+scroll) for the upcoming v5 release or in a future release following v5. 

No features or tests have been added or removed. All of the pre-existing tests continue to pass.

Updates to Docsify's event handling are as follows:

- Replace code no longer required for legacy browser compatibility with modern, native web APIs
- Remove dependencies no longer required for legacy browser compatibility
- Optimize performance by utilizing modern, native web APIs to reduce/remove reflows, repaints, manual calculations, and successive/unnecessary function calls (e.g., [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) instead of [`scroll`](https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event) events)
- Simplify event-related codebase by improving function names, using a consistent naming convention, adding JSDoc-style comments, and isolating function tasks and side effects.

Included but not directly event-handling related:

- Removed unused SSR code
- Minor cleanup of NPM scripts

**NOTES:**

- Almost all of the event-related changes are found in `/src/core/event/index.js`. Other file updates are either small changes to accommodate the updates to `/src/core/event/index.js` or are unrelated build-related changes.
- If I could have isolated the event-handling changes into separate smaller PRs, I would have. The pre-existing event handling was intertwined between multiple files and poorly named methods, making it far easier to fix in a single PR.

## Related issue, if any:

None

Bugfix
Code style update
Refactor

## For any code change,

N/A

No

- [x] Chrome
- [x] Firefox
- [x] Safari


Koooooo-7 (approved):
LGTM.

sy-records (approved):

Issue #2336, opened 2023-12-15
All distributable files belong in the same directory (move all to `dist/`)


sy-records:
> We actually need it because it's referenced in the documentation. see https://docsify.js.org/#/quickstart?id=specifying-docsify-versions

Good catch, however I do not think we should restore the `/themes` directory. 

Having duplicate files in multiple locations in a single repo is a bad idea. Doing so invites confusion among those unfamiliar with the details of why duplicate files exist, which can lead to public-facing mistakes that require long-term support. This is an excellent example, and I believe it is a mistake we should correct for v5:

1. The next release is going to be a breaking change because our codebase no longer supports legacy browsers. If we want to make a change, this is the time to do it.
2. Our documentation should not recommend loading unminified CSS or JS files on a production site. Similarly, our CLI output should not load unminified CSS or JS by default. We can offer unminified CSS/JS as options, but our recommendation should be to use minified CSS/JS.
4. All distributable files belong in the same directory. For v4, this is `/lib`. For v5, the proposed location is `/dist`.
5. If we want to offer minified and unminified CSS files, we should differentiate using the filename and not the location, just as we do with `docsify.js` and`docsify.min.js`. For v5, this would mean something like `/dist/themes/vue.css` and `/dist/themes/vue.min.css`. #2338

I planned to address this issue in an upcoming PR as part of our v4 => v5 transition:

1. Modify build to output distributable files to `/dist` instead of `/lib`. This is part of the v4 => v5 transition.
2. Update URLs as needed in `docsify` and `docsify-cli` repos to reference `/dist` instead of `/lib`.
3. Create static redirects/imports from `/lib` and `/themes` to `@4/lib` and `@4/themes` CDN URLs. This allows v4 sites without a URL version lock to continue working when they unintentionally load v5 after its release.
4. Add a console deprecation / upgrade notice for sites loading resources from `/lib` or `/themes` redirects/imports.

The static redirects/imports in `/lib` and `/themes` are a temporary solution. They exist only to prevent sites that load docsify resources without URL version locks from breaking when v5 releases because v4 was released in 2017 but our docs did not recommend URL version locks until mid-2020. My hope is that they will can remove them by the time v6 is released.

_Originally posted by @jhildenbiddle in https://github.com/docsifyjs/docsify/pull/2335#pullrequestreview-1782657079_

trusktr:
Updated the title to reflect moving stuff into `dist/`

PR #2428, opened 2024-05-17
JS/CSS Build Refactor & Documentation Updates

jhildenbiddle:
## Summary

_This PR includes refactor and clean-up work in preparation for the next major release (v5). Only changes to build and docs are included. No functional or test-related changes have been made aside from those required to facilitate the new build output directory (see below)._

- **Change build output from `/lib` to `/dist` in preparation for v5 release**
  Changing the output directory ensures v4 users won't accidentally load v5 resources. This change is required due issues with CDN URL version locks in our documentation and templates (see next item).

- **Update docs to include `@5` version lock on all CDN URLs**
  Many CDN URLs in our documentation and templates included `@latest` as the version lock or omitted a version lock altogether. CDN URLs without a proper version lock will redirect to the latest version, potentially causing previously working sites to load breaking changes when a new version of Docsify is released. Adding a `@5` version lock to our CDN URLs will prevent future users from having to worry about this issue.

- **Simplify JS build by switching from `/build` script to a rollup config file**
  Rollup configuration files are easier to work. Making the switch reduced LOC for the JS build by ~50% and allowed for several dev dependencies to be removed. Each bundle now contains a "banner" comment to easily identify the version.

- **Simplify CSS builds by converting from `/build` scripts to CLI tools**
  No need for multiple build scripts. CLI tools are easier to work with and offer built-in watch functionality. Sourcemaps are now generated for both minified and un-minified versions, making it easy to see line numbers from the stylus source while using the web inspector for troubleshooting. Making the switch allowed for two build scripts and several dev dependencies to be removed. 

- **Update JS, CSS, and a few miscellaneous dependencies**
  These updates reduce our dependency security vulnerabilities from 15 to 0.
  
- **Clean up Quick Start, Themes, and CDN pages**
  Ensure all CDN URLs reference minified files and deprioritize uncompressed versions. Briefly explain semantic versioning. Clarify benefits of lock to major or specific versions.

- **Clean up package.json**
  Reordered items for easier scanning, removed unnecessary comments, and added a few keywords.

## Related issue, if any:

#2336 

Refactor
Docs
Build-related changes

## For any code change,

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge


Koooooo-7 (dismissed):
LGTM.
The `docsify@version` import way should be the only recommend way since now and then.

TODO:
- `CLI` generated paths, resources.
- Discussion of the `CLI` scope with build CSS and more.

trusktr:
> * Refactor CSS builds from `/build` scripts to CLI tools

Would love to remove Stylus and just use plain CSS, as it now has pretty much everything we used stylus for. Maybe it would have been a better spend of time if you just deleted Stylus build and converted to CSS?

Stylus deletion is listed here:

- #2376 


> * "engines" prop to specify node >= 20

Is it necessary? I'd only add it if a problem is encountered (did that happen?), otherwise no need to unnecessarily block someone if they're on an older version of Node.



> 1. Deciding if/how we want to handle CDN URLs out in the wild that will automatically redirect to the `/lib` directory of the latest Docsify versions.

I think the only way is to continue to publish the old lib files in the new version for some time, but add a deprecation warnings to them (f.e. with console.warn, or maybe even going as far as injecting a small message at the bottom of the sidebar or something) telling people to stop using versionless URLs and that these files will be removed in N number of months (6 months?).

jhildenbiddle:
> Would love to remove Stylus and just use plain CSS, as it now has pretty much everything we used stylus for. Maybe it would have been a better spend of time if you just deleted Stylus build and converted to CSS?

Agreed. Happy to make this change either to this PR or as a separate PR. Since this PR already has already been approved by @Koooooo-7, I'd prefer we get this one merged and handle removing Stylus in a separate PR.

> > * "engines" prop to specify node >= 20
> 
> Is it necessary? I'd only add it if a problem is encountered (did that happen?), otherwise no need to unnecessarily block someone if they're on an older version of Node.

Yes. The new rollup build uses [`import.meta.dirname`](https://nodejs.org/api/esm.html#importmetadirname) which was introduced in node 20. 

I think requiring the latest LTS version of node is acceptable and non-controversial. I do not think the same is true for non-LTS versions (21) or "current" versions that have not yet entered LTS support (22).

Node v20 is the latest LTS version and all of our tooling supports it (I've already made the change from node 18 to 20 in our Vercel preview settings). Node 22 is the "current" release and will enter long-term support (LTS) in October of this year, so I don't see requiring node 20 today as an issue.

As for developers being blocked, tools like [nvm](https://github.com/nvm-sh/nvm) make version switching version trivial and the `engines` prop will ensure that they receive a warning if they forget to switch to >= v20. If/when we encounter a maintainer / contributor blocked by the node 20 requirement, I'd lean towards getting them setup with a tool like nvm rather than limiting our access to features available in the latest LTS version of node.

> 
> > 1. Deciding if/how we want to handle CDN URLs out in the wild that will automatically redirect to the `/lib` directory of the latest Docsify versions.
> 
> I think the only way is to continue to publish the old lib files in the new version for some time, but add a deprecation warnings to them (f.e. with console.warn, or maybe even going as far as injecting a small message at the bottom of the sidebar or something) telling people to stop using versionless URLs and that these files will be removed in N number of months (6 months?).

Yep. This is the option we discussed previously on Discord and the one I'm leaning towards. There are a few details and options to discuss that are better handled in Discord. I mentioned the need to have that discussion because I didn't want maintainers to think this PR represented everything I think we need to do to release v5. It isn't. It is one of several PRs that I intend to file, but I'm trying to create them 1) in an order that makes sense and 2) with high-level consensus among the maintainers.

Koooooo-7 (approved):

docs/quickstart.md
```diff
@@ -46,10 +47,7 @@ If you don't like `npm` or have trouble installing the tool, you can manually cr
   <head>
     <meta name="viewport" content="width=device-width,initial-scale=1" />
     <meta charset="UTF-8" />
-    <link
-      rel="stylesheet"
-      href="//cdn.jsdelivr.net/npm/docsify@4/themes/vue.css"
-    />
+    <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsify@5/themes/vue.min.css" />
```
sy-records, line 50: Everything looks OK to me. But until we release v5, these links are broken.

jhildenbiddle: Yes, but I'm not sure why that matters. The docs in the `develop` branch already contain new features and changes that will only be available when v5 is released.

- These changes will not be visible on [docsify.js.org](https://docsify.js.org) until `develop` is merged into `main`.
- These `@5/dist` URLs are not broken during local development or Vercel previews because CDN URLs are [rewritten](https://github.com/docsifyjs/docsify/blob/1ceb03b013e1454c0fe3a76971c49bdde00944e6/middleware.js#L8-L30) to `/dist` in those environments. For example, notice how this PR contains `@5/dist` CDN URLs in [`/docs/index.html`](https://github.com/docsifyjs/docsify/blob/v5-build-overhaul/docs/index.html) but the Vercel [preview](https://github.com/docsifyjs/docsify/blob/v5-build-overhaul/docs/index.html) works as expected.

jhildenbiddle:
@trusktr --

One last item I forget to address:

> Speaking of this, how can we enforce it for v5 to v6? And idea is while we're on v5, the dist folder could be dist/v5/ and when v6 comes out then we move to dist/v6/ but keep dist/v5/ around for N months with deprecation warnings for those still not using version number. Etc.

CDNs handles this for us. jsDelivr provide access to all published versions via `@` version locks. For example, `@4` CDN URLs will continue to work when v5 is published and becomes the CDN redirect target or `@latest`. Same goes for `@5` URLs when v6 is published. 

Other CDNs (like [bootcdn.cn](https://www.bootcdn.cn/docsify/)) do the same using the version as path of the path. For example: https://cdn.bootcdn.net/ajax/libs/docsify/4.13.1/docsify.js

---

If the only outstanding issue is the removal of Stylus, can you approve this PR with so we can move forward knowing that I will remove Stylus in a separate PR (likely later tonight)?

Thanks!

trusktr:
> Yes. The new rollup build uses [`import.meta.dirname`](https://nodejs.org/api/esm.html#importmetadirname) which was introduced in node 20.

Good enough for me!

> I'd prefer we get this one merged and handle removing Stylus in a separate PR.
> 
> >

SGTM

> `@4` CDN URLs will continue to work when v5 is published and becomes the CDN redirect target or `@latest`. Same goes for `@5` URLs when v6 is published.

Yeah, people *should* use versioned URLs. Totally. What I mean is that someone out there might (for whatever reason that might not be good) use an unversioned URL like `https://unpkg.com/docsify/dist/docsify.js` or similar to use v5, then that could cause their app to break when we release v6.

So, the idea was that if we publish something like `https://unpkg.com/docsify/dist/v5/docsify.js` where the `v5/` folder is actually part of what we publish, then later when we release v6 we'd just update the dist folder to `dist/v6/` while leaving `dist/v5/` with a the old code and deprecation warnings (delete it after some time).

Would this be worth it? Or should we not worry about helping people who may still use unversioned URLs, and we let their sites break without a warning?

jhildenbiddle:
@trusktr --

Moving the versioned URL discussion to Discord.

Sine you appear to be okay with the changes in this PR, please approve it so that we unblock other PRs that are waiting on this one to be merged. 

Thx!

trusktr (approved):

Issue #2432, opened 2024-05-20
Update or remove outdated / abandoned documentation translations


jhildenbiddle:
## Bug Report

The documentation translations available on [docsify.js.org](https://docsify.js.org/) are outdated and in some cases completely abandoned. Inaccurate documentation in a preferred language is (IMHO) far worse than accurate documentation in a non-preferred language.

Last updated:

- [docsifyjs/docs-es](https://github.com/docsifyjs/docs-es): 11/7/2023
- [docsifyjs/docs-de](https://github.com/docsifyjs/docs-de): 4/23/2020
- [docsifyjs/docs-ru](https://github.com/docsifyjs/docs-ru): 4/23/2020
- [docsifyjs/docs-zh](https://github.com/docsifyjs/docs-zh): 8/28/2023

Without native language speakers to keep these translations up to date, it seems wise to consider the following options (ordered from most to least preferred):

1. Retire these translations altogether and rely on modern browser language translation features (see below).
2. Explore automated translation services to keep these repos up to date. [GitLocalize](https://gitlocalize.com) is one free option worth considering.
3. Continue manually supporting only the translation(s) we can verify site visitors use and that maintainer(s) are willing to actively maintain.

**IMPORTANT:** At the time Docsify's documentation translations were created, browsers did not offer built-in language translation functionality. Today, all major browsers offer built-in language translation functionality. This means that non-native English speakers can visit our English docs and have them auto-translated to their native language by the browser automatically if they prefer. 

Below are screenshots from taken while viewing [docsify.js.org](https://docsify.js.org) translations as an English-preferred visitor:

**Chrome - 简体中文**

![CleanShot 2024-05-20 at 12 35 26@2x](https://github.com/docsifyjs/docsify/assets/442527/3d8f5373-f6c2-44ae-b991-67abbf70fec0)

**Safari - Deutsch**

![CleanShot 2024-05-20 at 12 45 49@2x](https://github.com/docsifyjs/docsify/assets/442527/726f468d-94a2-4dfa-a00d-d6cc036d9d7a)

**Firefox - Spanish**

![CleanShot 2024-05-20 at 12 44 33@2x](https://github.com/docsifyjs/docsify/assets/442527/9a291428-4559-464a-bd22-3b7cd1e46fc0)

paulhibbitts:
[...]

PR #2437, opened 2024-05-23
ci: create crowdin action

sy-records:
Translations can now be done at https://translate.docsifyjs.org

ref: #2432


paulhibbitts (approved):
GitHub action looks good to me, thanks for this PR @sy-records !

jhildenbiddle (dismissed):
LGTM.

src/core/render/emoji-data.js
```diff
@@ -1,4 +1,4 @@
- 
+/* eslint-disable */
```
jhildenbiddle, line 1: Looks like this file was checked in by accident. No need to include the updated emoji short codes with this PR.

jhildenbiddle (changes requested):

src/core/render/emoji-data.js
sy-records: I merged it here on purpose, I wanted to modify the emoji action.

This file does not require `/* eslint-disable */`?

jhildenbiddle: I believe this is correct now that `eslint-data.js` is ignored via `/eslint.config.js`. The comment is being added by the `build/emoji.js` script while auto-generating `emoji-data.js` content.

PR #2438, opened 2024-05-23
chore: Update lint configuration (ESLint 9, Prettier 3)

jhildenbiddle:
## Summary

No functional changes. This PR contains updates to our lint-related dev dependencies and a number of file resulting from the updated lint tools and configuration.

- Update ESLint from v8 to v9
- Update Prettier from v2 to v3
- Migrate ESLint configuration to new flat configuration format
- Remove deprecated ESLint rules
- Remove ESLint rules from config that match default values
- Remove [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import) (no ESLint v9 support)
- Fix lint issues throughout repo

## Related issue, if any:

N/A

Build-related changes

## For any code change,

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge


paulhibbitts (approved):

Koooooo-7 (approved):
LGTM.

paulhibbitts (approved):

playwright.config.js
```diff
@@ -23,7 +23,7 @@ const config = {
   testDir: './test/e2e',
   timeout: 30 * 1000,
   workers: process.env.CI ? 1 : undefined, // No parallel tests on CI
-  forbidOnly: !!process.env.CI, // Fail on CI if test.only in source
+  forbidOnly: Boolean(process.env.CI), // Fail on CI if test.only in source
```
trusktr, line 26: Is this enforced? The other one is more concise and well known.

src/core/render/compiler.js
```diff
@@ -282,7 +282,7 @@ export class Compiler {
       }
 
       const tree = this.cacheTree[currentPath] || genTree(toc, level);
-      html = treeTpl(tree, /* html */ `<ul>{inner}</ul>`);
+      html = treeTpl(tree, /* html */ '<ul>{inner}</ul>');
```
trusktr, line 285: This one is interesting. There's a new sentiment floating around that preferring back ticks is nicer because then you can easily add interpolations at any time without having to manually convert the quotes.

src/core/router/index.js
```diff
@@ -43,7 +43,6 @@ export function Router(Base) {
       this.updateRender();
       lastRoute = this.route;
 
-      // eslint-disable-next-line no-unused-vars
       router.onchange(params => {
```
trusktr, line 46: Maybe better to just have `() =>` instead of `params =>` or `_ =>`

src/core/util/ajax.js
```diff
@@ -1,5 +1,5 @@
 // @ts-check
-/* eslint-disable no-unused-vars */
+
 import progressbar from '../render/progressbar.js';
```
trusktr, line 3: Would prefer no unused cars (and not adding ignore comments but actually removing them)

src/plugins/disqus.js
```diff
@@ -26,7 +26,6 @@ function install(hook, vm) {
     div.style = `width: ${main.clientWidth}px; margin: 0 auto 20px;`;
     dom.appendTo(dom.find('.content'), div);
 
-    // eslint-disable-next-line
     window.disqus_config = function () {
```
trusktr, line 29: I think we should leave this error as a warning, so we can remember to fix it. Globals are bad, and that's what we'll want to move away from with ESM and custom elements.

src/plugins/search/component.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
 import { search } from './search.js';
```
trusktr, line 1: Let's remove unused var

src/plugins/search/index.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
```
trusktr, line 1: Let's remove all unused vars if any (note I'm reading this in the GitHub UI, not sure if it was only a spurious comment)

src/plugins/search/search.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
```
trusktr, line 1: Let's remove all unused vars if any (note I'm reading this in the GitHub UI, not sure if it was only a spurious comment)

test/helpers/wait-for.js
```diff
@@ -25,7 +25,7 @@ function waitForFunction(fn, arg, options = {}) {
 
       try {
         result = fn(arg);
-      } catch (e) {
+      } catch {
         // Continue...
```
trusktr, line 29: I'm always weary if silence *any* type of error. I too often experience time wasted when some issue goes silenced and we don't know where it is. Funny enough the emoji script was a recent example in this very project.

Perhaps we should at least console.* any error just in case. Or better would be, if we know what error we want to skip, we should specifically skip that error. That way actually-unexpected errors won't accidentally be silenced.

test/integration/emoji.test.js
```diff
@@ -150,7 +151,8 @@ describe('Emoji', function () {
   test('Ignores emoji shorthand codes in style url() values', async () => {
     await docsifyInit({
       markdown: {
-        homepage: /* html */ `<style>@import url(http://domain.com/:smile/file.css);</style>`,
+        homepage:
+          /* html */ '<style>@import url(http://domain.com/:smile/file.css);</style>',
```
trusktr, line 155: Did we shorten our allowable line length? Would prefer not to do that.

playwright.config.js
jhildenbiddle: Yes. I believe it is enforced via ESLint's [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast) rule which is enabled when using their recommended settings. I could be Prettier, too.

The key point is that this change was not caused by custom settings in our lint configuration. We can look into overriding this setting if possible (Prettier is proudly infamous for being "opinionated" and therefore non-configurable), but my preference is to favor ESLint's and Prettier's recommended settings and then selectively enable additional rules that are not enabled by ESLint.

src/core/util/ajax.js
jhildenbiddle: ESLint's `no-unused-vars` is configured in `eslint.config.js`. The comment is auto-removed when running `lint:fix` because it is unnecessary. The rule is still enforced.

src/plugins/search/component.js
jhildenbiddle: ESLint's `no-unused-vars` is configured in `eslint.config.js`. The comment is auto-removed when running `lint:fix` because it is unnecessary. The rule is still enforced.

src/plugins/search/index.js
jhildenbiddle: ESLint's `no-unused-vars` is configured in `eslint.config.js`. The comment is auto-removed when running `lint:fix` because it is unnecessary. The rule is still enforced.

src/plugins/search/search.js
jhildenbiddle: ESLint's `no-unused-vars` is configured in `eslint.config.js`. The comment is auto-removed when running `lint:fix` because it is unnecessary. The rule is still enforced.

test/integration/emoji.test.js
jhildenbiddle: Nope. Line length was not changed. My guess is that this due to a change in how Prettier handles comments in v3.

test/helpers/wait-for.js
Koooooo-7: I think @trusktr suggestion is reasonable.
It more likely the code style or best practice thing:
- Don't make the Exception silence.
- Let it broken.

Based on this PR's points, I think the change it's fine, cos this is the Lint PR for formatting current codebase (not a purely refine/refactor one).
And we could raise another topic/issue to drive going through the whole project to make sure all those parts have a `console` in `warn` or `error` at least.

src/core/render/compiler.js
jhildenbiddle: It's an interesting position that some devs have been proposing for a long time. Here's an issue in the Prettier issue from way back in 2017: https://github.com/prettier/prettier/issues/54

My position is that we had single quotes before this PR, this PR does not change that, and so we should therefore not allow a potential debate about a lint configuration change nobody is asking for hold up this PR.

Worth considering:

- The idea of using backticks by default is not new. Neither ESLint or Prettier have adopted this position and those maintainers have undoubtably spent *way* more time thinking about it than anyone here has. If one of those tools adopted backticks by default, I would be more likely to adopt it by default as well. Until then, I'm less likely to do so.
- Presumably the reason to adopt a tool like Prettier is to avoid getting mired down in debates about stylistic opinions like this. Prettier doesn't offer an option to favor backticks over single or double quotes. We can enforce backticks by default with ESLint if we use it to override Prettier.

If people are interested in switching to backticks by default, let's discuss that separately. I don't have strong feelings one way or the other, but I lean towards sticking with what we have (probably just out of familiarity and muscle memory).

src/core/router/index.js
jhildenbiddle: `params` is needed here. I have no idea why that `eslint-disable` comment was there, but like the others it's being auto-removed by `lint:fix` because it is unnecessary.

playwright.config.js
jhildenbiddle: Fixed.

Koooooo-7 (dismissed):

src/plugins/disqus.js
jhildenbiddle: What error are you referring to?

As far as globals go, they aren't "bad" for our product as it currently ships. When we are ready to promote an ESM build we can worry about globals.

test/helpers/wait-for.js
jhildenbiddle: I agree that this function could be improved by sharing *additional* exception details if/when exceptions occur. 

The existing `waitForFunction` helper behavior already causes execution to halt and an error message to be displayed when the returned promise rejects, but error details are not easily accessible (devs would need to include their own `try/catch` and log errors within the function being passed to `waitForFunction`). It would be better to provide error details if/when the promise rejects. This will make it easier to determine if `waitForFunction` rejected because the test function never returned a truthy value or because the test function caused an unexpected error.

It's good feedback. However, this is not the right time or place to discuss it. The code in question is **four years old** and has nothing to do with the lint configuration updates contained in this PR (as @Koooooo-7  has pointed out).

That said, I've updated the `waitForFunction` to address all concerns so we can (hopefully) move forward.

The `waitForFunction` helper now captures the last error within the `catch` block and makes it available if/when the promise returned by `waitForFunction` rejects. Execution should *not* be halted nor should errors be logged to the console within the `catch` block because 1) halting would defeat the purpose of the helper (in this context errors are expected and details are uninteresting compared to the promise being resolve/rejected) and 2) logging errors resulting from each function invocation would spam the console unnecessarily.

Here's what it looked before the update when the `waitForFunction` was rejected as a result of an exception in the test function:

CleanShot 2024-05-27 at 00 49 32@2x

Here's what it looks like now:

CleanShot 2024-05-27 at 00 51 39@2x

@trusktr -- Please review so we can approve, merge, and move on to other tasks. Thx.

Koooooo-7 (approved):
You're really on the ball ! 🚀

jhildenbiddle:
@Koooooo-7 --

Just trying to keep things moving! Thanks for the quick review and helpful input. Much appreciated. :)

src/plugins/disqus.js
trusktr: Ah ok. I didn't run the code, I assumed the line was removed because it no longer errored.

Globals are "bad" because we know we want to move away from them. Prism.js being a global is annoying when it comes to converting to ESM for example. I know we'll keep them for now.

I wouldn't mind at all having warnings for globals (and not failing on warnings, only on errors).

But if you are fine hunting down globals later without a warning helping you, then I can live with it.

src/plugins/search/component.js
trusktr: Oh ok great!

test/helpers/wait-for.js
trusktr: Issue made to check on that later:

- https://github.com/docsifyjs/docsify/issues/2442

trusktr (approved):

src/plugins/disqus.js
jhildenbiddle: > Globals are "bad" because we know we want to move away from them. Prism.js being a global is annoying when it comes to converting to ESM for example.

Understood. Here's my $0.02:

1. Lint warnings are not TODO items. They are intended to be an ephemeral notification that a *potential* issue has been found. The goal is to address the issue and remove the need for the warning sooner than later because--theoretically, at least--people don't want to see lint warnings unnecessarily or for a prolonged period of time.
2. Removing globals is future work we have discussed but nobody has committed to doing or proposed a timeline for other than "after v5". We don't want to repeatedly "warn" developers about existing code that is not an issue today, won't be for the near future (v5), and may never be because as over time the project direction may change, feature/enhancement requirement may change, or perhaps nobody commits to doing the "future" work that justified the warning in the first place.
3. If we want to tag certain lines or blocks of code, I think comments are a better way of doing this. This is already being done in our repo with the `/* html */` comment. We could continue this precedent using a `/* global */` which would provide the same level of assistance with "hunting down globals later".

If/when someone commits to removing globals (perhaps as part of a larger effort) with a reasonable timeline, I think it would be beneficial to enable lint rules that warned about the use of globals 1) on the branch where the development is taking place or 2) in `main` after the work is merged to prevent accidental global usage in the future. Until then, I don't think the use of globals in v4 and v5 is anything contributors need to be "warned" about.

jhildenbiddle:
Thanks @trusktr for the quick follow-up after last week's review. Much appreciated! 🙏

PR #2454, opened 2024-06-13
feat: introduce a text compress for search function

Koooooo-7:
## Summary

The search INDEX we store in `localStorage`, and it has the limitation about 5M.
And we discussed about the `Indexed DB` and other solutions couple years ago.
Today, a sudden inspiration struck me: instead of switching the persistence layer, we can use text compression to address our retrieval needs. 
So I have found some implementations of text compression algorithms，and pick the suitable one of LZW [lz-string](https://pieroxy.net/blog/pages/lz-string/index.html) since we are more on the content/docs. 
Have a test based on our site docs, it reduces the 80% cost.
In short term, it could be a chance for users to make search still works when their docs big but not that too much.

## As-is

Screenshot 2024-06-13 at 20 23 53

## Enable compress

Screenshot 2024-06-13 at 20 25 46

Feature
## For any code change,

Yes
No

No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -9,6 +9,8 @@ By default, the hyperlink on the current page is recognized and the content is s
 <script>
   window.$docsify = {
     search: 'auto', // default
+    // Enable text compression to reduce the size, especially if your contents are very large.
+    largeContent: false, // default
 
     search: [
       '/',            // => /README.md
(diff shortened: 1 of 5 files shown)


Koooooo-7:
closed via #2464 .

PR #2114, opened 2023-07-03
chore: update miscellaneous parts of the source to reasonably modern language alternatives, remove polyfills, improve some JSDoc comments, remove traces of IE

trusktr:
Continues from

- https://github.com/docsifyjs/docsify/pull/2109 (review that one first)

Part of

- https://github.com/docsifyjs/docsify/issues/2104

## **Summary**

Updated a bunch of pieces of code to using modern patterns. Avoided refactoring, mostly changes happened inline, or were small refactorings only.

Some things that we updated:

- `var` -> `let`/`const`
- `function () {}` -> `() => {}`
- `merge()` -> `Object.assign()`
- `a.concat(b)` -> `[...a, ...b]`
- `Object.assign({}, a, b)` -> `{...a, ...b}`
- `(function() {...})()` -> `{...}`
- `[].slice.call(a)` -> `Array.from(a)`
- `hasOwn(a, b)` -> `Object.keys(a).includes(b)`
- `for (let i = 0; i<len; i++) const item = arr[i]` -> `for (const item of arr)`
- `{foo: function() {}}` -> `{foo() {}}`
- `"..." + a + "..."` -> ``` `...${a}...` ```
- etc

Deleted

- CSS var polyfill
- replaceState/pushState fallback
- IE fallbacks

BREAKING: In a minority of cases syntax updates may break apps running in very old browsers (such as Internet Explorer), or apps that build Docsify in a custom way with old build tools. To upgrade, build Docsify for older browsers in a custom way, or update existing build tools to handle newer syntax.

DEPRECATED: `$docsify.themeColor` is deprecated and will be eventually removed, use a `--theme-color` CSS variable in your style sheet.

## **What kind of change does this PR introduce?**

  Code style update
  Refactor
  Docs

## **For any code change,**

- [x] Related documentation has been updated if needed
- [x] Related tests have been updated or tests have been added

## **Does this PR introduce a breaking change?** (check one)

- [x] Yes (see above)
- [ ] No

## **Related issue, if any:**

Simplification and modernization issue: https://github.com/docsifyjs/docsify/issues/2104

## **Tested in the following browsers:**

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] ~~IE~~

PR #1527, opened 2021-03-08
fix: Cannot read property 'classList' of null

sy-records:
**Summary**

**What kind of change does this PR introduce?** (check at least one)

fix https://github.com/docsifyjs/docsify/issues/1526

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Repo settings
- [ ] Other, please describe:

If changing the UI of default theme, please provide the **before/after** screenshot:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)

You have tested in the following browsers: (Providing a detailed version will be better.)

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] IE

If adding a **new feature**, the PR's description includes:

- [ ] A convincing reason for adding this feature
- [ ] Related documents have been updated
- [ ] Related tests have been updated

To avoid wasting your time, it's best to open a **feature request issue** first and wait for approval before working on it.

PR #2457, opened 2024-06-27
fix(search): clean markdown elements in search contents

Koooooo-7:
### Changes
- Process the content before store it into indexDB, hence we don't need parse it every times.
   TODO In v5+: move it as a async job instead of handing main thread too long, especially with large contents. (it looks fine in our site for now)
- Adaption to use `marked v13+` with pure new renderer rewrite.
  - Remove every markdown element stylings.
  - Remove the helper `?>` `!>` of docsify either.
- Copied functions instead import to reduce the package size (import it will block the build min optimize since it is over 500kb).
- Hardcode the `...` to matched contents as truncation surroundings.
- Test cases for the changes.

### Snapshot (before -> after)

Screenshot1
Screenshot2

Bugfix
## For any code change,

Yes
No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


jhildenbiddle (changes requested):
This appears to be working but there are a few issues:

1. There is a noticeable delay when display search results. On macOS 14.5, the delay Is very noticeable using Safari and less noticeable using Chrome. This appears to be a result of storing search data unmodified, then doing a lot of text processing every a search query is performed. Why not do the processing while retrieving the search data and store the result so we only have to do basic text matches on search queries are performed?
2. E2E tests are failing. This may ba result of bringing the branch up to date with `develop`.

paulhibbitts:
Thanks so much @Koooooo-7 for working on this! I've also tried the latest Preview in this thread, which I assume has this change included, and noticed two things:

1) The matching text snippet does not have ellipses (...) at the start of the text which can help reduce the readers understanding of the truncation going on. For example, first result searching for `class` displayed `several CSS classes for customizing your Docsify site. These classes should be...`

2) The highlighting of the found text is on longer happening, which may be an understandable side effect of the merged v5 style updates etc.

I hope the above helps.
Paul

Koooooo-7:
Hi  @jhildenbiddle  @paulhibbitts  ---
Thx for the points.
I think the performance issue does need to resolve.
I will sync with @sy-records  after the #2464  storage layer change merged to do the new storage adaption and performance refine .

Koooooo-7:
> This appears to be a result of storing search data unmodified, then doing a lot of text processing every a search query is performed. Why not do the processing while retrieving the search data and store the result so we only have to do basic text matches on search queries are performed?

Make sense.
I think we could store the formatted data in storage and simple format the search content to do the retrieve instead of format it every time.

Koooooo-7:
@paulhibbitts  
> 1. The matching text snippet does not have ellipses (...) at the start of the text when truncated which can reduce the readers understanding of the truncation going on. For example, the fifth result returned when searching for `class` displayed `erty 'classList' of null (#1527) (d6df2b8), closes...`



> 2. The highlighting of the found text is on longer happening, which may be an understandable side effect of the merged v5 style updates etc.

thx for the nice catch, notes the styling issue. 👌

paulhibbitts (commented):
Thanks so much for the update @Koooooo-7 ! The display of preceding ellipses look good 🙂

I've done some more testing, and have a few comments/questions:

1. Empty ellipses (......) are being displayed when searching for items matching only a Header and no immediate content below. For example, search for "Headings" which is on the UI Kit page. If no content within ellipses perhaps do not display ellipses/content at all?
2. Should we include Markdown image paths/names? For example, search for "icon.svg"?

Koooooo-7:
> 2\. icon.svg



> Thanks so much for the update @Koooooo-7 ! The display of preceding ellipses look good 🙂
> 
> I've done some more testing, and have a few comments/questions:
> 
> 1. Empty ellipses (......) are being displayed when searching for items matching only a Header and no immediate content below. For example, search for "Headings" which is on the UI Kit page. If no content within ellipses perhaps do not display ellipses/content at all?

Nice catch! I didn't aware that there may have a empty search `content`,  I will update it when there is empty content, no `...` display.

> 2. Should we include Markdown image paths/names? For example, search for "icon.svg"?

For now, I keep the images path and names/titles meta, although we can not see it in the content directly.

paulhibbitts (approved):

sy-records (dismissed):
LGTM

sy-records:
There seems to be a problem with diacritics.

image

Koooooo-7:
> There seems to be a problem with diacritics.

I checked the previews behavior, it is different from v4 result, since last year (The previews broken https://github.com/docsifyjs/docsify/pull/2114 but its search part changes looks fine to me). which has a pure wrong result highlight for `cafe`.

Current behavior in this PR more looks like a "patch" to correct search contents, but we still need figure it out when and why the search content changed.

Koooooo-7:
ping @sy-records

sy-records (approved):

PR #2464, opened 2024-07-14
feat(search): use dexie.js instead of localStorage

### LINKED ISSUE #1538: Uncaught DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of 'docsify.search.index' exceeded the quota.
## Bug Report
Uncaught DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of 'docsify.search.index' exceeded the quota.
#### Steps to reproduce

All documents size exceeds 400MB  . exceeded the quota error will occur

```
JSON.stringify(INDEXS).length
14474634
```

You can compact storage

- Your OS: 
- Node.js version: 
- npm/yarn version: 
- Browser version: 
- Docsify version: 
- Docsify plugins: 

### LINKED ISSUE #2281: Feature: Use localForage instead of localStorage
## Feature request

Use localForage instead of localStorage.

#### Problem or desire

Solve the problem that too many search indexes cannot be stored in localStorage.

#### Proposal

https://github.com/localForage/localForage

#### Implementation

Because localForage uses async storage, it has an async API. It's otherwise exactly the same as the [localStorage API](https://hacks.mozilla.org/2009/06/localstorage/).

Should just need to replace localStorage in [`src/plugins/search/search.js`](https://github.com/docsifyjs/docsify/blob/817a7649153e7fe01bca1c719c06733578f59d6a/src/plugins/search/search.js)

----

### Other

- https://dexie.org/

sy-records:
## Summary

use dexie.js instead of localStorage.

Maybe we also need to remove JSON parsing?

## Related issue, if any:

close #2281
close #1538

Feature

## For any code change,

No

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
+        "dexie": "^4.0.8",
         "medium-zoom": "^1.1.0",
         "opencollective-postinstall": "^2.0.2",
         "prismjs": "^1.29.0",
@@ -5928,6 +5929,11 @@
       "engines": {
         "node": ">= 0.8.0"
       }
+    },
+    "node_modules/dexie": {
+      "version": "4.0.8",
+      "resolved": "https://registry.npmjs.org/dexie/-/dexie-4.0.8.tgz",
+      "integrity": "sha512-1G6cJevS17KMDK847V3OHvK2zei899GwpDiqfEXHP1ASvme6eWJmAp9AU4s1son2TeGkWmC0g3y8ezOBPnalgQ=="
     },
     "node_modules/diff-sequences": {
       "version": "29.6.3",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "*.js": "eslint --fix"
   },
   "dependencies": {
+    "dexie": "^4.0.8",
     "medium-zoom": "^1.1.0",
     "opencollective-postinstall": "^2.0.2",
     "prismjs": "^1.29.0",
--- a/src/plugins/search/search.js
+++ b/src/plugins/search/search.js
@@ -13,7 +13,9 @@ db.version(1).stores({
 });
 
 async function saveData(maxAge, expireKey, indexKey) {
-  INDEXES = Object.values(INDEXES).flatMap(innerData => Object.values(innerData));
+  INDEXES = Object.values(INDEXES).flatMap(innerData =>
+    Object.values(innerData),
+  );
   await db.search.bulkPut(INDEXES);
   await db.expires.put({ key: expireKey, value: Date.now() + maxAge });
 }
--- a/src/plugins/search/search.js
+++ b/src/plugins/search/search.js
@@ -12,7 +12,7 @@ db.version(1).stores({
   expires: 'key, value',
 });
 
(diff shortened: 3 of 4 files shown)


jhildenbiddle (changes requested):
@sy-records --

I think this is a great start and on its own addresses the localStorage limitations we were hoping to address by switching to IndexedDB. Well done.

Now that we have a database with rows and columns that we can query, my hope is that we can move away from the localStorage-based technique of storing and parsing one huge JSON string for our search plugin. Doing this adds significant and noticeable blocking tasks to the main thread that negatively impact our user experience and performance ratings (e.g., Lighthouse) which can also have an impact on SEO (even for SPAs like Docsify).

As an example, the docsify.js.org search plugin JSON is 309399 character (~300k). Chrome seems to handle this without issue, but Safari has a very noticeable delay when loading a Docsify site for the first time. The delay in Safari is resolved as soon as I disable the search plugin.

Koooooo-7 (commented):
Hi @sy-records  could you provide more details (thoughts or design) on this changes?
The plain code is hard to see what happens.
Additionally, I hope there could have some benchmark or basic metric to show on what the changes bring.
- Compare the localStorage with the row/column storage in Dexies on the basic usage, e.g. our own site.
- The large content performance (5MB --- localStorage limitation, 10MB, or more?) based on current Dexies solution.

As per our goal is to handle more larger contents and the performance ++.
FYI, Dexies supports the fuzzy search https://github.com/dexie/Dexie.js/blob/master/samples/full-text-search/FullTextSearch.js and other functions. Hope is helps.

Koooooo-7 (changes requested):
Hi @sy-records  any updates on this.
As sync with @jhildenbiddle  (details in discord)

Could you please have a basic performance test on this, hope it could not that worse then we can merge it and move on.
Besides, I suppose you can add more details either in comments or in this PR about this changes.
It will easily for us to trace the code for debugging and handover since we already have lot of `anonymous` codebase.
Thx in advance.

sy-records:
Implemented dexie.js instead of localStorage in this PR.

Removed parsing json strings.

```js
db.version(1).stores({
  search: 'slug, title, body, path, indexKey',
  expires: 'key, value',
});
```

No changes to the search logic, no full-text search using Dexies. Putting indexKey into INDEXES for querying from IndexedDB.

And the storage size, which depends on the user's device and browser limitations, can be tesed using it.

```
if (navigator.storage && navigator.storage.estimate) {
  navigator.storage.estimate().then(estimate => {
    console.log(`Quota: ${estimate.quota}`);
    console.log(`Usage: ${estimate.usage}`);
  });
}
```

image

Koooooo-7 (approved):
Although I have a little question on the changes storage structure, we can have sync later in more details.
Regarding to the search functionality, it works fine.

jhildenbiddle:
@Koooooo-7 -- Can we get your approval so we can merge this PR?

Koooooo-7:
> @Koooooo-7 -- Can we get your approval so we can merge this PR?

Hi @jhildenbiddle  I approved it already.

jhildenbiddle (approved):
LGTM! Thanks, @sy-records.

PR #2467, opened 2024-07-17
chore: upgrade marked from 12.0.2 to 13.0.2

Koooooo-7:
## Summary

Since the markedjs release to v13, all the `renderer`'s params change to TOKENS instead of the pre versions' behaviors.
So, we need manually build the contents as the default renderer of markedjs in pre versions to get 
those contents which we need.

**Refer**

The default renderer impl of markedjs [markedjs/Renderer.ts](https://github.com/markedjs/marked/blob/master/src/Renderer.ts)
The changes before the v13. [markedjs/Instance.ts](https://github.com/UziTech/marked/blob/50b3cc76548b94e1421fb0af803e848102eda40e/src/Instance.ts#L259)
The new params for each TOKEN. [markedjs/Tokens.ts](https://github.com/markedjs/marked/blob/master/src/Tokens.ts) 

close #2461 

## For any code change,

Yes
No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


jhildenbiddle (approved):
Nice work! Happy to see now tests or snapshots needed to be updated :).

sy-records (approved):

Issue #2471, opened 2024-07-19
Support of multiple class names with Markdown helper


paulhibbitts:
#### Steps to reproduce

1. Use Markdown helper to assign multiple classes to link or image, for example with Docsify v5 `[Button Title](https://google.com ':class=button primary')` 

Only the first class in quotes is applied, and the others are ignored.

#### Current behaviour

Currently only the first class included in quotes is applied, and the others are ignored.

#### Expected behaviour

All classes contained in the quotes should be applied.

#### Other relevant information

- Docsify version:

- [x] Bug still occurs when all/other plugins are disabled?

- Docsify plugins (if the bug happens when plugins enabled, please try to isolate the issue):

#### Please create a reproducible sandbox

[![Edit 307qqv236](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/307qqv236)

PR #2472, opened 2024-07-22
refactor: rewrite getAndRemoveConfig(str) function

Koooooo-7:
## Summary

Reactor `getAndRemoveConfig(str)` function to a generic Lexer instead of a complex regex. 
Correct current behavior and indicate docsify needs to resolve valid config keys. Warning invalid configs also.

Remove to use a tricky returned `title` as config options.
Instead, new `KEY_appened_props` introduced. 

It will let the config parse as `:KEY=VALUE append_props...`, each config processor can handle its own logic 
for the `Value` and `append_props`. e.g.

> `:class=classA classB classC :include`
The classA assigns to `class`=`classA`, the rest of `classB classC` add to `class_appened_props`.

> config
```
{
  class: "classA",
  class_appened_props: "classB classC",
  include: true
}
```

Which can support `multi values` configs of one key such as #2471 .

 

Refactor
## For any code change,

Yes
No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/core/render/compiler.js
+++ b/src/core/render/compiler.js
@@ -30,27 +30,27 @@ const compileMedia = {
       url,
     };
   },
-  iframe(url, title) {
+  iframe(url, props) {
     return {
       html: `<iframe src="${url}" ${
-        title || 'width=100% height=400'
+        props || 'width=100% height=400'
       }></iframe>`,
     };
   },
-  video(url, title) {
+  video(url, props) {
     return {
-      html: `<video src="${url}" ${title || 'controls'}>Not Support</video>`,
+      html: `<video src="${url}" ${props || 'controls'}>Not Support</video>`,
     };
   },
-  audio(url, title) {
+  audio(url, props) {
     return {
-      html: `<audio src="${url}" ${title || 'controls'}>Not Support</audio>`,
+      html: `<audio src="${url}" ${props || 'controls'}>Not Support</audio>`,
     };
   },
-  code(url, title) {
+  code(url, props) {
     let lang = url.match(/\.(\w+)$/);
 
-    lang = title || (lang && lang[1]);
+    lang = props || (lang && lang[1]);
     if (lang === 'md') {
       lang = 'markdown';
     }
@@ -143,9 +143,9 @@ export class Compiler {
    * @return {type} Return value description.
    */
   compileEmbed(href, title) {
-    const { str, config } = getAndRemoveConfig(title);
+    const { config } = getAndRemoveConfig(title);
     let embed;
-    title = str;
+    const appenedProps = config.type_appened_props;
 
     if (config.include) {
       if (!isAbsolutePath(href)) {
@@ -158,7 +158,7 @@ export class Compiler {
 
       let media;
       if (config.type && (media = compileMedia[config.type])) {
-        embed = media.call(this, href, title);
+        embed = media.call(this, href, appenedProps);
         embed.type = config.type;
       } else {
         let type = 'code';
@@ -174,7 +174,7 @@ export class Compiler {
           type = 'audio';
         }
 
-        embed = compileMedia[type].call(this, href, title);
+        embed = compileMedia[type].call(this, href, appenedProps);
         embed.type = type;
       }
 
--- a/src/core/render/compiler/image.js
+++ b/src/core/render/compiler/image.js
@@ -27,7 +27,11 @@ export const imageCompiler = ({ renderer, contentBase, router }) =>
     }
 
     if (config.class) {
-      attrs.push(`class="${config.class}"`);
+      let classes = config.class;
+      if (config.class_appened_props) {
+        classes = `${config.config.class} ${config.class_appened_props}`;
+      }
+      attrs.push(`class="${classes}"`);
     }
 
     if (config.id) {
--- a/src/core/render/compiler/image.js
+++ b/src/core/render/compiler/image.js
@@ -29,7 +29,7 @@ export const imageCompiler = ({ renderer, contentBase, router }) =>
     if (config.class) {
       let classes = config.class;
       if (config.class_appened_props) {
-        classes = `${config.config.class} ${config.class_appened_props}`;
+        classes = `${config.class} ${config.class_appened_props}`;
       }
       attrs.push(`class="${classes}"`);
     }
--- a/src/core/render/compiler/link.js
+++ b/src/core/render/compiler/link.js
@@ -55,7 +55,7 @@ export const linkCompiler = ({
     if (config.class) {
       let classes = config.class;
       if (config.class_appened_props) {
-        classes = `${config.config.class} ${config.class_appened_props}`;
+        classes = `${config.class} ${config.class_appened_props}`;
       }
       attrs.push(`class="${classes}"`);
     }
--- a/src/core/render/compiler/link.js
+++ b/src/core/render/compiler/link.js
@@ -10,13 +10,12 @@ export const linkCompiler = ({
 }) =>
   (renderer.link = (href, title = '', text) => {
     const attrs = [];
-    const { str, config } = getAndRemoveConfig(title);
+    const { config } = getAndRemoveConfig(title);
     linkTarget = config.target || linkTarget;
     linkRel =
       linkTarget === '_blank'
         ? compilerClass.config.externalLinkRel || 'noopener'
         : '';
-    title = str;
 
     if (
       !isAbsolutePath(href) &&
@@ -53,15 +52,19 @@ export const linkCompiler = ({
     }
 
     if (config.class) {
-      attrs.push(`class="${config.class}"`);
+      let classes = config.class;
+      if (config.class_appened_props) {
+        classes = `${config.config.class} ${config.class_appened_props}`;
+      }
+      attrs.push(`class="${classes}"`);
     }
 
     if (config.id) {
       attrs.push(`id="${config.id}"`);
     }
 
-    if (title) {
-      attrs.push(`title="${title}"`);
+    if (config.ignore_appened_props) {
+      attrs.push(`title="${config.ignore_appened_props}"`);
     }
 
     return /* html */ `<a href="${href}" ${attrs.join(' ')}>${text}</a>`;
--- a/src/core/render/utils.js
+++ b/src/core/render/utils.js
@@ -6,17 +6,28 @@
  * An example of this is ':include :type=code :fragment=demo' is taken and
  * then converted to:
  *
+ * Specially the exitra values following a key, such as `:propKey=propVal additinalProp1 additinalProp2`
+ * or `:propKey propVal additinalProp1 additinalProp2`, those additional values will be appended to the key with `_appened_props` suffix
+ * as the additional props for the key.
+ * the example above, its result will be `{ propKey: propVal, propKey_appened_props: 'additinalProp1 additinalProp2' }`
+ *
  * ```
+ * [](_media/example.html ':include :type=code text :fragment=demo :class=foo bar bee')
  * {
  *  include: '',
  *  type: 'code',
- *  fragment: 'demo'
+ *  type_appened_props: 'text',
+ *  fragment: 'demo',
+ *  class: 'foo',
+ *  class_appened_props: 'bar bee'
  * }
  * ```
  *
+ * Any invalid config keys will be logged warning to the console intead of swallow them silently.
+ *
  * @param {string}   str   The string to parse.
  *
- * @return {{str: string, config: object}} The original string formatted, and parsed object, { str, config }.
+ * @return {{str: string, config: object}} The string after parsed the config, and the parsed configs.
  */
 export function getAndRemoveConfig(str = '') {
   const config = {};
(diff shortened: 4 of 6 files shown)


jhildenbiddle:
@Koooooo-7 --

1. How does this handle multiple attributes when multiple values are provided?

   ```md
   [Text](http://url ':target=_blank :class=foo bar')
   [Text](http://url ':class=foo bar :target=_blank ')
   ```

   I would have assumed multiple values would need to be wrapped in quotes:

   ```md
   [Text](http://url ':target=_blank :class="foo bar"')
   [Text](http://url ':class="foo bar" :target=_blank ')
   ```

2. Can we not fix the "strict ' or " star/end the configs" by trimming the string found between quotes? If we don't, a single extra space could throw errors which may be hard to track down.

3. Why not just apply unknown attributes and values instead of comparing them against a "known" list?

   ```md
   [Text](http://url ':foo=bar :baz=buzz')
   ```
   ```html
   <a href="..." foo="bar" baz="buzz">Text</a>
   ```

paulhibbitts:
Thanks very much @Koooooo-7 for this! I've put my original CodeSandbox example online with your Docsify Preview build and @jhildenbiddle  examples to help with testing - looking good so far  🙂 https://paulhibbitts.github.io/docsify-v5-preview/#/test-3

Koooooo-7:
> @Koooooo-7 --
> 
> 1. How does this handle multiple attributes when multiple values are provided?
>    ```
>    [Text](page.md ':target=_blank :class=foo bar')
>    [Text](page.md ':class=foo bar :target=_blank ')
>    ```

It will be resolve to 

```
{
  target: "_blank",
  class: foo,
  class_appened_props: "bar"
}
```

Cos the `append props` will look back to the first leading `KEY`.

---

> 2. Can we fix the "strict quote at star/end the configs" by trimming spaces and line endings from the start and end of the string found between quotes? If we don't, a single extra space could throw errors which may be hard to track down.

I see, how about change  to eat the most closed pre/next most `'/"` instead?



> 3. Why not just apply all attribute names and values regardless of whether they are "known"? This would allow people to add things like data attributes to elements in markdown.
>    ```
>    [Text](page.md ':foo=bar :data-baz=buzz')
>    ```

I think it doesn't make sense, since the `:config`  is not a markdown syntax, it enriched by docsify, should only support docsify supported configs.
If user wanna add attributes to elements on markdown,  they should use the `< !-- -->` or other markdown syntax.

Koooooo-7:
Hi @jhildenbiddle  ---
Already update:
- Support loose quote `'   :class=any :include        '`
- Support unknown attributes ` ':class=foo :foo=bar' `

And I add more test cases and comments on those cases we discussed.

paulhibbitts:
Thanks so much @Koooooo-7 , this is going to be a really nice improvement 🙂 I've updated my online Docsify Preview build with your latest Preview and things look good! https://paulhibbitts.github.io/docsify-v5-preview/#/test-3

Koooooo-7:
> Thanks so much @Koooooo-7 , this is going to be a really nice improvement 🙂 I've updated my online Docsify Preview build with your latest Preview and things look good! https://paulhibbitts.github.io/docsify-v5-preview/#/test-3

Thx for you live preview for the changes !!!

jhildenbiddle (changes requested):
Suggestion on this:

> > 1. How does this handle multiple attributes when multiple values are provided?
> >    ```
> >    [Text](page.md ':target=_blank :class=foo bar')
> >    [Text](page.md ':class=foo bar :target=_blank ')
> >    ```
> 
> It will be resolve to
> 
> ```
> {
>   target: "_blank",
>   class: foo,
>   class_appened_props: "bar"
> }
> ```

Instead of capturing multiple values using  `class` and `class_appended_props` properties, why not require multiple values to be wrapped in quotes? This seems more natural since it is how attributes and values work in HTML. It will also make it easier to read multiple attribute values because all values are contained in a single property.

For example:

```md
[Text](page.md ':foo :target=_blank :class="bar baz" This is the title')
```
```js
{
  foo: true,
  target: "_blank",
  class: "bar baz",
  title: "This is the title" // links support title attribute in markdown
}
```

The rules would then be:

1. All attributes must start with `:`
2. Attributes without values are valid (`:foo`)
3. Quotes are optional for single values (`:target=_blank` or `:target="_blank"`)
4. Quotes are required for multiple attribute values (`:class="bar baz"'`)

The question is then if/how we want to handle stray, unquoted values. Consider the following example:

```md
[Text](page.md ':foo=val1 val2 :bar=val3 val4'). // Missing quotes around multiple values
```

I would expect this to produce the following:

```js
{
  foo: "val1",
  bar: "val3",
  title: "val4",
}
```

What happens to `val2`? 

The easiest thing to do would be to just drop `val2` because according to the rules this is neither an attribute (does not start with `:`) or an attribute value.

Thoughts?

Koooooo-7:
> Instead of capturing multiple values using  `class` and `class_appended_props` properties, why not require multiple values to be wrapped in quotes? This seems more natural since it is how attributes and values work in HTML. It will also make it easier to read multiple attribute values because all values are contained in a single property.
> 
> For example:
> 
> ```md
> [Text](page.md ':foo :target=_blank :class="bar baz" This is the title')
> ```
> ```js
> {
>   foo: true,
>   target: "_blank",
>   class: "bar baz",
>   title: "This is the title" // links support title attribute in markdown
> }
> ```
> 
> The rules would then be:
> 
> 1. All attributes must start with `:`
> 2. Attributes without values are valid (`:foo`)
> 3. Quotes are optional for single values (`:target=_blank` or `:target="_blank"`)
> 4. Quotes are required for multiple attribute values (`:class="bar baz"'`)
> 


That would be more clearly. but it brings a breaking change for all the config rules for now. So I choose a compromise way to make it and doesn't breaking anything.


> The question is then if/how we want to handle stray, unquoted values. Consider the following example:
> 
> ```md
> [Text](page.md ':foo=val1 val2 :bar=val3 val4'). // Missing quotes around multiple values
> ```
> 
> I would expect this to produce the following:
> 
> ```js
> {
>   foo: "val1",
>   bar: "val3",
>   title: "val4",
> }
> ```
> 
> What happens to `val2`? 
> 
> The easiest thing to do would be to just drop `val2` because according to the rules this is neither an attribute (does not start with `:`) or an attribute value.
> 
> Thoughts?

It is hard to say a date attribute nor a docsify config within a string.

If we decided include a breaking change on this. I think we can introduce a docsify config block. e.g.


> ```md
> [Text](page.md ' docsify=[:foo :target=_blank :class="bar baz"]  This is the title')
> ```

Then, we have a clear line to distinguish between the docsify part and custom made for markdown owned.

jhildenbiddle:
> That would be more clearly. but it brings a breaking change for all the config rules for now. So I choose a compromise way to make it and doesn't breaking anything to get ride of the regex.
> 
> Besides, the config of docsify is not that consistent. 

Makes sense. I didn't realize that the attribute parser is inconsistent across tags. I appreciate trying to not introduce a non-breaking change, but perhaps this is the opportunity make attribute parsing consistent in docsify for all markdown tags that support it. See below...

> If we decided include a breaking change on this. I think we can introduce a docsify config block. e.g.
> 
> > ```
> > [Text](page.md ' docsify=[:foo :target=_blank :class=bar baz]  This is the title')
> > ```
> 
> Then, we have a clear line to distinguish between the docsify part and custom made for markdown owned.

Can we add something like the proposed "block" of attributes above as a new v5 feature but retain the existing `:prop=val` behavior for backwards compatibility? This would allow us to update our docs with the new "preferred" method alongside the older `:prop=name` method, but note in the docs that the older method is deprecated in v5 and will be removed in v6.

For example, we can drop the `docsify=` and use the `:` as the start of the pattern:

```md
[Text](page.md ':[foo target="_blank" class="bar baz"]  This is the title')
```

Alternatively, we could use single or double brackets to start/end the pattern which would make it easy for us to parse using `JSON.stringify` once the string was extracted:

```md
[Text](page.md '{foo: true, target: "_blank" class: "bar baz"}  This is the title')
```

```md
[Text](page.md '{{foo: true, target: "_blank" class: "bar baz"}}  This is the title')
```

If we go this route, I would propose we *not* allow mixing new and old attribute styles like this:

```md
[Text](page.md '{{foo: true, class: "bar baz"}} :target=_blank This is the title')
```

Doing this promotes continued use of the old `:prop=name` style. When this happens, we can either handle both old and new styles with a warning in the console, or we can process *only* the new style and ignore the older `:prop` style.

This seems like a better approach than updating  the `getAndRemoveConfig` function with new capabilities that only work in some scenarios and don't address Docsify's inconsistencies in markdown attribute parsing.

Koooooo-7:
--- @jhildenbiddle 
> This seems like a better approach than updating  the `getAndRemoveConfig` function with new capabilities that only work in some scenarios and don't address Docsify's inconsistencies in markdown attribute parsing.

Yes, there is a `plain` refactor of current implementation without any changes.
The purpose of the changes for me is to refactor the `regex` to a easy to handle `lexer` first. 
Then we could update it to a new configs rules and strict to what we want as a docsify configuration.

---
Based on this refactor doesn't breaking any current behavior, let's think about the new config functions.

> Can we add something like the proposed "block" of attributes above as a new v5 feature but retain the existing :prop=val behavior for backwards compatibility? This would allow us to update our docs with the new "preferred" method alongside the older :prop=name method, but note in the docs that the older method is deprecated in v5 and will be removed in v6. 


First of all, I agree the notes way to reminder users to update its configs if needs in v5 and future, give user a buffer to know the changes.  ✅

---

About how to do the redesign about the `docsify helper configs`, tbh, I don't have a very clear assumption on this for now.
Don't worry about what it should be in future and what it is for now. I believe we can handle them in the v5 with compatibility. 

On your ideas.

A: 

```
[Text](page.md ':[foo target="_blank" class="bar baz"]  This is the title')
```

It is okay, `:[` is the config heading   and the `:[]` has a clear config block.
The extra effort for us is the `quote` since it means user can config in:
```
single quote  ------   ':[foo target="_blank" class="bar baz"]  This is the title'
double quote ------  ":[foo target="_blank" class='bar baz']  This is the title"
```

There contains a extra quote validation between the markdown config block and the "values group". Although we have to face it currently, I suppose we'd better drop it if we could. 
How about this:

`[Text](page.md ':[ :foo :target=_blank :class=bar baz]  This is the title')`

- the `:[]` is the config block, `:[` is the config heading.
- keep the `:Key` rules for each config key.
- remove the `"`, all the values after `:Key=` will append to the leading `:Key` until to the end or meet another `:Key2`.


---- 

B: 
> Alternatively, we could use single or double brackets to start/end the pattern which would make it easy for us to parse using JSON.stringify once the string was extracted:
```
[Text](page.md '{foo: true, target: "_blank" class: "bar baz"}  This is the title')
[Text](page.md '{{foo: true, target: "_blank" class: "bar baz"}}  This is the title')
```
Personally, I think this is a good idea. what we need we is getting the `{}` block and do a JSON deserialization directly. 💯 
The only concern is on the modern mvvm front-end libs. Almost all use the `{{  msg }}` as a value placeholder.
I'm not quite sure if we use the `{}`/`{{}}` could be a problem for user integration or something.
If not, I prefer this one. 👍

jhildenbiddle:
I think both styles have pros and cons:

- Colon style (`:foo=bar`) is clearly unique to Docsify and therefore should not impact utilities which interpret the brackets to as JavaScript like [MDX](https://mdxjs.com). Colon style also allows applying valueless attributes: 
  ```md
  [Text](page.md ':[data-foo] My Title)` = `<a href="..." data-foo title="My Title">
  ```
- Bracket style (`{..}`or `{{...}}`) is easier for us to parse and familiar to developers.

Here is some additional feedback on each style:

### Colon Style

> ```
> [Text](page.md ':[foo target="_blank" class="bar baz"]  This is the title')
> ```
> 
> It is okay, `:[` is the config heading and the `:[]` has a clear config block. The extra effort for us is the `quote` since it means user can config in:
> 
> ```
> single quote  ------   ':[foo target="_blank" class="bar baz"]  This is the title'
> double quote ------  ":[foo target="_blank" class='bar baz']  This is the title"
> ```

As you mention, this is something we already have to handle today. All of the following examples are valid regardless of how we handle attributes:

```md
[Text](page.md 'This is the title')
[Text](page.md "This is the title")
[Text](page.md "This is the title with 'single quotes' and a \"double quotes\"")
```

Even if we went with "bracket style", we'd still have to handle single vs. double quotes:

```md
[Text](page.md '{foo: "bar"} This is the title')
[Text](page.md "{foo: 'bar'} This is the title")
[Text](page.md "{foo: 'bar', baz: \"buzz\"} This is the title with 'single quotes' and a \"double quotes\"")
```

> There contains a extra quote validation between the markdown config block and the "values group". Although we have to face it currently, I suppose we'd better drop it if we could. How about this:
> 
> `[Text](page.md ':[ :foo :target=_blank :class=bar baz] This is the title')`

This format feels like we're forcing users to work around how our parser works instead of designing our parser to work the way that makes the most sense to users.

The closer we can get to the familiar `name="value"` HTML format or the `name: "value"` JS/JSON format, the better. The point of the "config block" format below is to align with this goal:

```md
`[Text](page.md ':[foo target="_blank" class="bar baz"] This is the title')`
```

### Bracket Style

> > Alternatively, we could use single or double brackets to start/end the pattern which would make it easy for us to parse using JSON.stringify once the string was extracted:
> > 
> > ```
> > [Text](page.md '{foo: true, target: "_blank" class: "bar baz"}  This is the title')
> > [Text](page.md '{{foo: true, target: "_blank" class: "bar baz"}}  This is the title')
> > ```
> 
> Personally, I think this is a good idea. what we need we is getting the `{}` block and do a JSON deserialization directly. 💯 The only concern is on the modern mvvm front-end libs. Almost all use the `{{ msg }}` as a value placeholder. I'm not quite sure if we use the `{}`/`{{}}` could be a problem for user integration or something. If not, I prefer this one. 👍

My understanding is that single brackets (`{...}`) are typically used for string interpolation while double brackets (`{{...}}`) typically indicate JavaScript to be parsed/executed. Given that, I think double-brackets are the better option:

```md
[Text](page.md '{{foo: true, target: "_blank" class: "bar baz"}}  This is the title')
```

### Final Thoughts 

I am open to adding either colon style or bracket style as a new v5 feature, however I do not think we should promote the ability to specify multiple value with the previous `:name=value` format. If we are going to introduce a new method of assigning attributes and values because we believe it is better than the old way, then let's encourage users to adopt the new method so we can deprecate (and eventually remove) the older method.

If we go this route, do the changes in this PR (regex => lexer) still make sense? If so, we can keep them. If not, then let's focus on getting the new-and-improved format in place instead of "improving" code which we hope to remove in the next release.

paulhibbitts:
Thanks so much @Koooooo-7 @jhildenbiddle for working this out some more. My feeling are also mixed re: colon or brackets style...  the use of [] is already in Markdown but that also means they are now used in multiple ways in a single line of Markdown which might be less than clear and not expected.

To confirm, if this change went ahead existing single class assignments would still work - correct?

jhildenbiddle:
> To confirm, if this change went ahead existing single class assignments would still work - correct?

Correct. My proposal is that keep the current behavior and add new functionality via a new attribute/value format.

paulhibbitts:
BTW, after coming across other examples of Markdown dialect classes (https://commonmark.thephpleague.com/2.4/extensions/attributes/ and https://masteringlaravel.io/daily/2023-10-05-applying-a-class-to-a-markdown-link) I am now tending even more to the use of {}. These examples were new to me 🙂

Koooooo-7:
--- @jhildenbiddle 
> ### Final Thoughts
> I am open to adding either colon style or bracket style as a new v5 feature, however I do not think we should promote the ability to specify multiple value with the previous `:name=value` format. If we are going to introduce a new method of assigning attributes and values because we believe it is better than the old way, then let's encourage users to adopt the new method so we can deprecate (and eventually remove) the older method.

I think the `{{}}` if acceptable for me when the `"/'` inside the markdown contents is fine, thx for the details.

> 
> If we go this route, do the changes in this PR (regex => lexer) still make sense? If so, we can keep them. If not, then let's focus on getting the new-and-improved format in place instead of "improving" code which we hope to remove in the next release.

If we decide to replace it as a brand-new configs styles for users, I think this `improving` is not that necessary since we will replace it totally instead of a updating.
Instead, I will raise an issue on this for the new config ways in docsify.

Koooooo-7:
close via #2476 2476

jhildenbiddle:
@paulhibbitts --

The main issue with the CommonMark extension format used in the links you provided is that it makes the markdown far less portable. If the markdown parser is does not support the extended syntax, the `{..}` will appear in the rendered output.

For example, the following markdown:

```md
[Some Text](some-url){ .foo .bar }
```

Will render as following on GitHub:



With either of the formats being proposed above, the extended syntax will not render as text but will be visible on hover as part of the `title` attribute: 



Document portability is arguably *the* most critical feature of markdown, so I personally am not a fan of extensions which result in unsupported syntax rending as content when moved between markdown environments.

Just FYI. :)

paulhibbitts:
Oh thanks for the additional info @jhildenbiddle , I was not aware of that factor in regards to the choice of single or double brackets!

Koooooo-7:
> ps - as a follow up test, it looks like `[Text](page.md ':[foo target="_blank" class="bar baz"] This is the title') ` renders in GitHub with the additional formatting displayed as a hover https://github.com/paulhibbitts/docsify-v5-preview/blob/main/docs/test-3.md

Yes, cos we haven't support it yet.

jhildenbiddle:
> @paulhibbitts ”
> ps - as a follow up test, it looks like `[Text](page.md ':[foo target="_blank" class="bar baz"] This is the title') ` renders in GitHub with the additional formatting displayed as a hover https://github.com/paulhibbitts/docsify-v5-preview/blob/main/docs/test-3.md

Correct. The markdown standard states that anything contained within the quotes will be included as part of the title attribute. It won’t matter which format we choose. As long as out “configuration block” is contained within the quotes, it will be included as part of the title value when the markdown is viewed in an environment other than docsify (like GitHub).

PR #2478, opened 2024-07-31
fix: sync the page title regarding the title config

Koooooo-7:
## Summary
Currently, on the v5 changes, each headings/links on the sidebar has the `title` via `#addTextAsTitleAttribute`.
https://github.com/docsifyjs/docsify/blob/42f25482fa1ea6ec666800145f6da33e9119aa8c/src/core/render/index.js#L20

After the `event` refactor, the page `title` update function changed, especially the search page title.

---
Example from our site.
v4:
   On the main content  load the page title such as `Configuration`, no matter you switch the sections to `Configuration`.
   or directly jump to its sub headings [configuration?id=alias](https://docsify.js.org/#/configuration?id=alias).

    The page title is always `Configuration`.

v5 as-is:
   On the main content  load the page title such as `Configuration`, switch the sections to `Configuration`. (same to v4 )
   :warning:  directly jump to its sub headings [configuration?id=alias](https://docsify-preview-git-fork-eldemcan-patch-1-docsifyjs.vercel.app/preview/#/configuration?id=alias) via the URL，it will update the page title to `sub title` (`alias`) instead.
   ⚠️  On the search function, when clicks the search result, it comes to the mess innerText as page title instead. 

The behavior for the sidebar headings for main content, I think it makes sense to indicate the specific target and respect the title.
So, I fix the search problem via a consistent way that all retrieve page title from to the `title` attribute also.
Now, it will use the `${name} - ${title}` as the page title, if no name config, use title directly.

Bugfix
## For any code change,

Yes
No

No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/core/event/index.js
+++ b/src/core/event/index.js
@@ -299,10 +299,17 @@ export function Events(Base) {
      * @void
      */
     onRender() {
+      const { name } = this.config;
       const currentPath = this.router.toURL(this.router.getCurrentPath());
-      const currentTitle = dom.find(
-        `.sidebar a[href='${currentPath}']`,
-      )?.innerText;
+      const currentSection = dom
+        .find(`.sidebar a[href='${currentPath}']`)
+        ?.getAttribute('title');
+
+      const currentTitle = name
+        ? currentSection
+          ? `${name} - ${currentSection}`
+          : name
+        : currentSection;
 
       // Update page title
       dom.$.title = currentTitle || this.#title;
--- a/src/core/event/index.js
+++ b/src/core/event/index.js
@@ -307,7 +307,7 @@ export function Events(Base) {
 
       const currentTitle = name
         ? currentSection
-          ? `${name} - ${currentSection}`
+          ? `${currentSection} - ${name}`
           : name
         : currentSection;
 
--- a/src/plugins/search/component.js
+++ b/src/plugins/search/component.js
@@ -49,9 +49,10 @@ function doSearch(value) {
 
   let html = '';
   matches.forEach((post, i) => {
+    const title = (post.title || '').replace(/<[^>]+>/g, '');
     html += /* html */ `
       <div class="matching-post" aria-label="search result ${i + 1}">
-        <a href="${post.url}">
+        <a href="${post.url}" title="${title}">
           <p class="title clamp-1">${post.title}</p>
           <p class="content clamp-2">${post.content}</p>
         </a>


trusktr:
Hmm, I see yeah it is inconsistent depending how you land on the page. Clicking on your link, the title is `alias`, but if you navigate to the page from inside Docsify, the title is `Configuration`.

Also I think that it would be better for the title to be `Docsify - Page` for each page whre `Page` is the name of the page, or similar. F.e. `Docsify - Configuration`, `Docsify - Quick start`, etc.

Koooooo-7:
> Hmm, I see yeah it is inconsistent depending how you land on the page. Clicking on your link, the title is `alias`, but if you navigate to the page from inside Docsify, the title is `Configuration`.
> 
> Also I think that it would be better for the title to be `Docsify - Page` for each page whre `Page` is the name of the page, or similar. F.e. `Docsify - Configuration`, `Docsify - Quick start`, etc.

Yea, It makes sense that we could concat the `name` of the site as a prefix,

src/core/event/index.js
```diff
@@ -296,13 +296,20 @@ export function Events(Base) {
...
-      const currentTitle = dom.find(
-        `.sidebar a[href='${currentPath}']`,
-      )?.innerText;
+      const currentSection = dom
+        .find(`.sidebar a[href='${currentPath}']`)
+        ?.getAttribute('title');
+
+      const currentTitle = name
+        ? currentSection
+          ? `${name} - ${currentSection}`
```
sy-records, line 310: I suggest putting the site title after the page title.

```suggestion
          ? `${currentSection} - ${name}`
```

sy-records (dismissed):

src/core/event/index.js
```diff
@@ -296,13 +296,20 @@ export function Events(Base) {
     // =========================================================================
     /**
      * Handle rendering UI element updates and new content
+     * @param {string} name the site name
```
sy-records, line 299:
```suggestion
```

src/core/event/index.js
```diff
@@ -296,13 +296,20 @@ export function Events(Base) {
...
      * @void
      */
-    onRender() {
+    onRender(name) {
       const currentPath = this.router.toURL(this.router.getCurrentPath());
```
sy-records, line 303: Why not use it like this?

```suggestion
    onRender() {
      const { name } = this.config;
      const currentPath = this.router.toURL(this.router.getCurrentPath());
```

Koooooo-7: update.

src/core/event/index.js
Koooooo-7: I think the format are both fine, and I test them locally.
Personally I agree with @trusktr  to do something like `docsify - Configuration` to mention the site first.
Cos I checked some of the v4 sites behaviors such as the [docsify-themeable/QuickStart](https://jhildenbiddle.github.io/docsify-themeable/#/quick-start?id=new-site), they looks fine.

Maybe we could have a discuss on this.

Koooooo-7: Also, we could introduce a `pageTitleFormatter` config, by default use the `name - title`, users can concat them on what they want in v5+.
The Formatter kinda like:
```js
<script>
  window.$docsify = {
    pageTitleFormatter: (config, section) => { },
  
  };
</script>
```

sy-records (approved):

src/core/event/index.js
paulhibbitts: I would also prefer page title first, then site title, as page name is often the more valuable source of contextual info (i.e. looking at open tab titles, list of history, bookmarks etc.).

paulhibbitts (approved):

sy-records (approved):

BrysonC:
Since the `name` configuration can "also contain custom HTML for easier customization" (due to its primary/first use in the sidebar) using the name for page title generation can produce some ugly results.

I liked @Koooooo-7 recommendation to make a `pageTitleFormatter` config option.

PR #2546, opened 2025-05-05
chore: retire outdate multiple languages

sy-records:
We are about to release version V5.0.0-RC1, and since the documentation for these languages is not up to date and I am not sure if the previous maintainer is willing to continue to maintain it, I will ping here and if there is time to continue to maintain it, then it can be added back.

> btw, while we can use Ai for translations, we lack native language speakers for confirmation, and if there is no one to maintain it, then I'd still recommend removing it.
> We also welcome new contributors! ❤️

- [es](https://github.com/docsifyjs/docs-es) @SidVal
- [ru](https://github.com/docsifyjs/docs-ru) @lex111
- [de](https://github.com/docsifyjs/docs-de) @SidVal

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -17,6 +17,8 @@ if [[ $REPLY =~ ^[Yy]$ ]]; then
 
   # test
   npm run test
+  # update snapshot
+  npm run test:update:snapshot
 
   # commit
   git add -A
--- a/docs/_navbar.md
+++ b/docs/_navbar.md
@@ -4,6 +4,3 @@
 
   - [English](/)
   - [简体中文](/zh-cn/)
-  - [Deutsch](/de-de/)
-  - [Español](/es/)
-  - [Русский](/ru-ru/)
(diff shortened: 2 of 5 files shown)


paulhibbitts:
Thanks for this @sy-records. What are your thoughts about the zh-cn translation, which also looks out of date (at least when I try to display it and then translate back to English)?

sy-records:
We use crowdin, Translations can now be done at [translate.docsifyjs.org](https://translate.docsifyjs.org/)

Koooooo-7 (approved):

SidVal:
> We use crowdin, Translations can now be done at [translate.docsifyjs.org](https://translate.docsifyjs.org/)

Great! I didn't know that — a new task for me.

By the way, it's **Spanish** only.
I don't translate from "DE" (German), sorry for the confusion.

sy-records:
Thanks for clarifying. @SidVal

PR #2549, opened 2025-05-11
chore: add depth configuration for search

sy-records:
There is h6 content in the document, but the search configuration defaults to h2, so the relevant content is not searched.


paulhibbitts (approved):
LGTM.

PR #2568, opened 2025-07-15
do not cut off bottom portion of text

dajrivera:
## Summary

The bottom portion of texts get cut off on Edge, Chrome, & Firefox; this issue does not appear in Safari. Using CSS property `line-height: normal` seems to fix the issue and does not affect Safari's rendering of the page.

before:
image

afeter:
image

## Related issue, if any:

N/A

  Other: minor UI fix

## For any code change,

No

- [X] Chrome
- [X] Firefox
- [X] Safari
- [X] Edge


sy-records:
Thanks for your contribution.

But I think this is by design, and you can modify it through the theme properties.

```css
      :root {
        --heading-line-height: normal;
      }
```

paulhibbitts:
I'd tend to include this fix, as by default I would expect line-height to be automatic by default in all Browsers.

dajrivera:
@sy-records ah, thanks; I did not check for an existing variable. I set the global variable and it did indeed fix the problem but as  @paulhibbitts commented, I would expect the text to look correct by default in all browsers. I've moved the change to be the default `--heading-line-height` variable value so it can still be overridden by themes.

paulhibbitts:
Thanks @dajrivera , now that I think about it I may also have run across this issue earlier when trying different fonts out with v5 and noticing I needed to then further adjust line-height to avoid slight cut-off one some particular ones.

src/themes/shared/_vars.css
```diff
@@ -115,7 +115,7 @@
   --form-element-color            : ;
   --heading-color                 : var(--strong-color);
   --heading-font-weight           : 600;
-  --heading-line-height           : calc(2ex + 5px); /* Unit required */
+  --heading-line-height           : normal; /* Unit required */
```
sy-records, line 118:
```suggestion
  --heading-line-height           : normal;
```

paulhibbitts (approved):
LGTM

sy-records:
I think there might be a problem.

image

dajrivera:
@sy-records looks like the h2 bottom-margin is calculated and was the reason for the prior "Unit required" comment; using value normal here broke that calculation. I've updated the bottom-margin calculation using the line height unit instead. At this point the bottom margin was too large so I included a local line-height-multiplier to reduce the margin size as a workaround. Let me know if you can think of a better alternative or a more appropriate method to implement this.

paulhibbitts:
Good catch @sy-records , my apologies for not viewing additional pages in the PR preview... I will be more careful in the future.

dajrivera:
@sy-records @paulhibbitts Please review this again. I've changed the approach to only allow y overflow on h1 & h2 headings. Tested this on Edge, Chrome, Firefox, and Safari and all look display the fonts without issue without additional changes to the rest of the UI.

If OK, let me know I can clean up the commits.

paulhibbitts:
No worries, now the preview looks identical to John's original layout ✅

Here is the current Docsify Preview:


And here is the updated PR Preview:


All looks good to me!

PR #2575, opened 2025-07-23
fix: handle hash navigation to prevent duplicate callbacks

### LINKED ISSUE #522: Always scroll to clicked sidebar content
Docsify does not scroll to the clicked sidebar link content if the link was the last sidebar link clicked.

1. Visit the [Markdown](https://jhildenbiddle.github.io/docsify-themeable/#/markdown) section on the [docsify-themeable](https://jhildenbiddle.github.io/docsify-themeable/) site
1. Click any section link (Ex: [Code](http://localhost:3000/#/markdown?id=code))
1. Scroll the page
1. Click the same section link

I believe the expectation is that docsify would scroll back to the section clicked. 

Perhaps removing the hash from the URL, then updating it with the section hash would provide a quick fix?

sy-records:
## Summary

This PR improves hash navigation handling by correctly setting the navigating flag even when the clicked link points to the current hash.

## Related issue, if any:

Close #522

Bugfix

## For any code change,

No

- [x] Chrome
- [ ] Firefox
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/core/router/history/hash.js
+++ b/src/core/router/history/hash.js
@@ -47,6 +47,15 @@ export class HashHistory extends History {
 
       if (el && el.tagName === 'A' && !isExternal(el.href)) {
         navigating = true;
+
+        // Do not compare hash containing these classes.
+        if (['app-name-link', 'page-link'].includes(el.className)) {
+          return;
+        }
+
+        if (el.hash === location.hash) {
+          cb({ event: e, source: 'navigate' });
+        }
       }
     });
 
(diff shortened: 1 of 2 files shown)


src/core/router/history/hash.js
```diff
@@ -46,7 +46,14 @@ export class HashHistory extends History {
       const el = e.target.tagName === 'A' ? e.target : e.target.parentNode;
 
       if (el && el.tagName === 'A' && !isExternal(el.href)) {
+        const hrefHash = new URL(el.href).hash;
+        const currentHash = location.hash;
```
Koooooo-7, line 50: y not directly use `el.hash`?

PR #2576, opened 2025-07-23
fix: remove block display from anchor links to prevent layout issues

sy-records:
see https://preview.docsifyjs.org/#/changelog

image

also fix https://github.com/erickjx/docsify-fontawesome/issues/6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,19 @@
-# [5.0.0-rc.1](https://github.com/docsifyjs/docsify/compare/v4.13.1...v5.0.0-rc.1) (2025-05-27)
+# Changelog
+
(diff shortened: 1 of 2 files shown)


paulhibbitts:
Thanks very much @sy-records , I've now updated the earlier Docsify v5 Font Awesome plugin test site with the PR Preview and you can see things are looking good!

Using updated PR Preview:
https://paulhibbitts.github.io/docsify-v5-font-awesome-test/#/font-awesome

Earlier test CSS you provided:
https://paulhibbitts.github.io/docsify-v5-font-awesome-test/#/font-awesome-fix-1

So from my perspective everything looks good so no changes now required in plugin, correct? cc @erickjx

sy-records:
yep

PR #2577, opened 2025-07-24
docs: update sidebar link formatting

sy-records:
## Summary

There are some issues with clicking on awesome and changelog. https://preview.docsifyjs.org/#/awesome

The solution is to remove the p tag from the top level `<li><p><a></a></p></li>`.

Bugfix

## For any code change,

Yes

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge


Koooooo-7 (commented):
I think the root cause may be the styling since the `display:none` apply to all the non `a` elements.
As a single line in `_sidebar.md` , which does't as a group, we need reconsider how to structure the sidebar and refine the css selector.

https://github.com/docsifyjs/docsify/blob/be2163725ae314de3aa9a56c1216ce8c774cfcc6/src/themes/shared/_sidebar.css#L91

```css
&.collapse {
      > :not(a) {
        display: none;
      }
    }
```

cc @jhildenbiddle

Koooooo-7 (approved):

PR #2578, opened 2025-07-24
refactor: simplify auto header insertion logic

### LINKED ISSUE #810: autoHeader not compatible with plugins that write before md content
`autoHeader` looks for the first node in `#main` and if it's an `h1` then it does nothing, if not it will create the auto header.
Fine.

The issue arises when using any plugin writing content **before** the user defined markdown.
A plugin like ["Edit this Page" link plugin](https://github.com/cyrilf/docsify-edit-link) for instance.

`autoHeader` doesn't detect my `H1` (because the edit link is first) and create an header.
The [line responsible is here](https://github.com/docsifyjs/docsify/blob/1a945d445a636d778b242f9d74af4f72c7d89e16/src/core/render/index.js#L116-L117).

Screenshot from 2019-04-06 18-28-54
_(in this screenshot `Introduction` shouldn't be here as `My super title` is defined.)_

A solution could be that instead of [looking if the 1st element is an H1](https://github.com/docsifyjs/docsify/blob/1a945d445a636d778b242f9d74af4f72c7d89e16/src/core/render/index.js#L116-L117) we could do instead something like: `dom.getElementsByTagName('h1').length` and trigger our `autoHeader` if it's equal to `0`.

Otherwise we need to find a better way of making the distinction between user content and content added via plugins. But I'm not familiar enough with `docsify` codebase to find out at the moment.

----

If you think this my first solution is acceptable, then I could open a PR.

sy-records:
- Use querySelector instead of getElementsByTagName for clarity
- Avoid redundant DOM wrapping by directly accessing firstElementChild
- Add null check to prevent potential DOM insertion errors

Close #810

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -72,7 +72,9 @@ window.$docsify = {
 - Type: `Boolean`
 - Default: `false`
 
-If `loadSidebar` and `autoHeader` are both enabled, for each link in `_sidebar.md`, prepend a header to the page before converting it to HTML. See [#78](https://github.com/docsifyjs/docsify/issues/78).
+If `loadSidebar` and `autoHeader` are both enabled, for each link in `_sidebar.md`, prepend a header to the page before converting it to HTML — but only if the page does not already contain an H1 heading.
+
+For more details, see [#78](https://github.com/docsifyjs/docsify/issues/78).
 
 ```js
 window.$docsify = {
--- a/src/core/render/index.js
+++ b/src/core/render/index.js
@@ -342,11 +342,15 @@ export function Render(Base) {
 
       if (autoHeader && activeEl) {
         const main = dom.getNode('#main');
-        const firstNode = main.children[0];
-        if (firstNode && firstNode.tagName !== 'H1') {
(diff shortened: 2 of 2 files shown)


Koooooo-7 (commented):
I think the `main.children[0]` seems more reasonable/semantic to determine if there contains a `Header` (although other plugin may broken it, maybe we need distinguish docsify owned doms).
 If we change to search `h1` in whole content, it seems we breaks the behavior since we don't allow user to use autoHeader for those content which contains h1 for different purpose.

```md
// # AutoHeader add from _sidebar.md
this is a summary for the whole content.

# There is a new section for the next content below
Should I never use the H1 in this content anymore? 
```

Maybe we need keep this behavior? or update the docs for the `autoHeader`, but it will constrict the users' content tho.

sy-records:
Searching for `a` does not solve the problem.
image

Koooooo-7:
> Searching for `a` does not solve the problem.
> image
> 

I understand that, since it's strict for the first item, which may exchanged by plugins. It's fine to change the behavior in v5 that we only support autoHeader in the content without any H1 heading presently.

PR #2582, opened 2025-07-28
feat: support extract content between fragment markers from Markdown

### LINKED ISSUE #1973: Enable fragment for non-code
## Feature request

It seems like the "Embed code fragments" feature only works for code, but it would be nice to generalize it to include rendered Markdown.

I would like to generate a single file with a bunch of snippets like:

```md
/// [chunk1]
<p>Chunk 1</p>
/// [chunk1]

/// [chunk2]
<p>Chunk 2</p>
/// [chunk2]
```

And then in the relevant page, be able to do

```md
[chunk](./chunks.md ':include :fragment=chunk1')
```

See above

Break out this block into a function `resolveFragment(token, text: string) => string`

https://github.com/docsifyjs/docsify/blob/c49c39a4a2d6528ec27a314ab5fdc1e274af685f/src/core/render/embed.js#L49-L55

Call the function in every conditional block immediately before passing `text` to `embedToken`

Sure

sy-records:
Fix #1973

### DIFF
--- a/src/core/render/embed.js
+++ b/src/core/render/embed.js
@@ -3,6 +3,29 @@ import { get } from '../util/ajax.js';
 
 const cached = {};
 
+/**
+ * Extracts the content between matching fragment markers in the text.
+ *
+ * Supported markers:
+ * - ### [fragment] ... ### [fragment]
+ * - /// [fragment] ... /// [fragment]
+ *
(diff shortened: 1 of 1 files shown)


Koooooo-7 (commented):
LGTM, may add some tests for this cuz this feature seems no test cases for now？

Issue #2531, opened 2025-02-07
Is it possible not to push unnecessary state when clicking an <a> element?


jiaoliao946:
#### Problem or desire

I click an <a> element which has the href just like the location.href when setting routerMode to "history", and then it pushes a new state which is not necessary.
It might be better to use history.back() than history.pushState() when clicking an <a> element which has the href just like the location.href of the previous page in session history.

#### Proposal

Do nothing when clicking an <a> element which has the href just like the location.href.
Use history.back() when clicking an <a> element which has the href just like the location.href of the previous page in session history.

#### Implementation
Just change the onchange function to something like this:
```
e.prototype.onchange = function (i) {
        void 0 === i &&
        (i = d),
        p(
          'click',
          function (e) {
            var n = 'A' === e.target.tagName ? e.target : e.target.parentNode;
            n &&
            'A' === n.tagName &&
            !g(n.href) &&
            (
              e.preventDefault(),
              n = n.href,
              n === location.href ? 1 :      //Add
              n === hrefPre ? window.history.back() :      //Add(hrefPre is used to record the href before)
              window.history.pushState({
                key: n
              }, '', n),
              i({
                event: e,
                source: 'navigate'
              })
            )
          }
        ),
        p('popstate', function (e) {
          i({
            event: e,
            source: 'history'
          })
        })
      }
```

PR #2668, opened 2026-01-19
Fix #2628, #2531: Improve Docsify Router & Scroll Behavior for H1–H6

### LINKED ISSUE #2628: Auto scroll doesn't work for h6 headings.
### Description

I discovered this because some of the links that show up in the search plugin do not auto scroll to the correct location.

It seems that for quite some time, the [onNavigate  query selector](https://github.com/docsifyjs/docsify/blob/8016f8f35c23e9bde9d92e88770979da3960826c/src/core/event/index.js#L354) only goes down to `h5`.

Surely this line should be `:where(h1, h2, h3, h4, h5, h6)` no?

### Expected behavior

When clicking on a link that includes an anchor that is an h6 header `###### Header`, the page should navigate to the correct page and then auto scroll to the correct header.

### Actual behavior

The page navigates to the correct page but it does not scroll at all. It doesn't scroll to the anchor or to the top of the page like it should. It ends up in a location based on the `scrollTop` position of the page navigated way from, which can be quite disorienting.

### Steps to reproduce

1. Create a document `test.md` with a large amount of content to make sure that there is enough to create scrolling
2. At the bottom of the document add `###### Test`
3. In `_sidebar.md` add `- [Test](test.md#test)`
4. Load the page and click the `Test` in the sidebar
5. Observe that `test.md` is loaded but does not auto scroll to the `Test` section.

### Environment

- Docsify Version: 5.0.0-rc
- Browser Version: all

### Additional Information

- [x] Bug still occurs when all/other plugins are disabled?

KashviYadav09:
This PR addresses multiple navigation and scrolling issues in Docsify:

Auto-scroll for H6 headings (#2628):

Ensures smooth scrolling works correctly for all headings, including H6, by updating the scroll logic in render.js.

Prevent unnecessary history push (#2531):

Avoids redundant history.pushState calls when clicking a link pointing to the same page or hash.

Reduces browser history clutter and improves user experience.

Smooth scroll enhancement (additional improvement):

Ensures all headings (H1–H6) are scrolled smoothly into view using scrollIntoView({ behavior: "smooth" }).

Supports dynamic content loaded at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,3 +81,27 @@ This project exists thanks to all the people who contribute. [[Contribute](CONTR
 ## License
 
 [MIT](LICENSE)
+
+
+
+## Hosting Your Docsify Site for Free
+
+Docsify websites can be hosted easily on free platforms. Here are some popular options:
+
+### 1. GitHub Pages
+- Upload your Docsify project to a GitHub repository.
+- In `Settings` → `Pages`, select the branch to deploy.
+- Your site will be available at `https://<username>.github.io/<repo-name>/`.
+
+### 2. Netlify
+- Drag and drop your Docsify folder into [Netlify Drop](https://app.netlify.com/drop).
+- Your site will be live instantly with a free Netlify URL.
+- Supports custom domains and continuous deployment via Git.
+
+### 3. Cloudflare Pages
+- Connect your GitHub repository to [Cloudflare Pages](https://pages.cloudflare.com/).
+- Cloudflare will automatically build and host your Docsify site for free.
+- Provides fast global CDN and HTTPS by default.
+
+### 4. Other Options
+- Vercel, Surge.sh, and Firebase Hosting can also host Docsify sites for free.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -81,3 +81,27 @@ This project exists thanks to all the people who contribute. [[Contribute](CONTR
...
+- Your site will be live instantly with a free Netlify URL.
+- Supports custom domains and continuous deployment via Git.
+
+### 3. Cloudflare Pages
+- Connect your GitHub repository to [Cloudflare Pages](https://pages.cloudflare.com/).
+- Cloudflare will automatically build and host your Docsify site for free.
+- Provides fast global CDN and HTTPS by default.
+
+### 4. Other Options
+- Vercel, Surge.sh, and Firebase Hosting can also host Docsify sites for free.
```
sy-records: No need to add these.

PR #2704, opened 2026-03-27
fix(router): implement normalize() method for HTML5History

### LINKED ISSUE #2700: v5 history router crashes because HTML5History does not implement normalize()
### Repro

Using the current v5 pre-release (`docsify@5.0.0-rc.4`), configure Docsify with:

```html
<script>
  window.$docsify = {
    routerMode: 'history',
  }
</script>
<script src="//cdn.jsdelivr.net/npm/docsify@5.0.0-rc.4/dist/docsify.min.js"></script>
```

Then load the page.

### Actual result

Docsify crashes during startup with:

```text
Uncaught Error: Subclass should implement
```

In the distributed bundle this comes from the abstract `History.normalize()` implementation being called during `Router.updateRender()`.

### Expected result

`routerMode: 'history'` should initialize and render normally.

### Why this happens

The router code currently assumes every router implements `normalize()`:

- `src/core/router/index.js` calls `this.router?.normalize()` inside `updateRender()`.
- `src/core/router/history/hash.js` implements `normalize()`.
- `src/core/router/history/html5.js` does not implement `normalize()`.
- `src/core/router/history/base.js` defines `normalize()` as an abstract method that throws `Error('Subclass should implement')`.

I also checked the current `develop` branch sources on March 25, 2026, and the same mismatch is still present there.

### Notes

I hit this while migrating a Docsify site from v4 to v5. A local workaround was to patch `updateRender()` / `router.normalize` to a no-op for history mode, after which history routing worked as expected, including direct deep links.

cyphercodes:
## Summary

Fixes #2700

The HTML5History class was missing the normalize() method that is called in Router.updateRender(). This caused a crash when using routerMode: 'history' with the error 'Subclass should implement'.

## Changes

- Added normalize() method to HTML5History class
- The method applies slug replacement via replaceSlug()
- Ensures paths start with '/'
- Updates the URL via replaceState() if the path changed (without page reload)

## Testing

The fix follows the same pattern as HashHistory.normalize() but adapted for HTML5 history mode (using replaceState instead of hash manipulation).

## Checklist

- [x] Code follows the project's style guidelines
- [x] Self-review completed
- [x] Changes are focused and minimal
- [x] Related issue is referenced

### DIFF
--- a/src/core/router/history/html5.js
+++ b/src/core/router/history/html5.js
@@ -1,6 +1,6 @@
 import { isExternal, noop } from '../../util/core.js';
 import { on } from '../../util/dom.js';
-import { parseQuery, getPath } from '../util.js';
+import { parseQuery, getPath, replaceSlug, cleanPath } from '../util.js';
 import { History } from './base.js';
 
 export class HTML5History extends History {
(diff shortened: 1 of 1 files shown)


sy-records:
Thank you for your contribution.

However, a declaration here is sufficient; no implementation is necessary.

PR #2712, opened 2026-04-03
fix: add normalize method to html5.js

### LINKED ISSUE #2700: v5 history router crashes because HTML5History does not implement normalize()

sy-records:
## Related issue, if any:

Close #2700

Close #2704

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## For any code change,

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -58,18 +58,34 @@ Your `firebase.json` file should look similar to this (I changed the deployment 
 
 Once finished, build the starting template by running `docsify init ./site` (replacing site with the deployment directory you determined when running `firebase init` - public by default). Add/edit the documentation, then run `firebase deploy` from the root project directory.
 
-## VPS
+## Nginx
 
-Use the following nginx config.
+Use the following Nginx configuration.
 
 ```nginx
 server {
   listen 80;
-  server_name  your.domain.com;
+  server_name your.domain.com;
 
   location / {
     alias /path/to/dir/of/docs/;
     index index.html;
+  }
+}
+```
+
+If [`routerMode`](configuration.md#routermode) is set to `history`, use this configuration instead:
+
+```nginx
+server {
+  listen 80;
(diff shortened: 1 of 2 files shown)


paulhibbitts (approved):
Re-tested in the below codesandbox, where the current RC results in a blank page but the build of the Preview PR works as expected:
https://codesandbox.io/p/sandbox/docsify-v5-historymode-mxmwh9

Thank you! Paul

Koooooo-7 (approved):

PR #2741, opened 2026-06-05
fix: handle failed embed fetches and falsy cached() returns

Diawhiz:
- embed.js: add error handler to get().then() so a failed :include fetch calls next('') instead of silently stalling page rendering
- core/util/core.js: use  instead of  in cached() so functions that legitimately return falsy values ('' or 0) have their results stored and reused correctly

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## For any code change,

- [ ] Yes
- [ ] No

- [x] Chrome
- [x] Firefox
- [ ] Safari
- [ ] Edge


Diawhiz:
embed.js: A failed: include fetch would silently stall page rendering because the rejected promise had no error handler. Added a rejection handler that calls next('') so rendering continues gracefully on fetch failure.

core/util/core.js: cached() used a truthy check (hit ||) which caused functions returning falsy values like '' or 0 to bypass the cache on every call. Changed to hit !== undefined so falsy-but-valid results are stored and reused correctly.

Diawhiz (commented):
embed.js: A failed: include fetch would silently stall page rendering because the rejected promise had no error handler. Added a rejection handler that calls next('') so rendering continues gracefully on fetch failure.

core/util/core.js: cached() used a truthy check (hit ||) which caused functions returning falsy values like '' or 0 to bypass the cache on every call. Changed to hit !== undefined so falsy-but-valid results are stored and reused correctly.

PR #2780, opened 2026-08-07
feat(navbar): add navbarPreservePath option

### LINKED ISSUE #1220: Switch to exact document if it exists in another language
## Feature request
If I am looking at /en-us/filename.md and "filename.md" also exists in /zh-cn/, when I switch languages to /zh-cn/ it should open up "filename.md" instead of "readme.md"

A user can switch to the translated version of the same document rather than having to re-navigate the website in the target language. 

(Not as important, but also allows one to have homepages not called "readme.md" in other languages besides the default one, since it can be the same name as the default language.)

Unknown.

One should only be sent to readme.md if a file with the same name does not exist in the target language.

I lack the necessary skills.

sy-records:
## Summary

add navbarPreservePath option to maintain current document path in language links

## Related issue, if any:

Fixes #1220

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## For any code change,

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -501,6 +501,21 @@ window.$docsify = {
 };
 ```
 
+## navbarPreservePath
+
+- Type: `Boolean`
+- Default: `false`
+
+If **true**, appends the current document path to navbar links that point to a language root. This makes it possible to switch languages while staying on the corresponding document.
+
+For example, when the current path is `/quickstart`, a navbar link to `/zh-cn/` becomes `/zh-cn/quickstart`.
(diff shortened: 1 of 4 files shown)


paulhibbitts (approved):
This is a really nice refinement for the user flow of switching to another langage @sy-records ! Tested the PR Preview Build from multiple pages and all worked as expected so far.

PR #1564, opened 2021-05-02
fix sidebar closing problem on mobile while clicking among catalog items

bigablecat:
## **For any code change,**

- [x] Related documentation has been updated if needed
- [ ] Related tests have been updated or tests have been added

## **Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

## **Tested in the following browsers:**

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] IE

PR #2784, opened 2026-08-12
feat(sidebar): add collapsible root sidebar groups

### LINKED ISSUE #900: collapsible sidebar root sidebar items
## Feature request

It'd be nice to be able to expand collapse the root sidebar items.

F.e., in the Docsify guide, clicking `Customization` would collapse that section in the UI.

can make it easier to focus on specific parts of the docs.

sy-records:
## Related issue, if any:

Fixes #900

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## For any code change,

- [ ] Yes
- [x] No

- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -306,7 +306,7 @@ window.$docsify = {
 };
 ```
 
-## collapseSidebar
+## collapseSidebarGroups
 
 - Type: `Boolean`
 - Default: `false`
@@ -317,7 +317,7 @@ navigating between pages.
 
 ```js
 window.$docsify = {
-  collapseSidebar: true,
+  collapseSidebarGroups: true,
 };
 ```
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -306,6 +306,21 @@ window.$docsify = {
 };
 ```
 
+## collapseSidebar
+
+- Type: `Boolean`
+- Default: `false`
+
+Initially collapses all root sidebar groups. Visitors can still expand and
+collapse each group by selecting its title. Their choices are preserved while
+navigating between pages.
+
+```js
+window.$docsify = {
+  collapseSidebar: true,
+};
+```
+
 ## sidebarPosition
 
 - Type: `String`
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -11,6 +11,7 @@ const defaultDocsifyConfig = () => ({
   autoHeader: false,
   basePath: '',
   catchPluginErrors: true,
+  collapseSidebar: false,
   cornerExternalLinkTarget:
     /** @type {'_blank' | '_self' | '_parent' | '_top'  | '_unfencedTop'} */ (
       '_blank'
--- a/src/core/config.js
+++ b/src/core/config.js
@@ -11,7 +11,7 @@ const defaultDocsifyConfig = () => ({
   autoHeader: false,
   basePath: '',
   catchPluginErrors: true,
-  collapseSidebar: false,
+  collapseSidebarGroups: false,
   cornerExternalLinkTarget:
     /** @type {'_blank' | '_self' | '_parent' | '_top'  | '_unfencedTop'} */ (
       '_blank'
--- a/src/core/event/index.js
+++ b/src/core/event/index.js
@@ -252,6 +252,17 @@ export function Events(Base) {
 
       // Collapse toggle
       dom.on(sidebarElm, 'click', (/** @type {MouseEvent} */ { target }) => {
+        const groupTitle = /** @type {HTMLElement | null} */ (
+          /** @type {HTMLElement} */ (target).closest(
+            '.group-title[role="button"]',
+          )
+        );
+
+        if (groupTitle) {
+          this.#toggleSidebarGroup(groupTitle);
+          return;
+        }
+
         const linkElm = /** @type {HTMLElement} */ (target).closest('a');
         const linkParent = /** @type {HTMLLIElement} */ (
           linkElm?.closest('li')
@@ -262,6 +273,38 @@ export function Events(Base) {
           linkParent.classList.toggle('collapse');
         }
       });
+
+      dom.on(sidebarElm, 'keydown', (/** @type {KeyboardEvent} */ event) => {
+        const groupTitle = /** @type {HTMLElement | null} */ (
+          /** @type {HTMLElement} */ (event.target).closest(
+            '.group-title[role="button"]',
+          )
+        );
+
+        if (groupTitle && (event.key === 'Enter' || event.key === ' ')) {
+          event.preventDefault();
+          this.#toggleSidebarGroup(groupTitle);
+        }
+      });
+    }
+
+    /**
+     * Toggle a root sidebar group and keep its accessible state in sync.
+     *
+     * @param {HTMLElement} groupTitle
+     * @void
+     */
+    #toggleSidebarGroup(groupTitle) {
+      const group = /** @type {HTMLLIElement | null} */ (
+        groupTitle.closest('li')
+      );
+
+      if (!group) {
+        return;
+      }
+
+      const isCollapsed = group.classList.toggle('collapse');
+      groupTitle.setAttribute('aria-expanded', String(!isCollapsed));
     }
 
     /**
--- a/src/core/render/index.js
+++ b/src/core/render/index.js
@@ -295,7 +295,7 @@ export function Render(Base) {
 
     _renderSidebar(text) {
       const {
-        collapseSidebar,
+        collapseSidebarGroups,
         maxLevel,
         subMaxLevel,
         loadSidebar,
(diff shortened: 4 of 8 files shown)


paulhibbitts (commented):
Thanks for taking this long-requested feature on @sy-records !

Here is my test sandbox:
https://codesandbox.io/p/devbox/vibrant-brattain-hj7935

I am concerned that the introduction of using bold to mark new expandable/collapsible sections might be confusing for users and add visual complexity to the sidebar.



One alternative would be to use the existing Chevron indicators to show expandable/collapsible sections (with existing rotation behavior), and do so without little additional visual complexity. Here is a quick mockup:

Image

What do you think?

paulhibbitts:
You might also find this closed issue a helpful reference @sy-records : https://github.com/docsifyjs/docsify/pull/1564

I am not suggesting additional functionality at this point, but the comment by [jhildenbiddle](https://github.com/jhildenbiddle) was also quite insightful for future consideration re: initial expaned/collapsed state.

sy-records:
The bold text represents the original logic; now it simply supports clicking to expand or collapse it and marks its status so you don’t have to click repeatedly when viewing it.

Using a V-shaped indicator will change the current style.

paulhibbitts:
I am not sure I follow, if bold text is used, and is also in multiple locations, this will likely be confusing for users re: cue of feedback. If I understand correctly possible, I suggest we use the established chevon feedback which is also used by other similar sidebar navs like Docusaurus https://docusaurus.io/docs/category/guides.

If bold is only possible I suggest setting aside this feature as the downsides would outweigh the possible upsides to me.

sy-records:
I added logic to the sidebar-chevron.

paulhibbitts:
Oh that is starting to work much better @sy-records ! Sorry if I was not specific enough before but along with adding the Chevrons would be removing the bold styling. Once that is done, manually collapsible items will have a visual cue that they can expand/collapse and have the exact same style of the other items present in the sidebar as to not give any confusing feedback. This is going to be a great new feature!

sy-records:
My previous idea was to keep it, because it comes from the style of v4 and v5, which is just to add an unfolding and folding function.

paulhibbitts:
I am not sure for what context that style was used before (do you have an example?), but using it with the new expand/collapse feature does not seem to fit that well. I have not seen bold used in tandem with an expand/collapse behaviour.

Your addition of the chevrons works really well - could you give removing the bold a try and see how it looks/works? I think it should really unify the visual/UX design 🙂

sy-records:
Users can override the bold style by setting `--sidebar-group-title-font-weight`.

```css
--sidebar-group-title-font-weight : var(--font-weight);
```

image


What I mean is to keep it as is to highlight the hierarchical relationships.

paulhibbitts (commented):
Thanks for the additional information and context @sy-records , I did not realize the relationship to sidebar group titles - your screenshot was very helpful! Now that I see that, and there is an option to not bold as well, this combined chevron + optional bold approach looks good to me!

paulhibbitts (dismissed):
Thanks for the additional information and context @sy-records , I did not realize the relationship to sidebar group titles - your screenshot was very helpful! Now that I see that, and there is an option to not bold as well, this combined chevron + optional bold approach looks good to me!

sy-records:
@copilot resolve the merge conflicts in this pull request

paulhibbitts:
Thanks very much @sy-records , this is looking very good! I've updated the sandbox to now use group boxes and use the CSS var you shared to unbold the titles - and added more dummy menu items:

https://codesandbox.io/p/sandbox/black-https-hj7935

This made me remember an previous Docsify v4 site I created with John Hildenbiddle's Themeable which I have now converted over to v5, where I "hacked" together collapsible sidebar items for a site with a fair number of pages:

https://docs.hibbittsdesign.org/#/

You can see that it is quite valuable to be be able to also set all items as collapsed to provide a nice UX for someone to first pick a high-level category, and then explore related items etc.

This made me regret my earlier comment about not considering setting the initial collapsed state for collapsible items, seen in [PR 1563](https://github.com/docsifyjs/docsify/pull/1564) 🫤

Any possible options here from your viewpoint? Thanks very much!

paulhibbitts:
Your new config to set all sidebar groups to be initially collapsed is awesome @sy-records!

One small naming suggestion though, _collapseSidebar_ reads a bit ambiguous next to the existing hideSidebar option, since both could be read as affecting the whole sidebar panel. Since everywhere else in the implementation uses "group" (.group-title, data-group-id, #toggleSidebarGroup()), how about _collapseSidebarGroups_ instead?

paulhibbitts (approved):
This is looking great @sy-records , here is the updated CodeSandbox with the latest PR Preview Build and as well testing the ability to set **collapseSidebarGroups** to **true**:
https://codesandbox.io/p/sandbox/black-https-hj7935

All is looking good!

paulhibbitts:
Hi @sy-records , not sure if this is the best place or I should create a new issue for this - let me know!

I am starting to use the newest Docsify preview with my own Docsify-This test site, and have already run across some unintended consequences of the new collapsible sidebar being automatically applied, for example:

Docsify v5 release:


Docsify v5 Preview Build:


Note the bold group titles and new chevron icons appearing automatically, with no action taken by the user - this is of concern as we don't want to start changing the design of Docsify sites with no user confirmation/action.

Perhaps could we add a new config, set to `false` by default, to enable/disable collapsible sidebar items (i.e. `sidebarGroupsCollapsible`, `false` as default)? I think with this one change, no changes would occur on existing sites and users could then "opt-in" for the use of collapsible sidebar groups.

Alternatively, and I think the better long-term design, adopting the syntax @jhildenbiddle suggested in #1564 would also result in no change without user opt-in, and would remove the need for a config option altogether - both the already-shipped `collapseSidebar` initial-state setting and the `sidebarGroupsCollapsible` toggle I'm proposing above. Maybe this is still out-of-scope at this point, but still the best long-term solution to this issue I think.

```
* Text Heading (no expand/collapse)
  * [Page 1](page1.md)
  * [Page 2](page2.md)
+ Toggle Heading (click to expand)
  + [Page 1](page1.md)
  + [Page 2](page2.md)
- Toggle Heading (click to collapse)
  - [Page 1](page1.md)
  - [Page 2](page2.md)
```

I was not able to foresee this until I started to try out more possible Docsify test setups via Docsify-This.

Looking forward to hearing your thoughts, thanks very much Paul.

sy-records:
Can you share the link?

My original idea for this feature was simply to allow the root sidebar to be collapsed and expanded without adding a chevron icon. That way, it would essentially be a hidden feature for users and wouldn't affect them in any way.

If the user doesn't want it, they can set `':class=no-chevron'`
```markup
[Topic Three][three.md ':class=no-chevron']
```

sy-records:
Technically, you can read the symbols via Marked's `list_item.raw`, but it is not recommended to treat *, +, and - as interactive semantics.

image

paulhibbitts:
Hi @sy-records , the issue is more the unexpected change of previously non-bold items to bold. The addition of chevrons is also a visual change, but less noticeable. I reverted the collapsible sidebar group commit from my sync repo used for Docsify-This for this reason. Having done that, now I can use the Docsify-This production build vs. my preview build to show you before and afters.

Docsify-This without collapsible sidebar groups:
https://docsify-this.net/?basePath=https://raw.githubusercontent.com/hibbitts-design/docsify-this-multiple-page-open-publishing-site/main&homepage=home.md&sidebar=true&loadSidebar=_sidebar.md&hide-credits=true&core-theme-styles=true

Docsify-This Preview with collapsible sidebar groups:
https://preview.docsify-this.net/?basePath=https://raw.githubusercontent.com/hibbitts-design/docsify-this-multiple-page-open-publishing-site/main&homepage=home.md&sidebar=true&loadSidebar=_sidebar.md&hide-credits=true&core-theme-styles=true

As you can see, the now collapsible items are bold, which changes the visual appearance of Topic 3 with no user choice to do so. Chevrons are also added to two items, which is not ideal either but less of a visual change.

I hope the above very simple example shows the issue of unrequested changes happening to a user's docsify site. If a feature will change the visual design and behavior of a site we should make it opt-in, so no action will result in no site changes.

Possible ways to opt-in to this feature might be an additional config (i.e. `sidebarGroupsCollapsible`) set to false by default (maybe the simplest way?). Changing the markup required to activate collapsible items is another option, though as you noted, that's probably less ideal - and it has the same core problem: - is already the standard bullet character most existing sidebars use, so mapping it to "collapse by default" would silently apply the same unrequested bold/collapsible change to nearly every existing site, with no action taken by the user. Open to other approaches too!

Looking forward to hearing your thoughts, and anyone else here to chime in!

PS: For additional references, here is how several other doc generators handle this same situation:
https://vitepress.dev/reference/default-theme-sidebar#collapsible-sidebar-groups
https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#navigation-expansion

sy-records:
You can try #2790. There are quite a few configuration options, so I'm using class names to distinguish them.

paulhibbitts:
Thanks @sy-records , #2790 is a great step forward and solves the bold styling issue well!

That said, I'd still like to advocate for a fully opt-in approach, where no visual *or* behavioral changes happen unless a user explicitly chooses them - even with #2790, clicking now does something new on upgrade with no action taken, and there's still no way to fully disable the feature itself, only its visual presentation. In other words, a Docsify site with this additional feature should look and work identical to the current release UNTIL the user enables the new collapsible sidebar ability. 

I think the earlier proposed `sidebarGroupsCollapsible` config (`false` by default) is likely still the most straightforward way to achieve that based on things so far, please let me know what you think.

sy-records:
That's fine with me; I added `collapsibleSidebarGroups`.

paulhibbitts:
Awesome, thank you @sy-records - this new feature is looking really great 🙏🏻

